PR #25, opened 2014-02-16
setIgnoreNotImportedAnnotations(true) didn’t work for existing classes.

bfanger:
Because FuelPHP has a Package class, which isn’t an @Annotation the following error occurs:

> [Semantical Error] The class "package" is not annotated with @Annotation. Are you sure this class can be used as annotation? If so, then you need to add @Annotation to the _class_ doc comment of "package". If it is indeed no annotation, then you need to add @IgnoreAnnotation("package") to the _class_ doc comment of ...

Using @IgnoreAnnotation("package") won’t help because this issue also affects the AnnotationReader->preParser, so this error also occurs when trying to extract the @IgnoreAnnotations

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -675,7 +675,7 @@ final class DocParser
 
         // verify that the class is really meant to be an annotation and not just any ordinary class
         if (self::$annotationMetadata[$name]['is_annotation'] === false) {
-            if (isset($this->ignoredAnnotationNames[$originalName])) {
+            if ($this->ignoreNotImportedAnnotations || isset($this->ignoredAnnotationNames[$originalName])) {
                 return false;
             }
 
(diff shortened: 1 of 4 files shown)


KorvinSzanto:
What's the status on this? We have a package entity and it's causing any class that has an `@package` annotation to throw an exception. What's the best way to get around this?

bfanger:
I've rebased [the patch](https://github.com/bfanger/annotations/tree/fix-doc-parser-ignore) on the latest release v1.3.0
I hope the patch gets merged on 16 Feb 2017, on its third birthday 🎂🎂🎂

Ocramius:
@bfanger this was closed in 2014 - is there an open PR for the patch, or did I miss something?

bfanger:
Closed but never fixed, as I said:
> The behavior of setIgnoreNotImportedAnnotations() is still broken.

No open PR, you can run:
`git pull https://github.com/bfanger/annotations.git fix-doc-parser-ignore`
to merge the patch.

Ocramius:
@bfanger not merging without PR and public review

PR #29, opened 2014-04-14
Deprecate AnnotationRegistry in favor of class_exists()

kingcrunch:
See also: https://github.com/doctrine/common/issues/321

Actually it looks like there is no benefit in using a separate autoloading mechanism, which itself only makes (manual) use of "real" autoloaders. Instead `class_exists()` serves the same purpose and prevents one from the need to [manually add an autoloader](https://github.com/symfony/symfony-standard/blob/master/app/autoload.php#L11) for every loader, that should load annotations.

This is especially interesting now, that composer took over most of the autoloading related issues in many projects.

I don't know, which deprecation-rules apply here, so I just added the tag and kept the class as it is, but remove the use wherever useful.


KingCrunch:
See #61

PR #51, opened 2014-12-23
FileCacheReader::saveCacheFile::unlink fix

adammbalogh:
~~`$tempfile` is not correct because there's a `rename` func. call on `$tempfile` (to `$path`) before the `unlink` func. call~~


oleg-andreyev:
When this fix will land in 1.2.4?

krewetka:
:+1:

GromNaN:
:+1:

Can you tag a 1.2.4 @Ocramius ?

Ocramius:
Tagged.

PR #52, opened 2014-12-26
Add annotations of PSR-5 (PHPDoc standard)

dunglas:
What about adding `@type` too? It has been removed from the PSR-5 draft but is already used in the wild (for instance: https://github.com/milesj/decoda/pull/71).


Ocramius:
Won't fix until PSR-5 is an actual thing.

PR #53, opened 2015-03-23
Add $context property to annotation constructor

mdzzohrabi:
For some purpose we need to detect annotation target, class or property name.
for sample , in a Column annotation that has a tableName property i want to create this property if it not set by user from class property name.

```php
class Blog {
       /** @Column **/
       protected $AuthorId;
}

/**
 * @Annotation
 * @Target({ "PROPERTY" })
 */
class Column {
      public $tableName;
      public __construct( array $values , $context ) {
             $propertyName = substr( $context , strpos( $context , '$' ) + 1 );
             // Convert "AuthorId" to "author_id"
             $this->tableName = String::underscore( $propertyName );
      } 
}
```


Ocramius:
This is a technical requirement that doesn't fit the nature of an annotation. An annotation is agnostic to the environment it is placed in: use the annotation reader for that.

PR #54, opened 2015-04-14
Support of inherited annotations from parent method

mente:
Hello!

This PR adds support for inherited annotations in following cases:

* method doc comment is empty
* method doc comment contains `@inheritDoc`

In second case annotations will be merged from parent and current methods. I've doubts regarding first case as it can be a big BC in certain ways.

Please let me know what you think about it.

Regards,
Alex

### DIFF
--- a/lib/Doctrine/Common/Annotations/Annotation/InheritDoc.php
+++ b/lib/Doctrine/Common/Annotations/Annotation/InheritDoc.php
@@ -1,0 +1,19 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: avasilenko
+ * Date: 09/04/15
+ * Time: 15:28
+ */
+
+namespace Doctrine\Common\Annotations\Annotation;
+
+
+/**
+ * Annotation that signals to merge annotations from parent method/class with any found in this method/class
+ * @Annotation
+ */
+final class InheritDoc
+{
+
+}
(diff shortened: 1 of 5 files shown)


zeroedin-bill:
The first case definitely is a BC break.

mente:
I'm OK with removing it. Any other doubts?

zeroedin-bill:
@mente It's not entirely clear whether ```@inheritDoc``` will be part of the standard. As of now only the inline {@inheritDoc} is in the standard. The other parts of a docblock are supposed to be automatically inherited, but since annotations aren't officially part of PHP or the phpdoc standard, the path forward on this is unclear.

@guilhermeblanco any thoughts?

Ocramius:
Didn't review yet, but please note that `Foo#methodName()` and `Bar#methodName()` are different methods, and even if you are "used to" see them as an inheritance, annotations should not be inherited.

guilhermeblanco:
We should follow what other languages suggest here. In Java, only annotations marked as @Inherited are available in subclasses. We should not do this by default as this PR suggests.
Besides the big WTH that can happen, there would be no way to inherit documentation but not inherit annotations with this PR. Or would we add *another* support for that?
Let's not reinvent the wheel with confusing annotations. If you're interested to support this, add the support to @Inherited at the annotation level and manage to get them merged in subclasses elements.

mente:
I've just wanted to add annotation inheritance either via current `@inheritDoc` or via any other. `@Inherited` is ok for me. Stand by

mente:
See #55

PR #55, opened 2015-04-14
@Inherited annotation for annotation inheritance from parent method

mente:
This is a rewrite of #54 based on @guilhermeblanco suggestions.

Closes #54

### DIFF
--- a/lib/Doctrine/Common/Annotations/Annotation/Inherited.php
+++ b/lib/Doctrine/Common/Annotations/Annotation/Inherited.php
@@ -1,15 +1,29 @@
 <?php
-/**
- * Created by PhpStorm.
- * User: avasilenko
- * Date: 14/04/15
- * Time: 17:34
+
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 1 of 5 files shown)


mente:
There's some weird error in hhvm test. Can anybody help me with it?

```
Fatal error: Class already declared: PHPUnit_Framework_Error in phar://phpunit-4.6.2.phar/phpunit/Util/../Framework/Error.php on line 22
```

Or it's ok?

guilhermeblanco:
@mente sorry, you did not get my request.
The annotation helper @Inherited do not get applied where it's used, but at the annotation declaration level.

Here is a comparison of what you've done (first example) and what I've asked (second example):

What you've done:
```
@Annotation
class Foo {}

abstract class Bar {
    @Foo
    public function myMethod() {}
}

class Fuzz extends Bar {
    @Inherited
    public function myMethod() {}
}
```

In your case, it would return all Annotations declared in `myMethod()`. Using the example, you'd have `[<Foo instance>]` back

This is different than inherited annotations as detailed in Java ( https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/Inherited.html ) where this applies to Annotation type declaration. This is the updated example following what Java does:

```php
@Annotation
@Inherited
class Foo {}

abstract class Bar {
    @Foo
    public function myMethod() {}
}

class Fuzz extends Bar {
    public function myMethod() {}
}
```

That would still return to you `[<Foo instance>]` when asking for method annotations, but the inheritance acceptance is controlled by the Annotation itself, not by your code.

Makes sense now?

mente:
Now I see. But that's not what I wanted to have. My original idea was to control annotations on application level, not on annotations level

mente:
IMHO this implementation has more pluses than java-style `@Inherited`. My thoughts:

Current implementation:
+ allows control of inheritance in user space
+ no need for users to wait for library update - it's already in user space!

Java style @Inherited:
- harder to add for library creators to existing annotations: potential BC

Of course java creators had same thoughts during development but I don't understand why they choose annotation-level inheritance.

Let me know what you think about it

Ocramius:
Looking at this again, I don't see any particular use-case for it (except for a load of confusion).

I fully agree with @guilhermeblanco on the fact that the `@Inherited` should be at annotation level, but even then, I don't see good use-case scenarios for it.

PR #56, opened 2015-04-23
Allow SimpleAnnotationReader to use the global ignores

alexpott:
Whilst having a look at the classmap Drupal 8 generates I noticed we had a lot of classes called "see" in different namespaces. All mapped to NULL - ie. they don't exist. I tracked this down to the SimpleAnnotationReader finding doc blocks with @see annotations. The regular AnnotationReader ignores these so shouldn't the SimpleAnnotationReader too?

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -113,6 +113,16 @@ class AnnotationReader implements Reader
     }
 
     /**
+     * Gets the globally ignored annotation names.
+     *
+     * @return array An array of Annotation names to ignore.
+     */
+    static public function getGlobalIgnoredNames()
+    {
+        return self::$globalIgnoredNames;
+    }
+
+    /**
      * Annotations parser.
      *
      * @var \Doctrine\Common\Annotations\DocParser
--- a/tests/Doctrine/Tests/Common/Annotations/SimpleAnnotationReaderTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/SimpleAnnotationReaderTest.php
@@ -85,9 +85,9 @@ class SimpleAnnotationReaderTest extends AbstractReaderTest
         $class  = new \ReflectionClass('Doctrine\Tests\Common\Annotations\Fixtures\ClassDDC1660');
 
         $this->assertTrue(class_exists('Doctrine\Tests\Common\Annotations\Fixtures\Annotation\Version'));
-        $this->assertCount(1, $reader->getClassAnnotations($class));
-        $this->assertCount(1, $reader->getMethodAnnotations($class->getMethod('bar')));
-        $this->assertCount(1, $reader->getPropertyAnnotations($class->getProperty('foo')));
+        $this->assertCount(0, $reader->getClassAnnotations($class));
+        $this->assertCount(0, $reader->getMethodAnnotations($class->getMethod('bar')));
+        $this->assertCount(0, $reader->getPropertyAnnotations($class->getProperty('foo')));
     }
 
     protected function getReader()
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -672,7 +672,7 @@ final class DocParser
             $alias = (false === $pos = strpos($name, '\\'))? $name : substr($name, 0, $pos);
             $found = false;
 
-            if ($this->namespaces) {
+            if ( ! isset($this->ignoredAnnotationNames[$name]) && $this->namespaces) {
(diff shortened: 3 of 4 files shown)


dawehner:
Looks fine for me!

paul-m:
+1. Issue in Drupal 8 queue here, with test: https://www.drupal.org/node/2557433

If $parser were protected instead of private, we could subclass.

lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -113,6 +113,16 @@ static public function addGlobalIgnoredName($name)
     }
 
     /**
+     * Gets the globally ignored annotation names.
+     *
+     * @return array An array of Annotation names to ignore.
+     */
+    static public function getGlobalIgnoredNames()
```
Ocramius: The need for this smells of encapsulation breakage: there should be a solution that doesn't involve accessing the `AnnotationReader` directly. Maybe mirror the same structure inside the `SimpleAnnotationReader`?

tests/Doctrine/Tests/Common/Annotations/SimpleAnnotationReaderTest.php
```diff
@@ -85,9 +85,9 @@ public function testInvalidAnnotationButIgnored()
         $class  = new \ReflectionClass('Doctrine\Tests\Common\Annotations\Fixtures\ClassDDC1660');
 
         $this->assertTrue(class_exists('Doctrine\Tests\Common\Annotations\Fixtures\Annotation\Version'));
-        $this->assertCount(1, $reader->getClassAnnotations($class));
-        $this->assertCount(1, $reader->getMethodAnnotations($class->getMethod('bar')));
-        $this->assertCount(1, $reader->getPropertyAnnotations($class->getProperty('foo')));
+        $this->assertCount(0, $reader->getClassAnnotations($class));
```
Ocramius: There should be an additional (explicit) test for excluded annotations.

Ocramius:
Overall tempted to reject this as per http://www.doctrine-project.org/jira/browse/DCOM-295

Ocramius:
s/tempted/closing. Sorry, let's try keeping the `SimpleAnnotationReader` from spreading more.

TomasVotruba:
@Ocramius :+1: For gradual deprecating.

Ocramius:
Sigh, I really don't understand how this drupal tracker works. Trying again:

 - I just rejected https://github.com/doctrine/annotations/pull/56
 - see http://www.doctrine-project.org/jira/browse/DCOM-295

@paul-m ^

dawehner:
@Ocramius Is there a way for the normal AnnotationReader without having users to put in the various use statements? SimpleAnnotationReader was kinda a great DX, to be honest.

We can control for our particular example which annotation can ever exist, so uniqueness issues aren't a problem.

stof:
@dawehner it is not that great for DX actually: it is unable to warn you when you make a typo in an annotation. It will just ignore it entirely.

dawehner:
Won't it tell you that there is no such annotation? I'm not sure about that, this is for sure.

Ocramius:
> @Ocramius Is there a way for the normal AnnotationReader without having users to put in the various use statements?

No, it's supposed to be used with all the imports.

 > SimpleAnnotationReader was kinda a great DX, to be honest.

After looking at all the historically raised issues around the `SimpleAnnotationReader`, I can simply negate this. Explicitness makes the annotation reader work much much better, whereas the simplified annotation reader always led to bugs and problems due to its implicitness.

 > Won't it tell you that there is no such annotation?

It will silently fail by design (another reason to get rid of it).

Additionally, any project growing over a certain size will have annotation names conflicts (ZF+Doctrine already have that), so there's an even bigger can of worms hidden there.

dawehner:
I get your point and it certainly makes sense in the context of ORM annotations.

The thing is, that for our particular usecase a single class controls (by design) the available annotations.
There won't be any annotation name conflicts. Also for that particular usecase the available annotations
are bound to one domain, so an additional explicitness doesn't add any semantic meaning.

Ocramius:
> The thing is, that for our particular usecase a single class controls (by design) the available annotations.

That goes against the implemented readers design anyway: you'd use the parser directly and re-implement the reader instead.

 > There won't be any annotation name conflicts. 

This may be true for drupal, isn't true for the wider audience.

 > Also for that particular usecase the available annotations
are bound to one domain, so an additional explicitness doesn't add any semantic meaning.

It actually does, as you just explained something to me that would take ages for me to understand (as a doctrine developer) if I were to work with a drupal module operating under such assumptions. Assumptions are bad anyway: get rid of them whenever possible.

paul-m:
Heads up: https://github.com/doctrine/annotations/pull/211

PR #57, opened 2015-05-13
Switch to the docker-based infrastructure on Travis

stof:
This infrastructure is faster and allows to use the composer cache.


Ocramius:
:+1:

PR #58, opened 2015-06-01
Adds trait annotations to classes

crazycodr:
See https://github.com/doctrine/phpcr-odm/issues/639


crazycodr:
There are 4 skipped tests at the beginning of the test suite that i do not have on my local machine. I'm not sure if it has anything to do with the fix i pulled in for the version constraint. I doubt it, but if you can confirm that would be great!

frankdejonge:
This would be a nice compliment to the existing functionality, very much in the spirit of traits.

crazycodr:
Wow it's been like 2 months i did that... Will it ever be merged?

Ocramius:
@crazycodr 2 months is quite a short period of time compared to the backlog of todos :-\

Ocramius:
@frankdejonge consider adding a test case, if you need this merged. Dropping PHP 5.3 and 5.4 support is also a nice to have here.

frankdejonge:
@Ocramius a case can be made from adding timestamps through a trait, which adds behaviour, like annotation can do. In which case the entity would require lifecycle events, which can be inherited from the trait. Otherwise any class which relies on this behaviour would be required to add it to it to itself, which doesn't feel "natural". So, in the spirit of traits, full annotation support for their annotated counterparts.

crazycodr:
I'll see what i can do about that test case, i've got lots of things on my hands right now

codesource:
I did a pull request to your code to improve this new feature

lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -191,7 +191,24 @@ public function getClassAnnotations(ReflectionClass $class)
...
 
-        return $this->parser->parse($class->getDocComment(), 'class ' . $class->getName());
+        $annotations = $this->parser->parse($class->getDocComment(), 'class ' . $class->getName());
+        
+        if (!method_exists($class, 'getTraits')) {
+            return $annotations;
+        }
+        
+        $ignoredAnnotationNames = array_map('get_class', $annotations);
+        foreach($class->getTraits() as $trait) {
```
Ocramius: Breaks 5.3 compat

lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -191,7 +191,24 @@ public function getClassAnnotations(ReflectionClass $class)
...
+            $this->parser->setImports($this->getClassImports($trait));
+            foreach($this->parser->parse($trait->getDocComment(), 'trait ' . $trait->getName()) as $annotation){
```
Ocramius: What about methods imported from the trait (but replaced by the implementing class)?

Ocramius:
As mentioned before, tests are missing.

lenybernard:
+1

blackness90:
Why this was not merged?

crazycodr:
Tests are missing, i will remove myself from this thread as i have not used Doctrine in like 2 years now. Please fork my branch and reopen this if you want this feature...

allan-simon:
@Ocramius  so basically the PR is good, and we're just missing somebody to add some test to make sure it actually works ?

Ocramius:
@allan-simon yes, and verify all edge cases. Patches without tests are basically not considered at all.

allan-simon:
just to confirm , actually this MR is no more needed the current code (as of july 2017) already handle trait

PR #59, opened 2015-06-16
opcache.load_comments has been removed from PHP 7

mpalourdio:
When building a very fresh php7 version from source this morning, it appeared that one of my projects using doctrine/annotations didn't work anymore.

It seems that the support of this option has been removed from php7, see https://github.com/php/php-src/commit/0a21a0c752a06d25704e04c15464700f75164849

I don't know if this will be backported to php5.6 and/or 5.5 for the moment.

This PR simply checks the php version before checking the value of opcache.load_comments.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7
   - hhvm
 
 before_script:
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -162,13 +162,14 @@ class AnnotationReader implements Reader
             throw AnnotationException::optimizerPlusSaveComments();
         }
 
+        if (PHP_VERSION < 70000) {
+            if (extension_loaded('Zend Optimizer+') && (ini_get('zend_optimizerplus.load_comments') === "0" || ini_get('opcache.load_comments') === "0")) {
+                throw AnnotationException::optimizerPlusLoadComments();
+            }
 
-        if (extension_loaded('Zend Optimizer+') && (ini_get('zend_optimizerplus.load_comments') === "0" || ini_get('opcache.load_comments') === "0")) {
-            throw AnnotationException::optimizerPlusLoadComments();
-        }
-
-        if (extension_loaded('Zend OPcache') && ini_get('opcache.load_comments') == 0) {
-            throw AnnotationException::optimizerPlusLoadComments();
+            if (extension_loaded('Zend OPcache') && ini_get('opcache.load_comments') == 0) {
+                throw AnnotationException::optimizerPlusLoadComments();
+            }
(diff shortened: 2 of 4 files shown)


Ocramius:
@mpalourdio please also add PHP7 to the build matrix

mpalourdio:
That's interesting, because some tests fail now because of new reserved words. Particularly because of these classes https://github.com/doctrine/annotations/blob/master/tests/Doctrine/Tests/Common/Annotations/DocParserTest.php#L1360-L1366

What to do ? Skip those tests for PHP7 ? Or is it relevant to simply rename those classes and the references in  ```testReservedKeywordsInAnnotations``` to NullClass, TrueClass etc, as it will crash anyway with php7 ?

Ocramius:
@mpalourdio can it work with `class_alias()`?

stof:
@Ocramius no it cannot. class_alias still defines a class.

I suggest moving these annotations to a separate file and skipping the tests using them on PHP 7 (renaming them to ``NullClass`` or ``TrueClass`` would actually make the test useless)

mpalourdio:
I don't think so :(

Ocramius:
I'd rather drop the the functionality for 2.x, and support PHP 7 only in 2.x

stof:
@Ocramius all the codebase is compatible with PHP 7. We just need to skip this *test* on PHP 7 because it does not make sense on this version. So IMO, refactoring the test is still a good idea.

Ocramius:
Hmm, right, you won't be able to define classes `True`, `False` or `null` on PHP 7 anyway.

@mpalourdio can you try skipping the test on PHP7 (as long as defining the classes only if PHP < 7)?

mpalourdio:
Ok, i've just made a fix to this test, but there is now 43 other tests that fail because of some ```@expectedException```.  I have to investigate a bit more, but my vagrant instance is WTFing at the moment !

You should leave this PR open as the first commits are still relevant, because we need to skip checking opcache.load_comments for php7, no ?

Ocramius:
Yep, let the PR open for now

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

stof:
Well, tests are passing on Travis, so the current state of the PR seems fine. what is blocking the merge ?

guilhermeblanco:
Nothing... patch is ok as it is. =)

pkruithof:
Seems like [v1.2.5][1] is done. Could you guys create a release so I can start testing for PHP7 again?

[1]: https://github.com/doctrine/annotations/milestones/v1.2.5

lisachenko:
+1 for publishing a new release version. Just spent several minutes looking at the travis-ci failed build with 1.2.4 version and PHP7.

mpalourdio:
@lisachenko v1.2.6 has been tagged 3 days ago

lisachenko:
ah, composer.lock ) @mpalourdio thanks for pointing this

PR #61, opened 2015-06-22
Deprecate AnnotationRegistry in favor of class_exists()

kingcrunch:
Replacement for #29. See over there for discussion

I've removed the fork and therefore the PR got unlinked and unmanageable somehow. The previous PR had merge conflicts, but I wanted to keep it, so I've re-created it.

> See also: doctrine/common#321
> 
> Actually it looks like there is no benefit in using a separate autoloading mechanism, which itself only
> makes (manual) use of "real" autoloaders. Instead class_exists() serves the same purpose and
> prevents one from the need to manually add an autoloader for every loader, that should load
> annotations.
> 
> This is especially interesting now, that composer took over most of the autoloading related issues in
> many projects.
> 
> I don't know, which deprecation-rules apply here, so I just added the tag and kept the class as it is, but
> remove the use wherever useful.

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -171,8 +171,8 @@ class AnnotationReader implements Reader
                 throw AnnotationException::optimizerPlusLoadComments();
             }
         }
-
-        AnnotationRegistry::registerFile(__DIR__ . '/Annotation/IgnoreAnnotation.php');
+        
+        require_once __DIR__ . '/Annotation/IgnoreAnnotation.php';
 
         $this->parser    = new DocParser;
         $this->preParser = new DocParser;
--- a/lib/Doctrine/Common/Annotations/AnnotationRegistry.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationRegistry.php
@@ -21,6 +21,8 @@ namespace Doctrine\Common\Annotations;
 
 /**
  * AnnotationRegistry.
+ *
+ * @deprecated Rely on spl_register_autoloader() autoloaders instead
  */
 final class AnnotationRegistry
 {
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -438,13 +438,8 @@ final class DocParser
             return $this->classExists[$fqcn];
         }
 
-        // first check if the class already exists, maybe loaded through another AnnotationReader
-        if (class_exists($fqcn, false)) {
-            return $this->classExists[$fqcn] = true;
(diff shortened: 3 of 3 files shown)


DavidBadura:
a big +1 for this! in each lib from me i have to add

```php
\Doctrine\Common\Annotations\AnnotationRegistry::registerLoader('class_exists');
```

or i need to document it.

Ocramius:
I think this can be applied once http://www.doctrine-project.org/jira/browse/DCOM-295 is done.

Ocramius:
Personally fully agree with the patch.

As per discussions with internals, the autoloader was mainly meant for:
 - fixing problems with non-silent autoloaders (autoloaders that cry loud when trying to load non existing classes)
 - ignoring some classes (specifically)

I don't think non-silent autoloaders are a thing anymore, since @stof pretty much helped in killing them gradually.

As for the "ignoring" functionality, if [DCOM-295](http://www.doctrine-project.org/jira/browse/DCOM-295) lands, then there is no need for such features, as the imports decide what annotations should be loaded or not.

Still needs some discussion with the others at @doctrine/team-doctrine2

stof:
> I don't think non-silent autoloaders are a thing anymore, since @stof pretty much helped in killing them gradually.

Most of the work has been done by the composer adoption here actually: the composer autoloader is silent when it cannot find the class, and most projects are now using it.

Ocramius:
Yup, but also other autoloaders (including ours) were simply "fixed"

Jalle19:
Just wanted to say that this doesn't work if you use this in a ZF1 project because its autoloader doesn't fail silently. I've resolved this by using this instead:

```php
AnnotationRegistry::registerLoader(function($fqcn) {
    return @class_exists($fqcn);
});
```

stof:
@Jalle19 an autoloader breaking ``class_exists`` should die IMO. I suggest you to ask the ZF team if they could fix the ZF1 autoloader (as it seems ZF1 is still a thing)

Ocramius:
Fully agree: the ZF1 class loader should be fixed instead. Note that you may also just use the composer class loader also for ZF1 projects.

Jalle19:
I don't really care, I wouldn't touch ZF1 unless I had to (working on a legacy application). Any solution that means people don't have to register a loader manually anymore is good in my book.

Ocramius:
@Jalle19 for new projects, you wouldn't pick something like ZF1 anyway. This is why we're discussing about deprecation and removal in next major versions.

hboomsma:
:+1:

Majkl578:
@KingCrunch Hi, I reapplied your changes onto current master as part of #205, let's make this happen. :sunglasses: 
Meanwhile closing here, let's handle it in the referenced PR.

PR #62, opened 2015-09-10
[CachedReader] Test traits and parent class to see if cache is fresh

iamluc:
This PR is useful if for exemple you work with traits in your entities or with a parent class.
Before, the cache was not refreshed when you made changes in traits or parent class.

### DIFF
--- a/lib/Doctrine/Common/Annotations/CachedReader.php
+++ b/lib/Doctrine/Common/Annotations/CachedReader.php
@@ -226,10 +226,43 @@ final class CachedReader implements Reader
      */
     private function isCacheFresh($cacheKey, \ReflectionClass $class)
     {
-        if (false === $filename = $class->getFilename()) {
+        if (null === $lastModification = $this->getLastModification($class)) {
             return true;
         }
 
-        return $this->cache->fetch('[C]'.$cacheKey) >= filemtime($filename);
+        return $this->cache->fetch('[C]'.$cacheKey) >= $lastModification;
+    }
+
+    /**
+     * Returns the time the class was last modified, testing traits and parents
+     *
+     * @param \ReflectionClass $class
+     * @return int|null
+     */
+    private function getLastModification(\ReflectionClass $class)
+    {
+        if (false === $filename = $class->getFilename()) {
+            return;
+        }
+
+        $lastModification = filemtime($filename);
+
+        // Test traits
+        if (method_exists($class, 'getTraits')) {
+            foreach($class->getTraits() as $trait) {
+                if (false === $traitFilename = $trait->getFilename()) {
+                    continue;
+                }
+                $lastModification = max($lastModification, filemtime($traitFilename));
+            }
+        }
+
+        // Test parents
+        while ($parent = $class->getParentClass()) {
+            $lastModification = max($lastModification, $this->getLastModification($parent));
+            $class = $parent;
+        }
+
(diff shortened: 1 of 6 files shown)


curry684:
This PR probably fixes an issue I've been having with Symfony. Our workflow includes generating the mapping into a separate trait, so it can be regenerated when needed without hitting business functionality. For example:
```php
/*
 * @ORM\Entity
 */
class MyEntity
{
    use Generated\MyEntityTrait;
}
```
And the trait:
```php
trait PageTrait
{
    /**
     * @ORM\Column(type="string", length=255)
     */
    private $slug;
}
```
Running `doctrine:schema:update` does **not** detect any changes when I change the `length` property or make it nullable. It **does** pick it up if:
 - `$slug` is renamed or anything else is modified in the trait
 - Anything, including annotations, is changed in the class itself

While digging into this I ran into this PR which likely fixes it. Any timeline on merging?

mikeSimonson:
@Ocramius This seems good to go to me.
Maybe just adding a test case with a parent class that has a trait.

shochdoerfer:
Any chance the change can be merged in soonish? Hitting the same problem with [bitexpert/disco](https://github.com/bitExpert/disco) when using traits to split up the DI config into smaller components.

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -226,10 +226,43 @@ private function saveToCache($rawCacheKey, $value)
...
+    private function getLastModification(\ReflectionClass $class)
+    {
+        if (false === $filename = $class->getFilename()) {
+            return;
+        }
+
+        $lastModification = filemtime($filename);
+
+        // Test traits
+        if (method_exists($class, 'getTraits')) {
```
Ocramius: To be moved to a private method

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -226,10 +226,43 @@ private function saveToCache($rawCacheKey, $value)
...
+            foreach($class->getTraits() as $trait) {
+                if (false === $traitFilename = $trait->getFilename()) {
+                    continue;
+                }
+                $lastModification = max($lastModification, filemtime($traitFilename));
+            }
+        }
+
+        // Test parents
+        while ($parent = $class->getParentClass()) {
```
Ocramius: To be moved to a private method

tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php
```diff
@@ -13,23 +13,50 @@ class CachedReaderTest extends AbstractReaderTest
...
+        $cache = time() - 10;
+        touch(__DIR__.'/Fixtures/ControllerWithParentClass.php', $cache - 10);
+        touch(__DIR__.'/Fixtures/AbstractController.php', $cache + 10);
+
+        $this->doTestCacheStale('Doctrine\Tests\Common\Annotations\Fixtures\ControllerWithParentClass', $cache);
+    }
+
+    public function testIgnoresStaleCacheWithTraits()
+    {
+        if (version_compare(PHP_VERSION, '5.4.0') < 0) {
```
Ocramius: We don't support these versions anymore, luckily.

Ocramius:
Moved rest of the PR to #96

PR #63, opened 2015-09-25
[RFC] Parse trait class annotations

kinncj:
The parser wasn't parsing `trait` class annotations as per ticket doctrine/doctrine2#1517

### DIFF
--- a/lib/Doctrine/Common/Annotations/SimpleAnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/SimpleAnnotationReader.php
@@ -64,7 +64,17 @@ class SimpleAnnotationReader implements Reader
      */
     public function getClassAnnotations(\ReflectionClass $class)
     {
-        return $this->parser->parse($class->getDocComment(), 'class '.$class->getName());
+        if (!method_exists($class, 'getTraits')) {
+            return $this->parser->parse($class->getDocComment(), 'class '.$class->getName());
+        }
+
+        $parsed = array();
+
+        foreach ($class->getTraits() as $reflectedtrait) {
+            $parsed = array_merge($parsed, $this->parser->parse($reflectedtrait->getDocComment(), 'class '.$reflectedtrait->getName()));
+        }
+
+        return array_merge($this->parser->parse($class->getDocComment(), 'class '.$class->getName()), $parsed);
     }
 
     /**


kinncj:
ping @guilhermeblanco @FabioBatSilva

kinncj:
@FabioBatSilva, @guilhermeblanco, any updates regarding this PR?

guilhermeblanco:
Was supposed to review and merge yesterday.... forgot to do it... =(
Will do during this weekend.

mweimerskirch:
I had issues with this as well using 1.2.7. "Assert" annotations defined inside traits were ignored.
This patch worked for me.

nlemahieu:
:+1: i had the same issue with 1.2.7 and this fix works for me

kinncj:
@guilhermeblanco and @Ocramius doesn't like me... I guess they would never merge...
note for future me: don't drink with Blanco and Pivetta on future #WurstCon

:smile:

nlemahieu:
i can't add the class use on the top of the entity because php-cs-fixer which is included in Travis doesn't like it...

deeky666:
@kinncj this requires a test case.

kinncj:
@deeky666 you must read the description :-)

> The parser wasn't parsing trait class annotations as per ticket doctrine/doctrine2#1517

guilhermeblanco:
@deeky666 this cannot be merged since it may potentially have 2 Traits declaring the same Annotations.

kinncj:
> Some developers are more copy and paster than coders.
Anyway, you are splitting one unique behaviour in two different participants then... the `trait` with `properties` and `HasLifecycleCallbacks`(callbacks) and the annotation for `HasLifecycleCallbacks` in the main participant (class). which means: the class annotation won't work without the trait and the trait won't do anything if the class doesn't annotate that... so instead of a Trait, it's better to just add the `HasLifecycleCallbacks`(the callback methods) to the main participant then.

Ocramius:
Besides https://github.com/doctrine/annotations/pull/63#issuecomment-171659515, extensive testing is *required* before considering anything for merge

Majkl578:
I'm closing this PR because the approach is not correct - merging `use`s of class' scope and traits' scope is not acceptable and may have grave consequences.
See discussion in #102 for alternative and correct approach.

PR #64, opened 2015-09-30
FIxed issue with debug mode

iainconnor:


Ocramius:
@iainconnor requires a test case

iainconnor:
@Ocramius What sort of test would you be looking for? From http://docs.doctrine-project.org/projects/doctrine-common/en/latest/reference/annotations.html

> If you set the debug flag to true the cache reader will check for changes in the original files, which is very important during development. If you don’t set it to true you have to delete the directory to clear the cache. This gives faster performance, however should only be used in production, because of its inconvenience during development.

Which doesn't seem to be the behaviour here. If data incorrectly identified as "fresh" is in the cache, it will be returned from this method, regardless of the `$debug` flag.

Ocramius:
@iainconnor the test should verify the correct behavior. Use a cache mock (or an array cache) and try verifying that your code works, whereas the previous version fails?

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -191,7 +191,7 @@ private function fetchFromCache($rawCacheKey, \ReflectionClass $class)
     {
         $cacheKey = $rawCacheKey . self::$CACHE_SALT;
         if (($data = $this->cache->fetch($cacheKey)) !== false) {
-            if (!$this->debug || $this->isCacheFresh($cacheKey, $class)) {
+            if (!$this->debug && $this->isCacheFresh($cacheKey, $class)) {
```
Ocramius: The original code seems correct: If we aren't in debug mode, then the cache freshness doesn't need to be checked. If we are in debug mode, then check the cache freshness.

Ocramius:
Closing as `invalid` as per https://github.com/doctrine/annotations/pull/64/files#r128893377

PR #65, opened 2015-11-21
Remove cache salt making key unnecessarily long

Tobion:
The salt just makes the key unnecessarily long and some caches have a limit in the length of the cache key or long keys just make it slower or require more space.
The cache key must only be unique within its own domain. If one uses a shared cache, that's where namespaces are used. These are also supported by the doctrine cache. So adding a salt as well makes no sense.
For a concrete problem I found due to long cache key, see https://github.com/doctrine/cache/pull/107


TomasVotruba:
:+1:

Ocramius:
Moved to #98

PR #66, opened 2015-11-23
Fix of incorrect parsing multibyte strings

dkop:
If function overloading for mb_string (http://php.net/manual/en/mbstring.overload.php) is turned on then reading of annotations with unicode strings works incorrect.
Just don't use array access for string next time :)

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -347,8 +347,9 @@ final class DocParser
 
         // search for first valid annotation
         while (($pos = strpos($input, '@', $pos)) !== false) {
+            $preceding = substr($input, $pos - 1, 1);
             // if the @ is preceded by a space or * it is valid
-            if ($pos === 0 || $input[$pos - 1] === ' ' || $input[$pos - 1] === '*') {
+            if ($pos === 0 || in_array($preceding, array(' ', '*'))) {
                 return $pos;
             }
 
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -349,7 +349,7 @@ final class DocParser
         while (($pos = strpos($input, '@', $pos)) !== false) {
             $preceding = substr($input, $pos - 1, 1);
             // if the @ is preceded by a space or * it is valid
-            if ($pos === 0 || $preceding === ' ' || $preceding === '*')) {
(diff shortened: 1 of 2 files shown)


Ocramius:
@dkop this needs a test case

dkop:
@Ocramius ok, test is added

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -1300,6 +1300,32 @@ public function testSupportsEscapedQuotedValues()
...
     }
+
+    /**
+     * @see http://php.net/manual/en/mbstring.configuration.php
+     * mbstring.func_overload can be changed only in php.ini
+     * so for testing this case instead of skipping it you need to manually configure your php installation
+     */
+    public function testMultiByteAnnotation()
+    {
+        $overloadStringFunctions = 2;
```
Ocramius: Is there any constant to be referenced here? Or did you just point at the INI setting?

dkop: No, i didn't found constant for this value. You can check also this page http://php.net/manual/en/mbstring.overload.php . May be this link need to be added to test description?

Ocramius: I see you already did it :-)

dkop: No, there is two different links, but you can go from page in code to second page :-)

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -347,8 +347,9 @@ private function findInitialTokenPosition($input)
 
         // search for first valid annotation
         while (($pos = strpos($input, '@', $pos)) !== false) {
+            $preceding = substr($input, $pos - 1, 1);
             // if the @ is preceded by a space or * it is valid
-            if ($pos === 0 || $input[$pos - 1] === ' ' || $input[$pos - 1] === '*') {
+            if ($pos === 0 || in_array($preceding, array(' ', '*'))) {
```
stof: strict comparison should still be used

dkop:
@stof ok, you are right, i changed it back to strict comparison

stof:
@Ocramius what about merging this (even though I would prefer ``mbstring.func_overload`` to die, but it will not happen soon)

Ocramius:
@dkop sorry for the late review/merge, LGTM! 👍

Issue #68, opened 2016-01-08
Annotations that are indented by tab are not processed.


josefsabl:
Annotations that are indented by tab are not processed.

```
/**
 *     @ODM\Boolean
 * previous line is indented by tab
 */
public $thisWillNotBePersisted;
```

Directed from here: https://github.com/doctrine/mongodb-odm/issues/1337

poldridge:
This only affected the first annotation in a DocBlock.

I'm not sure if this is a related issue or even an issue at all, but DocBlocks also don't seem to need * on each line. This DocBlock is fully functional:

        /** 
     @Route("/review")
    @Method("POST")
         */

PR #70, opened 2016-01-14
Allow tab character before first annotation in DocBlock

poldridge:
https://github.com/doctrine/annotations/issues/68

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -348,7 +348,7 @@ final class DocParser
         // search for first valid annotation
         while (($pos = strpos($input, '@', $pos)) !== false) {
             // if the @ is preceded by a space or * it is valid
-            if ($pos === 0 || $input[$pos - 1] === ' ' || $input[$pos - 1] === '*') {
+            if ($pos === 0 || $input[$pos - 1] === ' ' || $input[$pos - 1] === "\t" || $input[$pos - 1] === '*') {
                 return $pos;
             }
 
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -347,7 +347,7 @@ final class DocParser
 
         // search for first valid annotation
         while (($pos = strpos($input, '@', $pos)) !== false) {
-            // if the @ is preceded by a space or * it is valid
+            // if the @ is preceded by a space, tab, or * it is valid
             if ($pos === 0 || $input[$pos - 1] === ' ' || $input[$pos - 1] === "\t" || $input[$pos - 1] === '*') {
                 return $pos;
             }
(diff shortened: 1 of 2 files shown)


kimhemsoe:
Could you add a tests to go this?

poldridge:
Okay, I added a test.

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -1239,6 +1239,21 @@ public function testTrailingCommaIsAllowed()
...
+    {
+        $docblock = <<<DOCBLOCK
+/**
+ *	@Name
+ */
+DOCBLOCK;
+
+        $parser = $this->createTestParser();
+        $result = $parser->parse($docblock);
+        $this->assertEquals(1, count($result));
```
Ocramius: `assertCount(1, $result)`

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -1239,6 +1239,21 @@ public function testTrailingCommaIsAllowed()
...
+        $annot = $result[0];
+        $this->assertTrue($annot instanceof Name);
```
Ocramius: `self::assertInstanceOf('Name', reset($result))`

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -348,7 +348,7 @@ private function findInitialTokenPosition($input)
         // search for first valid annotation
         while (($pos = strpos($input, '@', $pos)) !== false) {
             // if the @ is preceded by a space or * it is valid
```
Ocramius: Comment needs updating as well

poldridge:
@Ocramius While I agree using self with the scope resolution operator is more correct than `$this` for accessing a static method, it isn't used in any of the other tests. As far as I can tell `$this->assertX` is used throughout all the tests, and `self::assert` is never used; so I kept the same style for consistency reasons.

Ocramius:
@poldridge agree on the style, but at least you shouldn't `assertTrue($expression)`, and instead do `assertInstanceOf` there ;-)

poldridge:
@Ocramius Okay, I'll update it to that.

Note though that `assertTrue($x instanceof Y)` occurs 25 times in DocParserTest.php alone. If there is consensus that `$this->assertTrue($expression);` statements should be replaced with the more semantic `self::assertX($var)`, perhaps a new issue should be created for updating that in every test.

Ocramius:
@poldridge we'll probably do that during a cleanup when aiming at 2.x :-)

alvinbunk:
I also noticed that for annotations within the comment block, if the 1st line of the comment block does not have "/***", and has "/**" instead, the annotation doesn't get recognized. I'm not sure if this is a problem?

For example, this works:

```
/**
 * @Assert\NotBlank()
 */
```

But this doesn't:

```
/*
 * @Assert\NotBlank()
 */
```

Ocramius:
That is normal/expected behavior: without `/**`, it is just a comment, and php's reflection won't retrieve it anyway.

alvinbunk:
Thank you for letting me know @Ocramius !

Ocramius:
Moved to #99

PR #75, opened 2016-03-20
Annotations 2.0

FabioBatSilva:
- Move namespace to `Doctrine\Annotations` (Removing `Common`)
- Uses `hoa/compiler` instead of `doctrine/lexer`  ( see : [grammar](https://github.com/FabioBatSilva/annotations/blob/2.0/src/Parser/grammar.pp) )
- Drop `AnnotationRegistry` and all autoload magic
- Drop `Attribute/Attributes` annotations
- Drop `SimpleAnnotationReader`
- Drop `FileCacheReader`
- Drop `IndexedReader`
- Requires php 7

TODO:

Local reviews (checkout + run locally):
- [ ] @guilhermeblanco 
- [ ] @beberlei 
- [ ] @Hywan  
- [ ] @ocramius 
- [ ] @asm89
- [ ] @deeky666
- [ ] @alcaeus
- [ ] run it against the doctrine/doctrine2 test-suite
- [ ] run it against the doctrine/mongodb-odm test-suite


schmittjoh:
How does it affect userland? Any benefits/goals?

composer.json
```diff
@@ -10,22 +10,35 @@
...
         {"name": "Benjamin Eberlei", "email": "kontakt@beberlei.de"},
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"},
-        {"name": "Johannes Schmitt", "email": "schmittjoh@gmail.com"}
+        {"name": "Johannes Schmitt", "email": "schmittjoh@gmail.com"},
+        {"name": "Fabio B. Silva", "email": "fabio.bat.silva@gmail.com"}
     ],
     "require": {
-        "php": ">=5.3.2",
-        "doctrine/lexer": "1.*"
+        "php":          ">=7.0.0",
```
stof: please don't align constraints. It leads to a nightmare when adding new constraints (merge conflicts for nothing due to alignment changes)

composer.json
```diff
@@ -10,22 +10,35 @@
...
+        "instaclick/coding-standard"            : "~1.1",
+        "squizlabs/php_codesniffer"             : "~1.5",
+        "satooshi/php-coveralls"                : "~1.0",
+        "instaclick/object-calisthenics-sniffs" : "dev-master",
+        "instaclick/symfony2-coding-standard"   : "dev-remaster"
     },
     "autoload": {
-        "psr-0": { "Doctrine\\Common\\Annotations\\": "lib/" }
+        "psr-4": {
+            "Doctrine\\Annotations\\": "src/"
```
stof: couldn't we keep the same namespace to make the change less disruptive for userland (possibility to keep using the same API) ?

src/Parser/PhpParser.php
```diff
@@ -75,10 +75,11 @@ private function getFileContent($filename, $lineNumber)
             return null;
         }
 
-        $content = '';
         $lineCnt = 0;
-        $file = new SplFileObject($filename);
-        while (!$file->eof()) {
+        $content = '';
+        $file    = new SplFileObject($filename);
```
stof: I suggest not aligning ``=`` signs. This creates useless git conflicts when adding new variables, and makes ``git blame`` much less useful by filling the history with meaningless changes on lines unrelated to the actual change of the commit

src/Parser/grammar.pp
```diff
@@ -0,0 +1,90 @@
+%skip   space               [\x20\x09\x0a\x0d]+
+%skip   doc_                [/**]
+%skip   _doc                [*/]
+%skip   star                [*]
+
+%token  at                  @                           -> annot
+%token  text                [^@].*
+
+%token  annot:identifier    [\\a-zA-Z_][\\a-zA-Z0-9_]*  -> values
```
stof: Should be ``\\?[a-zA-Z_][a-zA-Z0-9_](\\[a-zA-Z_][\\a-zA-Z0-9_])*`` actually. Any namespace separator must be followed by a non-digit char (and not by a second separator).
And btw, this regex is more restrictive that the regex of valid PHP class name. Is it on purpose to forbid unicode chars in class names of annotations ?

src/Parser/grammar.pp
```diff
@@ -0,0 +1,90 @@
...
+%token  value:identifier    [\\a-zA-Z_][\\a-zA-Z0-9_]*
+%token  value:brace_        {
+%token  value:_brace        }
+%token  value:colon         :
+%token  value:comma         ,
+%token  value:equals        =
+%token  value:number        \-?(0|[1-9]\d*)(\.\d+)?([eE][\+\-]?\d+)?
+
+%token  value:quote_        "        -> string
+%token  string:string       [^"]+
```
stof: what about quote escaping for the case where the string needs to contain a quote ?

tests/AbstractReaderTest.php
```diff
@@ -0,0 +1,492 @@
...
+use Doctrine\AnnotationsTests\TestCase;
+
+use Doctrine\AnnotationsTests\Fixtures\Reader\TestParentClass;
+use Doctrine\Annotations\Annotation\IgnoreAnnotation;
+use Doctrine\Annotations\AnnotationReader;
+use Doctrine\Annotations\Annotation;
+use Doctrine\Annotations\Reader;
+use ReflectionClass;
+
+require  __DIR__ . '/Fixtures/Reader/TopLevelAnnotation.php';
```
stof: why not relying on autoloading ?

composer.json
stof: btw, the namespace change (and so the file renaming) also means that many changed files are actually considered as being removed and added in the new location, making it harder to see actual changes (and so harder to understand what are the changes in the tests for instance)

stof:
> Drop Attribute/Attributes annotations

Why is it dropped ?

src/Configuration.php
```diff
@@ -0,0 +1,198 @@
...
+    /**
+     * @param \Doctrine\Annotations\Metadata\MetadataFactory $factory
+     */
+    public function setMetadataFactory(MetadataFactory $factory)
+    {
+        $this->metadataFactory = $factory;
+    }
+
+    /**
+     * @return \Doctrine\Annotations\Metadata\MetadataFactory
```
yourwebmaker: Sometimes you're using FQCN and sometimes you're using alias. It'd be better to use only one.

composer.json
guilhermeblanco: Majors are for BC breaks, no?

src/Configuration.php
guilhermeblanco: @fabiobatsilva keep fqcn

src/Parser/grammar.pp
guilhermeblanco: Your suggested change is wrong because you're forcing to have 2 chars before ns separator

tests/ConfigurationTest.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Doctrine\AnnotationsTests;
+
+use Doctrine\Annotations\Configuration;
+
+class ConfigurationTest extends TestCase
+{
+    public function testConfigurationAccessors()
+    {
+        $metadata  = $this->getMock('Doctrine\Annotations\Metadata\MetadataFactory', [], [], '', false);
```
yourwebmaker: Why not ``::class`` for everything ?

composer.json
stof: Indeed. But breaking BC just because you can makes it harder for anyone using your library, and means that they will take much more time to upgrade to the newer version (and ask you to continue providing support for the older version in parallel for much longer).

Rewriting all internals of the library to use different dependencies may be worth it (but we still wait for benefits/goals asked by @schmittjoh), and breaking BC to use PHP 7 return typehints is something I can agree with (it will make it impossible for people to support for 1.x and 2.x if they are extending Doctrine classes, but they should not need to use inheritance on them IMO anyway). But being friendlier to consumers of the library would be good.
Due to the PHP 7 requirement, I can guarantee you that Symfony cannot switch to the v2 only until Symfony 4 (as Symfony 3.x still supports PHP 5.x). This gives us 2 options:
- supporting 1.x only
- supporting 1.x and 2.x at the same time (letting people using PHP 7 upgrade to doctrine/annotations 2.x in their projects)

The second case is better (as the version supported by Symfony also impacts all projects build on top of Symfony, and their compatibility with other packages using doctrine/annotations), but only if we can easily use both. Currently the only changes in the Reader interface are its namespace and scalar typehints. adding scalar typehints does not hurt us (as long as we always call it properly by passing a string, the scalar typehint change does not impact the caller. The signature change only impacts people wanting to implement the interface). the namespace change however makes it a huge pain, as we don't have an interface anymore, but 2 different interfaces, and so we cannot typehint anymore when supporting both.
I don't see any benefit in the class renaming, and only drawbacks here (btw, this is why Twig still does not use namespaces for instance, and won't use them either in 2.x: the BC break would split the community using it without bringing any real benefit)

src/Parser/grammar.pp
stof: sorry, a star is missing in my suggestion. should be ``\\?[a-zA-Z_][a-zA-Z0-9_]*(\\[a-zA-Z_][\\a-zA-Z0-9_])*``

tests/ConfigurationTest.php
guilhermeblanco: Yeeeeesss!!!!!  (Yoda style)

tests/ContextTest.php
```diff
@@ -0,0 +1,45 @@
...
+class ContextTest extends TestCase
+{
+    public function testContextAccessors()
+    {
+        $class        = new \ReflectionClass('Doctrine\AnnotationsTests\Fixtures\Controller');
+        $namespaces   = ['Doctrine\AnnotationsTests\Fixtures'];
+        $ignoredNames = ['todo' => true];
+        $imports      = [
+            'template' => 'Doctrine\AnnotationsTests\Fixtures\Annotation\Template',
+            'Route' => 'Doctrine\AnnotationsTests\Fixtures\Annotation\Route;'
```
yourwebmaker: ``;`` remaining

tests/ConfigurationTest.php
stof: I suggest doing this change in the whole testsuite, but in a separate PR. This will allow to review the v2 rewriting of the library without having all these changes in the diff, making it easier to spot what actually changed in the testsuite due to the rewriting (allowing to spot BC breaks).

src/AnnotationReader.php
```diff
@@ -0,0 +1,273 @@
...
+ * A reader for docblock annotations.
+ *
+ * @author Jonathan Wage <jonwage@gmail.com>
+ * @author Benjamin Eberlei <kontakt@beberlei.de>
+ * @author Roman Borschel <roman@code-factory.org>
+ * @author Johannes M. Schmitt <schmittjoh@gmail.com>
+ * @author Fabio B. Silva <fabio.bat.silva@hotmail.com>
+ * @author Guilherme Blanco <guilhermeblanco@hotmail.com>
+ */
+class AnnotationReader implements Reader
```
lcobucci: Could this class also be final?

src/Builder.php
```diff
@@ -0,0 +1,246 @@
...
+    public function __construct(Resolver $resolver, MetadataFactory $metadataFactory)
+    {
+        $this->resolver        = $resolver;
+        $this->metadataFactory = $metadataFactory;
+    }
+
+    /**
+     * @return Doctrine\Annotations\Resolver
+     */
+    public function getResolver() : Resolver
```
lcobucci: Is this getter really needed? I mean, it's a bit strange a `Builder` exposing a way to retrieve it's resolver.

If we take a look on `DocVisitor` we might be able to find where it's being used and it seems a LoD violation to me.

src/Exception/AnnotationException.php
```diff
@@ -0,0 +1,54 @@
...
+ * @author Roman Borschel <roman@code-factory.org>
+ * @author Fabio B. Silva <fabio.bat.silva@hotmail.com>
+ * @author Guilherme Blanco <guilhermeblanco@hotmail.com>
+ */
+class AnnotationException extends \Exception
+{
+    /**
+     * @return AnnotationException
+     */
+    public static function optimizerPlusSaveComments()
```
lcobucci: We could add a return type hint for those methods :smile:

src/Metadata/MetadataFactory.php
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * @param \ReflectionClass $class
+     * @param array            $annotations
+     *
+     * @return bool
+     */
+    private function isAnnotation(ReflectionClass $class, array $annotations) : bool
+    {
+        if ($class->isSubclassOf('Doctrine\Annotations\Annotation')) {
```
lcobucci: Add `::class` also in this class.

src/Parser/PhpParser.php
```diff
@@ -27,7 +29,7 @@
  * @author Fabien Potencier <fabien@symfony.com>
  * @author Christian Kaps <christian.kaps@mohiva.com>
  */
-final class PhpParser
+class PhpParser
```
lcobucci: What's the intent of removing the `final` modifier here?

.travis.yml
```diff
@@ -1,18 +1,26 @@
...
+    - ./vendor/bin/phpunit -v --coverage-clover ./build/logs/clover.xml
+
+after_script:
+    - ./vendor/bin/coveralls -v
 
-before_script:
-    - composer install -n
+matrix:
+    allow_failures:
+        - php: hhvm
```
localheinz: @FabioBatSilva 

How about allowing builds to finish quickly?

:information_desk_person: For reference see https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing.

.travis.yml
```diff
@@ -1,18 +1,26 @@
...
-  - 5.4
-  - 5.5
-  - 5.6
-  - 7
-  - hhvm
+    - 7
+    - hhvm
+
+install:
+    - composer self-update
```
localheinz: @FabioBatSilva 

How about updating `composer` itself in the `before_install` section?

:information_desk_person: For reference see https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle.

.travis.yml
```diff
@@ -1,18 +1,26 @@
...
+    - composer --prefer-source install
```
localheinz: @FabioBatSilva 

If you actually want to cache dependencies installed with composer between builds, they need to be installed with `--prefer-dist`.

composer.json
mickaelandrieu: > Due to the PHP 7 requirement, I can guarantee you that Symfony cannot switch to the v2 only until Symfony 4

It's ok... I mean, the world don't turn around Symfony framework full edition, and any developper is able to customize his dependencies

src/Parser/grammar.pp
```diff
@@ -0,0 +1,90 @@
...
+%token  string:_quote       "        -> value
+
+#dockblock:
+    (comments() | annotations())*
+
+#annotations:
+    annotation() ( annotation() )*
```
Hywan: Can be simplified as `annotation()+`.

src/Parser/grammar.pp
```diff
@@ -0,0 +1,90 @@
...
+
+#annotation:
+    ::at:: identifier() ( parameters() | comments() )?
+
+#comments:
+    text() ( text() )*
```
Hywan: Can be simplified as `text()+`.

src/Parser/grammar.pp
```diff
@@ -0,0 +1,90 @@
...
+
+#values:
+    value() ( ::comma:: value() )* (::comma::)?
```
Hywan: No need for a group if you have a single element in it. Thus:
```diff
- (::comma::)?
+ ::comma::?
```
would still be valid.

src/Parser/grammar.pp
```diff
@@ -0,0 +1,90 @@
...
+
+#map:
+    ::brace_:: pairs() (::comma::)? ::_brace::
+
+#list:
+    ::brace_:: ( (value() ( ::comma:: value() )*) (::comma::)? )? ::_brace::
+
+#pairs:
+    pair() ( ::comma:: pair() )*
```
Hywan: An empty list of pairs is not possible then. Is it what the expected behavior? If no, `pairs` should be called with `?`.

Hywan: I assume the trailing comma can be present with all pairs, so we can add `::comma::?` here and remove it in `map`.

Hywan: Also, some languages adopt `::comma::*` (several trailing commas). I don't like this personnally, but I would like to mention it.

src/Parser/grammar.pp
```diff
@@ -0,0 +1,90 @@
...
+    text() ( text() )*
+
+#values:
+    value() ( ::comma:: value() )* (::comma::)?
+
+#map:
+    ::brace_:: pairs() (::comma::)? ::_brace::
+
+#list:
+    ::brace_:: ( (value() ( ::comma:: value() )*) (::comma::)? )? ::_brace::
```
Hywan: An empty list is possible, but not an empty list of pairs.

src/Parser/grammar.pp
```diff
@@ -0,0 +1,90 @@
...
+    ::at:: identifier() ( parameters() | comments() )?
+
+#comments:
+    text() ( text() )*
+
+#values:
+    value() ( ::comma:: value() )* (::comma::)?
+
+#map:
+    ::brace_:: pairs() (::comma::)? ::_brace::
```
Hywan: Can be simplified in:
```diff
- (::comma::)?
+ ::comma::?
```

src/Parser/grammar.pp
```diff
@@ -0,0 +1,90 @@
...
+    pair() ( ::comma:: pair() )*
+
+#pair:
+    (identifier() | string() | number() | constant()) ( ::equals:: | ::colon:: ) value()
+
+#value:
+    <boolean> | <null> | string() | map() | list() | number() | pair() | annotation() | constant()
+
+parameters:
+    ( ::parenthesis_:: ( values() )? ::_parenthesis:: ) | ( string() )?
```
Hywan: Empty list of parameters is still valid, so I reckon the expected behavior for list of pairs is to accept empty ones.

Hywan: Again:
```diff
- ( values() )?
+ values()?
```
and
```diff
- ( string() )?
+ string()?
```

Hywan: Also the parenthesis around `::parenthesis_:: … ::_parenthesis::` are not useful, but it helps the reading so it's fine to keep them.

src/Parser/BaseVisitor.php
```diff
@@ -0,0 +1,246 @@
...
+            return $result;
+        }
+
+        if ($id === '#values') {
+            $values = [];
+
+            foreach ($element->getChildren() as $child) {
+                $result = $child->accept($this, $handle, $eldnah);
+                $values = $values + $result;
+                // array_merge won`t preserve numeric keys
```
Hywan:
```diff
- won`t
+ won't
```

src/Parser/DocParser.php
```diff
@@ -0,0 +1,86 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\Annotations\Parser;
+
+use Doctrine\Annotations\Context;
+use Doctrine\Annotations\Builder;
+use Doctrine\Annotations\Exception\ParserException;
+
+use Hoa\Compiler\Llk\Llk;
```
Hywan: I might be wrong but these two declarations seem to not be used.

src/Parser/DocParser.php
```diff
@@ -0,0 +1,86 @@
...
+use Hoa\File\Read;
```
Hywan: This is the second declaration.

Hywan: However, you can add `use Hoa\Compiler\Exception;` and then…

src/Parser/DocParser.php
```diff
@@ -0,0 +1,86 @@
...
+     * @return array
+     */
+    public function parse(string $docblock, Context $context)
+    {
+        try {
+            $visitor = new DocVisitor($context, $this->builder, $this->ignoreNotImported);
+            $result  = $this->parser->parseDockblock($docblock, $visitor);
+
+            return $result;
+        } catch (\Hoa\Compiler\Exception $e) {
```
Hywan: … use `Exception $e` here.

src/Parser/DocVisitor.php
```diff
@@ -0,0 +1,97 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Annotations\Parser;
+
+use Doctrine\Annotations\Exception\ClassNotFoundException;
+use Doctrine\Annotations\Reference;
+use Doctrine\Annotations\Builder;
+use Doctrine\Annotations\Context;
+
+use Hoa\Visitor\Visit;
```
Hywan: Again, it seems like these two declarations are not used.

composer.json
stof: @mickaelandrieu any project based on Symfony would still have to fulfill the dependency constraint set by Symfony, which would still mean using doctrine/annotations 1.x if Symfony cannot support both.
And we are not the only framework which still support PHP 5.x for some times, so we won't be the only ones with this issue. Symfony is just the one for which I can guarantee the policy. Any library depending on doctrine/annotations will have to do a similar choice.

src/Metadata/MetadataFactory.php
stof: @lcobucci see my previous comment. Review will be much easier if the switch to ``::class`` is done in a separate PR

.travis.yml
stof: you should just remove HHVM entirely for now, as it cannot work: they provide a PHP 5.6 runtime, not a PHP 7 one currently, meaning they cannot install deps

composer.json
localheinz: @FabioBatSilva 

Aligning constraints tells us you're manually editing `composer.son`, which should not be necessary when requiring dependencies.

composer.json
```diff
@@ -10,22 +10,35 @@
...
     "require-dev": {
-        "doctrine/cache": "1.*",
-        "phpunit/phpunit": "4.*"
+        "doctrine/cache"                        : "1.*",
+        "phpunit/phpunit"                       : "5.*",
+        "instaclick/coding-standard"            : "~1.1",
+        "squizlabs/php_codesniffer"             : "~1.5",
+        "satooshi/php-coveralls"                : "~1.0",
+        "instaclick/object-calisthenics-sniffs" : "dev-master",
+        "instaclick/symfony2-coding-standard"   : "dev-remaster"
```
localheinz: @FabioBatSilva 

How about keeping packages sorted? It's as easy as running

```
$ composer require --dev --sort-packages instaclick/symfony2-coding-standard:dev-remaster
```

for example.

composer.json
mickaelandrieu: Afaik Symfony project claims his PHP7 compatibility: no be part of Symfony "metapackage" edition "by default" is not an issue, right ?

> any project based on Symfony would still have to fulfill the dependency constraint set by Symfony,

strongly disagree: constraints depends on library needs, not on framework ecosystem (except if you plan to move `doctrine/annotations` into Symfony organisation?)

stof: @mickaelandrieu the symfony/symfony package depends on doctrine/annotations (well, on ``doctrine/common`` for now but we should switch to more specific components). So any project wanting to use symfony will also have to use a version compatible with the symfony requirement. This is what I'm telling here. It is just basic of how dependency resolution works.
And this is not specific to Symfony. It is true for *any package depending on your library*. I took the example of Symfony here, because it is a package with a big impact. I could have taken any of the 359 packages listed on https://packagist.org/packages/doctrine/annotations/dependents (btw, Zend Framework is also impacted), but https://packagist.org/packages/1of0/better-annotations would not have been as good as example as it only have 9 installs...

If you want to make it very hard to upgrade to doctrine/annotations 1.x, fine for you, but be aware that this means forcing a hard split of the packages relying on your library, between packages wanting 1.x and packages wanting 2.x (which cannot be installed at the same time), as they won't be able to easily accept both.
And the **only** thing forcing this hard split is the namespace change, as said previously. Without it, most packages will be able to support both 1.x and 2.x without changing any line of their code. This would avoid splitting the community of dependants.
There is no benefit for users in the class renaming, but there is a huge cost. And btw, there is also a maintenance cost as any library not able to upgrade to 2.x only will ask you to continue maintaining your 1.x version in parallel for a long time. If supporting both versions was easy, it would be acceptable to drop support for 1.x after some time (hopefully not as soon as 2.x is released though, but for instance at the end of the official support of PHP 5.6), as anyone running their project on PHP 7 would already be using the 2.x maintained version.

Majors versions are indeed the time where BC breaks are allowed, but each BC break are a cost on the upgrade path (and on maintenance of the library due to the slower adoption of hard upgrades). So breaking BC just because you can, without an actual benefit, is not a good idea. A good read about that are the rules set by the PHP team for the PHP 7 version. They avoided any BC break which would cause more pains than what it was worth (so no signature change for needle/haystack vs haystack/needle for instance), precisely to make adoption of PHP 7 easier. This is also well explained by the Facebook team, which recognize this points in their official policy for React: http://facebook.github.io/react/blog/2016/02/19/new-versioning-scheme.html#avoiding-the-major-cliff (there is nothing specific to React or NPM in the problem here, you could substitute any package name and any semver-based tool to apply it elsewhere, and this is actually also the policy applied by Symfony for Symfony 3)

mickaelandrieu: > Due to the PHP 7 requirement, I can guarantee you that Symfony cannot switch to the v2 only until Symfony 4

Don't get me wrong @stof, I'm just saying that we shouldn't maintain a new library for deprecated PHP version only to follow the Symfony requirements.

> And the only thing forcing this hard split is the namespace change, as said previously.

What is the "best" time to reorder namespace, if not when we introduce a new major version ? Symfony did it too, even before Symfony 3 (for Form component between 2.4 and 2.5, introducing a new API).

Also, what is the gain of this namespace renaming here ?

beberlei: Annotations is quite stable, we can continue to maintain the 1.x branch by doing almost nothing :-)

stof: > Don't get me wrong @stof, I'm just saying that we shouldn't maintain a new library for deprecated PHP version only to follow the Symfony requirements.

This is not what I asked



> What is the "best" time to reorder namespaces, if not when we introduce a new major version ? Symfony did it too, even before Symfony 3 (for Form component between 2.4 and 2.5, introducing a new API).

The best way to make a class renaming is to have a minor version supporting both names, and then removing the deprecated class in the next major. this would allow supporting both major at the same time (as the new API would be available in latest versions of the previous major). Please read the article I linked above. This is precisely what it explains.

> Also, what is the gain of this namespace renaming here ?

This is the real question. AFAIK, there is no benefit (except making it obvious that this class is not in ``doctrine/common`` itself anymore, but installing ``doctrine/common`` installs ``doctrine/annotations```too anyway) and only drawbacks.

> Annotations is quite stable, we can continue to maintain the 1.x branch by doing almost nothing :-)

That's not the only question. Making it very hard to support both majors at the same time (again, the only thing making it hard here is the namespace renaming) means that you expect the whole PHP community to upgrade their packages to doctrine/annotations 2.x at the same time (and this can only be done after they all drop support of PHP 5.x, which means it won't happen soon).
My whole point here is to make incremental migrations possible by allowing other packages to depend on ``^1.3 || ^2.0`` for most of them.

Ocramius: Folks, please bring the deps discussion into a new issue, if necessary.

While it is true that a major bump will cause issues for dependant
packages, we will deal with polyfilling once the cleanup by @FabioBatSilva
is done.

Focus on the design problems inside the patch, for now, and focus on what
you'd like the package to be, as if it had no dependents at all.

In composer.json
<https://github.com/doctrine/annotations/pull/75#discussion_r57308250>:

>      },
>      "autoload": {
> -        "psr-0": { "Doctrine\\Common\\Annotations\\": "lib/" }
> +        "psr-4": {
> +            "Doctrine\\Annotations\\": "src/"

Don't get me wrong @stof <https://github.com/stof>, I'm just saying that we
shouldn't maintain a new library for deprecated PHP version only to follow
the Symfony requirements.

This is not what I asked

What is the "best" time to reorder namespaces, if not when we introduce a
new major version ? Symfony did it too, even before Symfony 3 (for Form
component between 2.4 and 2.5, introducing a new API).

The best way to make a class renaming is to have a minor version supporting
both names, and then removing the deprecated class in the next major. this
would allow supporting both major at the same time (as the new API would be
available in latest versions of the previous major). Please read the
article I linked above. This is precisely what it explains.

Also, what is the gain of this namespace renaming here ?

This is the real question. AFAIK, there is no benefit (except making it
obvious that this class is not in doctrine/common itself anymore, but
installing doctrine/common installs doctrine/annotations`too anyway) and
only drawbacks.

Annotations is quite stable, we can continue to maintain the 1.x branch by
doing almost nothing :-)

That's not the only question. Making it very hard to support both majors at
the same time (again, the only thing making it hard here is the namespace
renaming) means that you expect the whole PHP community to upgrade their
packages to doctrine/annotations 2.x at the same time (and this can only be
done after they all drop support of PHP 5.x, which means it won't happen
soon).
My whole point here is to make incremental migrations possible by allowing
other packages to depend on ^1.3 || ^2.0 for most of them.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly or view it on GitHub
<https://github.com/doctrine/annotations/pull/75/files/393ca1de927f97ecb4ae55a2c2e780e2b9fd5e0d#r57308250>

stof: Well, the namespace renaming is currently part of this PR.
Bringing the discussion in a separate issue can only happen if the namespace renaming is reverted in this PR.

the **only** issue for dependants is the namespace change currently (there are hard BC breaks for libraries using inheritance due to additional typehints everywhere, but I'm totally fine with that as it probably impacts less than 1% of dependants)

tests/AbstractReaderTest.php
FabioBatSilva: ``TopLevelAnnotation`` it's not PSR compliant.

composer.json
FabioBatSilva: Having sub namespace  ``Common`` only made sense when annotations were part of common.
Now its the time to remove it !

Remember that ``sed`` is our friend :)
```bash
sed 's/Doctrine\\Common\\Annotations/Doctrine\\Annotations/g' src/*
```

stof: @FabioBatSilva sed is not a friend to write code supporting both 1.x and 2.x at the same time. Your namespace change means you force the whole PHP community to upgrade to the new version at the same time (during the migration, there would be a netsplit in the community: migrated packages and not migrated packages being incompatible).
And the fact that the new version requires PHP 7 means that this split will live for a long time, as libraries can migrate to force using ``doctrine/annotations`` 2.x only when they themselves decide to drop support for PHP 5.x.

The key of an easy upgrade path is the possibility to easily write code compatible with both major versions of the library for most dependants.

But if you don't care about the fact that 359 open-source packages depend directly on your library (and so thousands of projects depend on it), please tell it explicitly and I will stop arguing about this point (but I can assure you that people won't be able to migrate to doctrine/annotations 2.x soon in that case for many of them)

schmittjoh: @stof I guess one can introduce a compatibility layer so that you would not use the annotations library directly, but just the compatibility layer:

```php
interface MyAnnotationReader { }
class Doctrinev1Adapter implements MyAnnotationReader { }
class Doctrinev2Adapter implements MyAnnotationReader { }

// just code against MyAnnotationReader in Symfony
```

I'm still missing why one would want to use version 2 as proposed here at all. Is it faster? Does it have more features? Why would I want to use this over v1?

FabioBatSilva: @schmittjoh the main goal here is to drop our parser and also some code cleanup.
But those are mainly internal improvements with very little userland affect.

FabioBatSilva: @stof we won't force anyone to migrate to 2.x,
And we are probably going continue to maintain 1.x for a while.

Supporting both 1.x and 2.x it's not a very complicated problem,
Having a adapter layer will do the job just fine.

People can pick and choose their dependencies,
and eventually most people are going to end up using PHP7 anyways.

If we are not willing to do some cleanup or upgrade to PHP7 we might as well drop 2.0 and continue with 1.0

src/AnnotationReader.php
```diff
@@ -0,0 +1,279 @@
...
+     * @var \Doctrine\Annotations\Parser\MetadataParser
+     */
+    private $metadataParser;
+
+    /**
+     * Constructor.
+     *
+     * @param \Doctrine\Annotations\Configuration $config
+     */
+    public function __construct(Configuration $config = null)
```
guilhermeblanco: We're discussing the approach to consider `Configuration` instances immutable.
In order to achieve this, we'd take all dependencies of `Configuration` and assign to properties inside of class, preventing further changes to change the way classes that take `Configuration` behave.

The only dependency needed outside of the ones already assigned here is the `ignoredAnnotationNames`. Add this as a property of this class and remove the `config` property.

src/AnnotationReader.php
```diff
@@ -0,0 +1,279 @@
...
+     */
+    public function getClassAnnotations(ReflectionClass $class) : array
+    {
+        $docblock  = $class->getDocComment();
+        $namespace = $class->getNamespaceName();
+        $imports   = $this->getClassImports($class);
+        $ignored   = $this->getIgnoredAnnotationNames($class);
+        $context   = new Context($class, [$namespace], $imports, $ignored);
+
+        if ($docblock === false) {
```
guilhermeblanco: We should prevent all these local variables to be even instantiated if we realize that `$docblock` is empty. Here is a better stub code:

```php
if (($docblock = $class->getDocComment()) === false) {
    return [];
}
```

src/AnnotationReader.php
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /**
+     * Retrieves imports.
+     *
+     * @param \ReflectionClass $class
+     *
+     * @return array
+     */
+    private function getClassImports(ReflectionClass $class)
```
guilhermeblanco: These methods should be moved to a separate class, responsible to handle ReflectionClass specifics that could potentially disappear if we have better PHP support (such as extracting "use"s). I'd suggest ReflectionClass, ReflectionMethod and ReflectionProperty under a Reflection namespace. Also we'd require a ReflectionClassFactory that could keep the cached imports of a given class.

Example:
```php
$this->reflectionClassFactory = new Reflection\ReflectionClassFactory();

$reflectionClass = this->reflectionClassFactory->getClass($reflClass); // instanceof Reflection\ReflectionClass
$reflectionProperty = $reflectionClass->getProperty($reflProperty); // instanceof Reflection\ReflectionProperty
$reflectionMethod = $reflectionClass->getMethod($reflMethod); // instanceof Reflection\ReflectionMethod

$reflectionClass->getImports()
$reflectionProperty->getImports()
$reflectionMethod->getImports()
```

src/Builder.php
```diff
@@ -0,0 +1,246 @@
...
+    public function __construct(Resolver $resolver, MetadataFactory $metadataFactory)
+    {
+        $this->resolver        = $resolver;
+        $this->metadataFactory = $metadataFactory;
+    }
+
+    /**
+     * @return \Doctrine\Annotations\Resolver
+     */
+    public function getResolver() : Resolver
```
guilhermeblanco: Drop this method. We're violating Law of Demeter here.
Add a method called `resolveClass(Context $context, string $className)` and make it call the internal resolver.

src/Builder.php
```diff
@@ -0,0 +1,246 @@
...
+     * @param array         $values
+     */
+    private function injectValues($annotation, Context $context, ClassMetadata $metadata, array $values)
+    {
+        $properties      = $metadata->properties;
+        $defaultProperty = $metadata->defaultProperty;
+        $propertyNames   = array_keys($metadata->properties);
+
+        foreach ($values as $property => $value) {
+
```
guilhermeblanco: Remove dummy blank line

src/Builder.php
```diff
@@ -0,0 +1,246 @@
...
+     * @return object
+     */
+    private function assertPropertyTypes(Context $context, ClassMetadata $metadata, array $values)
+    {
+        $properties      = $metadata->properties;
+        $defaultProperty = $metadata->defaultProperty;
+
+        // checks all declared attributes
+        foreach ($metadata->properties as $propertyName => $property) {
+
```
guilhermeblanco: Remove dummy blank line

src/Builder.php
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /**
+     * @param Context       $context
+     * @param ClassMetadata $metadata
+     * @param array         $values
+     *
+     * @return object
+     */
+    private function instantiate(Context $context, ClassMetadata $metadata, array $values)
```
guilhermeblanco: I'd argue the instantiation of an annotation should be moved to another class, `Instantiator`.
However, this would defeat the purpose of having a `Builder` class. Would love to hear pros/cons of eliminating this class and merging the `create()` method into `DocVisitor` .

guilhermeblanco: Also, by having this change, `Instantiator` could potentially eliminate the dependency of `Context` instance

src/CachedReader.php
```diff
@@ -0,0 +1,238 @@
...
+use Doctrine\Common\Cache\Cache;
+
+/**
+ * A cache aware annotation reader.
+ *
+ * @author Benjamin Eberlei <kontakt@beberlei.de>
+ * @author Johannes M. Schmitt <schmittjoh@gmail.com>
+ * @author Fabio B. Silva <fabio.bat.silva@hotmail.com>
+ */
+final class CachedReader implements Reader
```
guilhermeblanco: This class might disappear once we introduce cache 2.0

src/Parser/DocParser.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * @param string                        $docblock
+     * @param \Doctrine\Annotations\Context $context
+     * @param bool                          $ignoreNotImported
+     *
+     * @return array
+     */
+    public function parse(string $docblock, Context $context, bool $ignoreNotImported = false)
```
guilhermeblanco: We should consider dropping the argument `$ignoreNotImported`

src/Parser/TokenParser.php
```diff
@@ -136,14 +159,17 @@ public function parseUseStatement()
...
      * @return array A list with all found use statements.
      */
-    public function parseUseStatements($namespaceName)
+    public function parseUseStatements(string $namespaceName) : array
     {
-        $statements = array();
+        $statements = [];
+
         while (($token = $this->next())) {
+
```
guilhermeblanco: Remove dummy blank line

src/SimpleAnnotationReader.php
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * Simple Annotation Reader.
+ *
+ * This annotation reader is intended to be used in projects where you have
+ * full-control over all annotations that are available.
+ *
+ * @author Johannes M. Schmitt <schmittjoh@gmail.com>
+ * @author Fabio B. Silva <fabio.bat.silva@gmail.com>
+ */
+class SimpleAnnotationReader implements Reader
```
guilhermeblanco: We should consider not having this class anymore, since our idea of 2.0 was to make "use" (importing) the standard way of using Annotations, and not incorporating a default one.
That would likely require a big update on ORM, but that's our own problem.

Hywan:
What is the state of this PR? Do you need more help?

guilhermeblanco:
@Hywan it's gonna sit idle for a little while until we collect other requirements for Annotations 2.0...
I know some people are willing to make `{}` objects and `[]` arrays, for example, while others are considering an alternate syntax for named parameters, such as: 

```php
/**
 * @ORM\Entity {
 *     tableName = "user"
 * }
 */ 
class User {}
```

I'll have to discuss internally what we would do/support in such changes before consider this work complete.

Ocramius:
@guilhermeblanco we could also just stop the 2.x migration here, and make it fully BC compliant first, then think about new features and changes in the behavior for the next iteration

src/AnnotationReader.php
```diff
@@ -0,0 +1,227 @@
...
+    private $ignoredAnnotationNames;
+
+    /**
+     * Constructor.
+     *
+     * @param \Doctrine\Annotations\Configuration $config
+     */
+    public function __construct(Configuration $config = null)
+    {
+        $this->config                 = $config ?: new Configuration();
```
Ocramius: This private property is unused after `__construct`. This state should not be retained

FabioBatSilva:
@Hywan I would appreciate some help with quote escaping https://github.com/FabioBatSilva/annotations/blob/2.0/tests/Parser/DocParserTest.php#L1072

@Ocramius Do you know why we have ```@Name(foo="""bar""")``` instead of ```@Name(foo="\"bar"\")``` ?
( see : https://github.com/doctrine/annotations/pull/44 )

FabioBatSilva:
@doctrine/doctrinecore Could you please review this again ? 
Id like to get it merged to ```doctrine/2.0```

Ocramius:
> @Ocramius Do you know why we have ```@Name(foo="""bar""")``` instead of ```@Name(foo="\"bar"\")``` ?
 > ( see : https://github.com/doctrine/annotations/pull/44 )

Sadly, I do not know - that escaping syntax was already specified before I even joined the project...

Hywan:
@FabioBatSilva What is the informal rule about quote escaping? A pair of `"` is a quote (so `"abcde"` is a string whom value is `abcde`), and a single `"""` represents a quote (so `"ab"""c"""de"` is a string whom value is `ab"c"de`), is it correct?

FabioBatSilva:
@Hywan I believe that ``""`` represents a escaped quote.
So ``@Value("ab""c""de")`` == ``ab"c"de``
and ``@Value("ab"""c"""de")`` is invalid

I'm wondering if we should take the change and drop ``""`` in favor of ``\"``
eq : 
 ``@Value("ab\"c\"de")`` == ``ab"c"de``

Any thoughts @doctrine/doctrinecore ?

Ocramius:
Absolutely for `\"` here, but it would be a BC break, and probably also a subtle one.

guilhermeblanco:
I'm a fan of `\"` too...

Hywan:
`\"` is a classic escaping and this is very easy to get it work as a single token. For instance:
```
("|')(.*?)(?<!\\)\1
```
represents all single- or double-quote strings, where single or double escaped quotes are allowed. So `"foobar"`, `'foobar'`, `"foo'bar"`, `'foo"bar'`, `"foo\"bar"`, `'foo\'bar'` are all valids (non exaustive list).

Consequently, if you go for `"` (only double-quote), the token will look like:
```sh
"(.*?)(?<!\\)"
```
This is a naive approach but it will work.

However, if you want to keep the BC (so continue using `""` for `"`), it might be… … … Just a though. **This is not possible to have an empty string with `""`**. Funny? I don't think so. Let's continue:
```
"(.*?)(?<!")"
```
Let's try:
```php
$regex = '/"(.*?)(?<!")"/u';
$data  = [
    'pre""post',
    'pre"foobar"post',
    'pre"foo\"bar"post',
    'pre"foo""bar"post'
];

foreach ($data as $datum) {
    var_dump($datum);
    preg_match(
        $regex,
        $datum,
        $matches
    );
    print_r($matches);
    echo "\n";
}

/**
 * Will output:
 *   string(9) "pre""post"
 *   Array
 *   (
 *   )
 *   
 *   string(15) "pre"foobar"post"
 *   Array
 *   (
 *       [0] => "foobar"
 *       [1] => foobar
 *   )
 *   
 *   string(17) "pre"foo\"bar"post"
 *   Array
 *   (
 *       [0] => "foo\"
 *       [1] => foo\
 *   )
 *   
 *   string(17) "pre"foo""bar"post"
 *   Array
 *   (
 *       [0] => "foo"
 *       [1] => foo
 *   )
 */
```
It does not work because it matches the first `"`, looks behind if there is no `"`, this is valid, and it stops. We must look ahead instead, so:
```php
<?php

$regex = '/"(.*?)"(?!")/u';
$data  = [
    'pre""post',
    'pre"foobar"in"bazqux"post',
    'pre"foo\"bar"in"baz\"qux"post',
    'pre"foo""bar"in"baz""qux"post'
];

foreach ($data as $datum) {
    var_dump($datum);
    preg_match(
        $regex,
        $datum,
        $matches
    );
    print_r($matches);
    echo "\n";
}


/**
 * Will output:
 *     string(9) "pre""post"
 *     Array
 *     (
 *         [0] => ""
 *         [1] => 
 *     )
 *     
 *     string(25) "pre"foobar"in"bazqux"post"
 *     Array
 *     (
 *         [0] => "foobar"
 *         [1] => foobar
 *     )
 *     
 *     string(29) "pre"foo\"bar"in"baz\"qux"post"
 *     Array
 *     (
 *         [0] => "foo\"
 *         [1] => foo\
 *     )
 *     
 *     string(29) "pre"foo""bar"in"baz""qux"post"
 *     Array
 *     (
 *         [0] => "foo""
 *         [1] => foo"
 *     )
 */
```
Oops, it does not work neither, because it reads a first quote, it looks if there is one after, it does, so it continues, it reads the second quote, there is nothing after, so it stops. We must combine them both ;-). Here we go:
```php
<?php

$regex = '/"(.*?)(?<!")"(?!")/u';
$data  = [
    'pre""post',
    'pre"foobar"in"bazqux"post',
    'pre"foo\"bar"in"baz\"qux"post',
    'pre"foo""bar"in"baz""qux"post'
];

foreach ($data as $datum) {
    var_dump($datum);
    preg_match_all(
        $regex,
        $datum,
        $matches
    );
    print_r($matches);
    echo "\n";
}


/**
 * Will output:
 *     string(9) "pre""post"
 *     Array
 *     (
 *         [0] => Array
 *             (
 *             )
 *     
 *         [1] => Array
 *             (
 *             )
 *     
 *     )
 *     
 *     string(25) "pre"foobar"in"bazqux"post"
 *     Array
 *     (
 *         [0] => Array
 *             (
 *                 [0] => "foobar"
 *                 [1] => "bazqux"
 *             )
 *     
 *         [1] => Array
 *             (
 *                 [0] => foobar
 *                 [1] => bazqux
 *             )
 *     
 *     )
 *     
 *     string(29) "pre"foo\"bar"in"baz\"qux"post"
 *     Array
 *     (
 *         [0] => Array
 *             (
 *                 [0] => "foo\"
 *                 [1] => "in"
 *                 [2] => "qux"
 *             )
 *     
 *         [1] => Array
 *             (
 *                 [0] => foo\
 *                 [1] => in
 *                 [2] => qux
 *             )
 *     
 *     )
 *     
 *     string(29) "pre"foo""bar"in"baz""qux"post"
 *     Array
 *     (
 *         [0] => Array
 *             (
 *                 [0] => "foo""bar"
 *                 [1] => "baz""qux"
 *             )
 *     
 *         [1] => Array
 *             (
 *                 [0] => foo""bar
 *                 [1] => baz""qux
 *             )
 *     
 *     )
 */
```
It works! With `"foo""bar""`, we capture `foo""bar`. With `"foo\"bar"`, we capture `foo"`, which is correct because the escaping is ignored (we can take it into account, see bellow). With `"foobar"`, we capture `foobar`. With `""`, we capture nothing, it fails, because this is invalid.

Bonus so. Having the backslash escaping **and** the quote escaping. Taking a deep breath.
```php
<?php

$regex = '/"(.*?)(?<!["\\\])"(?!")/u';
$data  = [
    'pre""post',
    'pre"foobar"in"bazqux"post',
    'pre"foo\"bar"in"baz\"qux"post',
    'pre"foo""bar"in"baz""qux"post'
];

foreach ($data as $datum) {
    var_dump($datum);
    preg_match_all(
        $regex,
        $datum,
        $matches
    );
    print_r($matches);
    echo "\n";
}


/**
 * Will output:
 *     string(9) "pre""post"
 *     Array
 *     (
 *         [0] => Array
 *             (
 *             )
 *     
 *         [1] => Array
 *             (
 *             )
 *     
 *     )
 *     
 *     string(25) "pre"foobar"in"bazqux"post"
 *     Array
 *     (
 *         [0] => Array
 *             (
 *                 [0] => "foobar"
 *                 [1] => "bazqux"
 *             )
 *     
 *         [1] => Array
 *             (
 *                 [0] => foobar
 *                 [1] => bazqux
 *             )
 *     
 *     )
 *     
 *     string(29) "pre"foo\"bar"in"baz\"qux"post"
 *     Array
 *     (
 *         [0] => Array
 *             (
 *                 [0] => "foo\"bar"
 *                 [1] => "baz\"qux"
 *             )
 *     
 *         [1] => Array
 *             (
 *                 [0] => foo\"bar
 *                 [1] => baz\"qux
 *             )
 *     
 *     )
 *     
 *     string(29) "pre"foo""bar"in"baz""qux"post"
 *     Array
 *     (
 *         [0] => Array
 *             (
 *                 [0] => "foo""bar"
 *                 [1] => "baz""qux"
 *             )
 *     
 *         [1] => Array
 *             (
 *                 [0] => foo""bar
 *                 [1] => baz""qux
 *             )
 *     
 *     )
 */
```
It works! With `"foo""bar""`, we capture `foo""bar`. With `"foo\"bar"`, we capture `foo"bar`. With `"foobar"`, we capture `foobar`. With `""`, we still capture nothing.

So the token might be:
```
/"(.*?)(?<!["\\\])"(?!")/u
```
With this regular expression, you still ensure BC and you can transition to traditional string escaping.

Thoughts?

Failed bonus: `/(["\'])(.*?)(?<![\1\\\])\1(?!\1)/u` does not work (single or double quote, with backslash escaping and double quote escaping), maybe because we cannot use a back reference inside an look behind assertion. I can dig into it only if you would like this feature. Tell me!

Ocramius:
We'd still have a BC break for existing strings containing \"

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

Hywan:
@Ocramius That's correct. And missing both escaping could lead to unexpected behaviors. You need to choose, but you have many possible tokens now 😄.

Ocramius:
It would be a feasible switch if we provided a tool to detect annotations with `\"` and `""` in them. Doesn't need to be based on the lib: simple regex+reflection could cut it.

Hywan:
Yup. That's pretty easy to do.

FabioBatSilva:
Thanks @Hywan,

I believe that translates to something like :

```diff
-%token  value:quote_        "        -> string
-%token  string:string       [^"]+
-%token  string:_quote       "        -> value
+%token  value:string        "(.*?)(?<!\\)"

 string:
-    ::quote_:: <string> ::_quote::
+    <string>
```

Is there a easy way to parse/remove quotes and escapes of the token value ?
such as ``"foo"`` to ``foo`` and  ``"\"bar\""`` to ``"bar"``.

Or is this something that should be done with a ```Visitor``` ?

Hywan:
This is something that may be done in the visitor.

src/Builder.php
```diff
@@ -0,0 +1,236 @@
...
+                    $metadata->class,
+                    $context->getDescription(),
+                    json_encode($values)
+                ));
+            }
+
+            $annotation->{$defaultProperty} = $value;
+        }
+    }
+
```
Hywan: Too much empty lines here.

src/Parser/BaseVisitor.php
```diff
@@ -0,0 +1,287 @@
...
+
+        if ($id === '#constant') {
+            return $this->visitConstant($element, $handle, $eldnah);
+        }
+
+        if ($id === 'token') {
+            return $this->visitToken($element, $handle, $eldnah);
+        }
+
+        throw new \RuntimeException("Unknown AST node : $id");
```
Hywan: No space before `:` :smiley:.

stof:
The drawback of using ``\"`` to escape a quote is that you now have to escape ``\`` as well.
When you have only the string delimiting quotes to escape, it is common to double it instead of entering the complexity of escape sequences:
- Yaml single quoted string
- SQL strings
- maybe other systems I cannot think of now.

the grammar suggested by @FabioBatSilva is broken, because it does not handle the escaping of ``\`` itself.

Hywan:
It's not broken. You have to select everything inside the string and then unescape characters.

However, the current token does not handle this case: `"foo\\"`, which represents `foo\`. cc @FabioBatSilva

src/Parser/grammar.pp
```diff
@@ -34,9 +34,7 @@
 %token  value:equals        =
 %token  value:number        \-?(0|[1-9]\d*)(\.\d+)?([eE][\+\-]?\d+)?
 
-%token  value:quote_        "        -> string
-%token  string:string       [^"]+
-%token  string:_quote       "        -> value
+%token  value:string        "(.*?)(?<!\\)"
```
Hywan: Do we support the escaping of `\` itself?

FabioBatSilva: I think we should

composer.json
stof: @FabioBatSilva I'm not arguing at all with cleaning up the internals.
the *only* point I'm arguing with is the namespace change, which does not bring any benefit (except maybe some theoretical one), but has a huge impact on users.
If you don't do this remaining, updating to 2.0 will be a no-brainer for most users as changes are mostly internal.
The fact that a major version *can* break BC does not mean it *must* break BC anywhere. The less BC break you have, the easier the migration is (and the less time you will have to maintain 1.x in parallel with its entirely different codebase)

tests/AbstractReaderTest.php
stof: you can rely on composer classmap autoloading (or make it PSR compliant)

src/AnnotationReader.php
```diff
@@ -0,0 +1,225 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Annotations;
+
+use ReflectionClass;
+use ReflectionMethod;
+use ReflectionProperty;
+
+use Doctrine\Annotations\Parser\DocParser;
+use Doctrine\Annotations\Parser\MetadataParser;
```
lennerd: This two namespaces are not used.

src/Reader.php
```diff
@@ -17,12 +17,19 @@
...
  */
 
-namespace Doctrine\Common\Annotations;
+declare(strict_types=1);
+
+namespace Doctrine\Annotations;
+
+use ReflectionClass;
+use ReflectionMethod;
+use ReflectionProperty;
```
donquixote: Why import those root namespace classes? I always prefer to read `\ReflectionClass`. If I see `ReflectionClass`, it makes me think it is some kind of custom reflection layer.

src/Reader.php
```diff
@@ -34,7 +41,7 @@
      *
      * @return array An array of Annotations.
      */
-    function getClassAnnotations(\ReflectionClass $class);
+    public function getClassAnnotations(ReflectionClass $class) : array;
```
donquixote: I agree with adding `public ` access modifier. But would it not be better to split this into a separate PR, which gets applied before this one, to every branch?

src/AnnotationReader.php
```diff
@@ -82,7 +82,7 @@ public function getClassAnnotations(ReflectionClass $class) : array
     /**
      * {@inheritDoc}
      */
-    public function getClassAnnotation(ReflectionClass $class, $annotationName)
+    public function getClassAnnotation(ReflectionClass $class, string $annotationName)
```
donquixote: I would suggest to split decorative CS changes and BC-breaking signature changes into separate distinct commits.
Adding a type hint changes the signature, which breaks any implementing classes.

donquixote: And then, the decorative code style changes can go into a dedicated PR to the 1.x branch.

src/Builder.php
```diff
@@ -0,0 +1,234 @@
...
+    /**
+     * @var \Doctrine\Annotations\MetadataFactory
+     */
+    private $metadataFactory;
+
+    /**
+     * Constructor.
+     *
+     * @param \Doctrine\Annotations\Resolver        $resolver
+     * @param \Doctrine\Annotations\MetadataFactory $metadataFactory
```
donquixote: The type is incorrect, it should be `\Doctrine\Annotations\Metadata\MetadataFactory`.

src/Metadata/MetadataFactory.php
```diff
@@ -0,0 +1,200 @@
...
+     */
+    public function getMetadataFor(string $className)
+    {
+        if (isset($this->cache[$className])) {
+            return $this->cache[$className];
+        }
+
+        $class       = new \ReflectionClass($className);
+        $constructor = $class->getConstructor();
+        $docComment  = $class->getDocComment();
```
donquixote: Unused local variable `$docComment`.

composer.json
donquixote: I don't know all the details of this, but I agree with @stof to focus on internal changes.

I always found the global AnnotationRegistry to be ugly. Instead, the parser should produce an AST first, where annotation names are not treated as classes. Then a separate step should resolve those names as objects.

I see that the PR does this. The first step is to let the hoa parser construct an AST. Then the DocVisitor (strange name, btw) turns it into an annotation object tree. And instead of the awkward class registry, ist just checks `class_exists() || interface_exists()`. Which is ok I guess.

This part looks reasonable, so this should be the focus - imo.

donquixote:
Did anyone run a benchmark to compare the hoa compiler to the one currently in use?
Maybe it is all "fast enough" so we don't have to worry. Just asking.

src/Metadata/MetadataFactory.php
```diff
@@ -107,11 +107,11 @@ public function getMetadataFor(string $className)
      */
     private function isAnnotation(ReflectionClass $class, array $annotations) : bool
     {
-        if ($class->isSubclassOf('Doctrine\Annotations\Annotation')) {
+        if ($class->isSubclassOf(Annotation::CLASS)) {
```
donquixote: `::CLASS` should be lowercase `::class`.

donquixote:
Architecture: I propose to drop Metadata and MetadataFactory.
Instead, have one factory object per class (or give it a different name, dunno (*)).

```php
class Builder
{
    [..]

    /**
     * @param Context   $context
     * @param Reference $reference
     *
     * @return object
     */
    public function create(Context $context, Reference $reference)
    {
        $target    = $reference->nested ? Target::TARGET_ANNOTATION : $context->getTarget();
        $fullClass = $this->resolver->resolve($context, $reference->name);
        $values    = $reference->values;

        if (null === $factory = $this->factoryProvider->classGetFactory($fullClass)) {
            throw InvalidAnnotationException::notAnnotationException($fullClass, $reference->name, $context->getDescription());
        }

        return $factory->instantiate($context, $values, $target);
    }
}
```

Now all the metadata stuff can be encapsulated in the $factory object.
There can even be separate factory classes depending how the class should be constructed, and how the annotation values are transformed into arguments.
There can also be a ClassNotFoundFactory, which always throws an exception in the ->instantiate() method.
The interface for those factories really only has this one method.

(*) I was initially going to call the "factory" "instantiator". But this name already exists in `Doctrine\Instantiator\Instantiator`. So...

src/Reflection/ReflectionMethod.php
```diff
@@ -0,0 +1,105 @@
...
+            if ($trait->getFileName() !== $this->getFileName()) {
+                continue;
+            }
+
+            $methodImports = $this->phpParser->parseClass($trait);
+            $traitImports  = array_merge($traitImports, $methodImports);
+        }
+
+        return $this->imports = array_merge($classImports, $traitImports);
+    }
```
donquixote: I think this is confusing. If the method is coming from a trait, then the annotation is living in the trait file, and the imports should be from the trait's file only. If the method is declared in the class itself, then the annotation is living in the class file, and it should use the imports from the class file. I don't see a case where it should combine imports from different files.

donquixote: Here is how to find out if a method is defined in a trait, https://stackoverflow.com/a/45912866/246724.

donquixote: Btw, imo all of this reflection + inheritance adds unnecessary complexity.

Hywan:
Just a link about `hoa/compiler` performances, https://blog.hoa-project.net/2016/08-Performance-boost-for-Hoa-Compiler.html. cc @donquixote

src/Reflection/ReflectionMethod.php
stof: Well, forbidding people to use inheritance and traits when using annotations would mean that nobody would migrate to version 2

Ocramius: From a functional perspective, it should stay the exact same, maybe with some API changes, but no behavioral ones.

The point here is getting rid of our own hacky parser, using a formalised one (HOA's)

donquixote: Oh, maybe I was unclear.

> all of this reflection + inheritance adds unnecessary complexity.

What I mean is we don't need to inherit from \ReflectionClass to find the imports. Instead, have an ImportFinder or something like that.

Of course, people who write annotated classes should be allowed to use inheritance, and traits!

> From a functional perspective, it should stay the exact same, maybe with some API changes, but no behavioral ones.

If the code in the PR is replicating existing behavior, then it needs to stay this way.
Or, if we agree that the old behavior is wrong, we could have two implementations of ImportFinder or of the class name resolver: One that operates the BC way, another that operates the "correct" way.

Why would we say that "the old behavior is wrong"?

Consider this example:

File T.php:
```php
<?php
namespace Acme\Foo;
use Acme\Annotation\Hello;
trait T {
  /**
   * @Hello("I am an annotation on a trait method.")
   * @Goodbye("I am annotation on a trait method, but the import is in the class file.")
   */
  function foo() {}
}
```

File C.php:
```php
<?php
namespace Acme\Bar;
use Acme\Annotation\Goodbye;
class C {
  use T;
}
```

With the behavior proposed in the PR, which I assume is also the current behavior, the second annotation `@Goodbye(..)` will use the import `Acme\Annotation\Goodbye` from the class file.

I am saying this is wrong. It should only use the imports from the trait file. So the `@Hello(..)` should work, but the `@Goodbye(..)` should not.

This would be consistent with how the language itself works.
Imports are only available within the same file.

Well personally I think having annotations on a method in a trait is probably a bad idea anyway. but if we support it, it should at least be "correct". Unless, of course, it is for BC reasons.

> The point here is getting rid of our own hacky parser, using a formalised one (HOA's)

Which I assume will be more maintainable, more reliable, more understandable (people have to look at the grammar only). So yeah, seems like a good idea.

Personally I care more about the registry going away.

donquixote:
Thanks @Hywan !
This does not compare it with a hand-written parser, only with previous hoa parsers.

In general, a hand-written or a [generated](https://stackoverflow.com/questions/3720362/what-is-a-good-parser-generator-for-php) parser should be faster than a parser [combinator](https://github.com/Vektah/parser-combinator), or one that needs to interpret a piece of grammar in every step. Since it adds overhead and indirection to every micro-operation, the difference could be something like factor 2 (or more, or less). I remember this, because I used to experiment with the vektah parser combinator.

Your linked article, "Exporting the parser into PHP code", claims that the parser can be exported to PHP code. If this is equivalent to a generator, then it should be similarly fast as a hand-written parser.

Of course, even if it would take 2x longer, does not mean we have to care, if overall it is still "fast enough".

src/Reflection/ReflectionMethod.php
donquixote: About the methods defined in traits: It gets even more interesting if the method is renamed.

```php
Trait T {
  /**
   * @Hello()
   */
  function foo() {}
}

class C {
  use T {
    foo as bar;
  }
}

$m = new \ReflectionMethod('C', 'bar');
$reader->getMethodAnnotations($m);
```

The current behavior will not understand that the method is defined in a trait under a different name.

donquixote: And about properties in traits - this is more difficult. There is no \ReflectionProperty::getFileName().
https://stackoverflow.com/questions/18257158/how-to-extract-start-line-of-a-property-declaration-in-php

As a heuristic, we could say that:
- If none of the traits of the class has a property with the same name, then the property belongs to the class, obviously.
- If one or more of the traits define the property then we compare the doc comment. See https://3v4l.org/KY9nl.

Maybe all of this should be discussed in a separate issue. I only brought it up here because the PR affects the code where this behavior is implemented.

Ocramius: @donquixote we kinda fixed all of these horrors in [`roave/better-reflection`](https://github.com/Roave/BetterReflection), although it is not the primary aim here to provide very precise reflection of ugly stuff like traits.

stof:
I'm still against the namespace change if it does not have a migration path:

- if packages are not able to easily support both v1 and v2 of the library, it means that a project cannot migrate to v2 until all its dependencies using doctrine/annotations are migrated, and it cannot update other dependencies already migrated until that time. Given the number of packages relying on doctrine/annotations out there to parse annotations, such community split is a bad news
- just migrating to v2 is not an option for packages needing to keep support for PHP 5.x (meaning that if they cannot support both versions, they will stay on v1 forever).

So this means we need a continuous migration path if you want to keep the namespace change.
This could be done by doing a new 1.x release adding class aliases using the new namespace. See how Twig did it for instance.

Hywan:
> This does not compare it with a hand-written parser, only with previous hoa parsers.

Yes it doesn't, but it gives an overview of the last big improvements :-). I should have clarified this, sorry.

> In general, a hand-written or a generated parser should be faster than a parser combinator, or one that needs to interpret a piece of grammar in every step. Since it adds overhead and indirection to every micro-operation, the difference could be something like factor 2 (or more, or less). I remember this, because I used to experiment with the vektah parser combinator.

True and false (well, you have started your sentence with “In general” 😉). In PHP, a parser combinator might be slower than a generated parser because of function calls and indirections, but it will not be the bottleneck I guess. The real bottleneck is the data copy. In a parser combinator, you have to copy the data being parsed into each parser. Even if PHP does a COW (Copy-On-Write), each data split (`substr`) will generate a new copy, and so it's going to be slow. This particular problem can also be present in a generated parser, but the API surface is much smaller. For instance, the lexer for `hoa/compiler` does not consume the data by doing a substring, it just read it with an offset: https://github.com/hoaproject/Compiler/blob/c86ccfbce9b9cad17cf84ffdf5c505c695d83d7a/Llk/Lexer.php#L276-L282 This is highly optimized.

In a parser combinator however, the lexer and the parser phases are “merged”, so the memory peak should be smaller than in a generated parser. However, regarding the last improvement in `hoa/compiler`, the lexer now works as a buffered iterator, so the behavior is similar to a parser combinator: https://github.com/hoaproject/Compiler/blob/c86ccfbce9b9cad17cf84ffdf5c505c695d83d7a/Llk/Parser.php#L162-L165

A parser combinator is like a hand-written parser, except it has a predefined formalism, is more testable, is more re-usable etc. Compared to a generated parser, the API is larger. However, a generated parser can be seen as a parser combinator with a small API. `hoa/compiler` has only one method: `_parse`, which adapts its behavior whether it meets a token, a concatenation, a choice, or a repetition, which are [the rules](https://github.com/hoaproject/Compiler/tree/master/Llk/Rule) (the grammar description language intrinsics/constructions). One method also means a better caching by the VM, and the CPU.

This thread is not the place to debate about this, but: A generated parser, a parser combinator, or a hand-written parser can all be fast and efficient, or slow and ineffective. It really depends of how they are implemented. They all have pros and cons. I personally prefer a parser combinator when working with Rust (see [nom](https://github.com/Geal/nom/)) because it is testable and brings interesting garantees, while when working with PHP, I prefer a generated parser.


> Your linked article, "Exporting the parser into PHP code", claims that the parser can be exported to PHP code. If this is equivalent to a generator, then it should be similarly fast as a hand-written parser.

I would claim that a hand-written parser is most of the time not fast. You have to re-optimise and re-implement everything, like the lexer (a good one is not simple) and the parser with all the optimisation. And the error-management, the AST builder, the memory management, the profiling etc. It's better to have a hackable compiler toolchain I guess.

But indeed, once a `Hoa\Compiler\Llk\Parser` is compiled into PHP code, it just creates an instance of `Hoa\Compiler\Llk\Parser` directly without loading the grammar from a textual file. It builds the grammar as a set of rules, which is fast to instanciate, and cachable by the VM. The lexing and parsing in themselves are optimized and use a very small API, which is also cachable correctly by the VM.

The most obvious way to be faster now is to use really good data structures instead of generic `array`, but we are limited by the language (I want [`php-ds`](http://php.net/ds) in the core, pleaaase).


> Of course, even if it would take 2x longer, does not mean we have to care, if overall it is still "fast enough".

Correct. I don't want to speak for the Doctrine team, but my understanding of the problem is the following: Drop a hand-written, hard to maintain, hacky, and maybe buggy parser by a formal parser which is easy to maintain and fast enough. `hoa/compiler` plays this role. Also, `hoa/compiler` brings interesting algorithms to generate data from a grammar (it is called Grammar-based Testing). More resources about this:

* https://hoa-project.net/En/Literature/Hack/Compiler.html#Generation the documentation,
* https://hoa-project.net/En/Literature/Research/Amost12.pdf the research paper,
* https://keynote.hoa-project.net/Amost12/EDGB12.pdf the presentation of the research paper,
* https://mnt.io/2014/09/30/generate-strings-based-on-regular-expressions/ explaining how to generate data from a regular expression.

These algorithms can help to test the Doctrine annotations, and DQL.

donquixote:
> I would claim that a hand-written parser is most of the time not fast. You have to re-optimise and re-implement everything, like the lexer (a good one is not simple) and the parser with all the optimisation. And the error-management, the AST builder, the memory management, the profiling etc. It's better to have a hackable compiler toolchain I guess.

Maybe I should have said "hardcoded" rather than "hand-written".
And instead of "should be faster", I should have said we are comparing the fastest theoretically possible parsers of each category. If you make all the right choices, what remains is the overhead and indirection. E.g. for this same reason, a parser in C would be faster than one in PHP.
You don't need a lexer or memory management, if all you do is string index lookups, like here:https://github.com/donquixote/annotation-parser/blob/1.0/src/Parser/AnnotationParser.php. Also an AST doesn't have to be complicated.

This is not an argument against the hoa parser, just a conversation.

Hywan:
We agree 😃.

src/Parser/grammar.pp
```diff
@@ -0,0 +1,85 @@
...
+parameters:
+    ( ::parenthesis_:: ( values() )? ::_parenthesis:: ) | string()?
+
+identifier:
+    <identifier>
+
+#constant:
+    <identifier> (<colon> <colon> <identifier>)?
+
+string:
```
Hywan: If a rule recognizes only one token, then it's faster to just use this token. I assume being fast is important in this context.

Same for the `text`, `number`, and `identifier` rules.

range-of-motion:
What's going on with this? I'd very much like to see this because of multi line support.

alcaeus:
Closing this PR: there has been a second effort to create 2.0 which has been just as successful as this one. We'll be revisiting this at a later date.

Issue #79, opened 2016-03-27
AnnotationReader may incorrectly throw exceptions


guilhermeblanco:
_From @arnaud-lb on July 6, 2012 17:11_

The AnnotationReader pre-parses classes' doc block with a PreParser that doesn't have import (*use*) informations.

Due to this, it can happen that an annotation resolves to a wrong class.

The parser will then attempt to validate the class and fail (no @Annotation, cannot be used on classes, etc).

Here is a test case: https://github.com/doctrine/common/pull/161

The ArrayObject annotation is resolved as \ArrayObject, which is not a valid annotation. This results in the following exception:

    Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The class "ArrayObject" is not annotated with @Annotation. Are you sure this class can be used as annotation? If so, then you need to add @Annotation to the _class_ doc comment of "ArrayObject". If it is indeed no annotation, then you need to add @IgnoreAnnotation("ArrayObject") to the _class_ doc comment of class Doctrine\Tests\Common\Annotations\Issue159\ParseMe.

_Copied from original issue: doctrine/common#159_

guilhermeblanco:
_From @stof on July 6, 2012 17:25_

could you create a pull request instead of pasting a diff ? It will be easier for us

guilhermeblanco:
_From @arnaud-lb on July 6, 2012 17:43_

@stof done

guilhermeblanco:
_From @popofr13 on October 12, 2012 9:26_

I have a similar error on a Symfony 2.1 project :

```php
[Semantical Error] The class "Annotation" is not annotated with @Annotation. Are you sure this class can be used as annotation? If so, then you need to add @Annotation to the _class_ doc comment of "Annotation". If it is indeed no annotation, then you need to add @IgnoreAnnotation("Annotation") to the _class_ doc comment of class @Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter.
```

How can i fix it, get around ?

schlessera:
I have encountered a similar error, that is probably caused by the same reason:
```
ERROR: [Semantical Error] The annotation "@Annotation" in method <namespace>\ExtendedClass::method() was never imported. Did you maybe forget to add a "use" statement for this annotation?
```

The PHP Parser only parse the class of the current object, but it does not parse all the parent classes.

PR #89, opened 2016-03-30
Parse parent classes too while fetching imports.

schlessera:
When parsing imports for a specific class, the imports of all its parent classes need to be parsed too, or else the `AnnotationReader` might throw erroneous exceptions when methods from a parent class are called, because the import of the annotation could be ina parent class' `use` statements without appearing in the child class' `use` statement.

See #79

### DIFF
--- a/lib/Doctrine/Common/Annotations/PhpParser.php
+++ b/lib/Doctrine/Common/Annotations/PhpParser.php
@@ -58,6 +58,11 @@ final class PhpParser
 
         $statements = $tokenizer->parseUseStatements($class->getNamespaceName());
 
+        while ($parent = $class->getParentClass()) {
+            $statements = array_merge($statements, $this->parseClass($parent));
+            $class = $parent;
+        }
+
         return $statements;
     }
 
--- a/tests/Doctrine/Tests/Common/Annotations/Fixtures/ChildClassInheritingFromParentClass.php
+++ b/tests/Doctrine/Tests/Common/Annotations/Fixtures/ChildClassInheritingFromParentClass.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Doctrine\Tests\Common\Annotations\Fixtures;
+
+class ChildClassInheritingFromParentClass extends ParentClassWithAnnotatedMethod
+{
+
+}
--- a/tests/Doctrine/Tests/Common/Annotations/Fixtures/ParentClassWithAnnotatedMethod.php
+++ b/tests/Doctrine/Tests/Common/Annotations/Fixtures/ParentClassWithAnnotatedMethod.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Doctrine\Tests\Common\Annotations\Fixtures;
+
+use Doctrine\Tests\Common\Annotations\Fixtures\Annotation\Autoload;
+
+class ParentClassWithAnnotatedMethod
+{
+
+    /**
+     * @Autoload
+     */
+    public function someMethod()
+    {
+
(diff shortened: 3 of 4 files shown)


Ocramius:
> the import of the annotation could be in a parent class' `use` statements without appearing in the child class' `use` statement.

This seems wrong. Imports are per-file or per-namespace, not per-class.

schlessera:
Well, I assume that most codebases will have each class in a separate file...

Ocramius:
Right, so why should the `use` statements of different files in the inheritance be used?

schlessera:
Because it's broken otherwise. If you call an inherited method, the current parser will throw an exception on the annotation because it could not find its `use` statement within the current class, as the annotation (as well as its corresponding `use` statement) is in the parent class.

So, although the code is correct, it will not work because the parser failed to check the parent class's `use` statements.

schlessera:
See here for a use case (with example): https://github.com/goaop/framework/issues/273

Ocramius:
@schlessera the `use` statement should be imported again for the current namespace. Inherited methods are a different story.

Can you write a test case (see existing tests in this repo) demonstrating the problem, and that will pass with this patch applied?

schlessera:
I've written a simple test case that has a parent class with an annotated method (including the corresponding `use` statement), and a child class extending the parent class.

To be able to call the inherited method `someMethod()` on the child class, the PHP parser needs to be able to parse down the inheritance tree to find the `use` statement in the parent class' file, or else it will erroneously throw an exception here: https://github.com/doctrine/annotations/blob/master/lib/Doctrine/Common/Annotations/DocParser.php#L703

Ocramius:
@schlessera just read the test, and it makes sense, although I think the fix is incorrect.

Basically, what is happening here is that the wrong `ReflectionClass` instance is used when detecting annotations for a method or property: the `ReflectionProperty#getDeclaringClass()` and `ReflectionMethod#getDeclaringClass()` should be used when reading annotations for that method instead.

schlessera:
I'm not sure I'm on the right track here, I'm not really familiar with the project, so exuces me if this is rubbish...

I've done some more tests, using the `getMethodAnnotations()` & `getPropertyAnnotations()` methods from `AbstractParser`. Both of these seem to handle inheritance just fine.

So, if one uses `getDeclaringClass` through a `RefelctionProperty` or `ReflectionMethod`, everything just works.

However, this still does not address the case where one parses an entire class and expect to have usable information in the parsed results. This will break on inheritance with the current code.

I understand that the fix might incur additional and often times unnecessary overhead, but I am currently not sure how else to address this specific issue.

Is this maybe a use case that is not supposed to be supported?

Ocramius:
> I understand that the fix might incur additional and often times unnecessary overhead, but I am currently not sure how else to address this specific issue.

The idea is that you are supposed to request the docblock from the reflection method's/property's declaring class:

```php
class Foo {
    /** @SomeAnnotation */
    public function bar() {}
}

class Baz extends Foo {}

// this is actually getting annotations from `Foo#bar()`
var_dump($annotationReader->getMethodAnnotations(new ReflectionMethod(Baz::class, 'bar')));
```

If you look at [`getMethodAnnotations`](https://github.com/doctrine/annotations/blob/2e1b1f7597e2f647e99e96e747fcae9718ac817b/lib/Doctrine/Common/Annotations/AnnotationReader.php#L253), that is actually what is happening.

 > Is this maybe a use case that is not supposed to be supported?

See `getMethodAnnotations`: it is supported, but you need to target the correct class.

PR #96, opened 2016-10-24
Fix/#62 check trait and parent class ttl in annotations

Ocramius:
This is a continuation of #62

I went a bit overboard with the changes, but overall it's also a cleanup/upgrade of all deps.


Ocramius:
Thanks @iamluc for your work on the original PR!

PR #98, opened 2016-10-24
Enhancement/#65 remove cache salt from cached reader

Ocramius:
Continuation/rebase of #65


Ocramius:
@Tobion merged, thanks!

Issue #93, opened 2016-08-10
FileCacheReader putting contents in a tempfile before being allowed to write to it


func0der:
Hey there,

we have a little problem with the https://github.com/doctrine/annotations/blob/master/lib/Doctrine/Common/Annotations/FileCacheReader.php.

In its [saveCacheFile](https://github.com/doctrine/annotations/blob/master/lib/Doctrine/Common/Annotations/FileCacheReader.php#L205) method it creates a temporary and tries to write to it before setting its permission.

This is a usually not a big problem, because the created file belongs automatically to the creator in a normal filesystem.
BUT if the location where the cache is saved is a location that is a shared storage for example in a virtual machine or maybe a network storage (which is pretty similar) the user is automatically changed.

What is the reason for this seemingly inverted behavior of trying to write before actually having permission to do so for sure?

mikeSimonson:
Can you confirm that assigning rights before the file is created works ? Or the rights are just ignored at the file creation.

PR #100, opened 2016-10-26
Fix 93

mikeSimonson:
#93

Any idea how to test this @Ocramius ?

### DIFF
--- a/lib/Doctrine/Common/Annotations/FileCacheReader.php
+++ b/lib/Doctrine/Common/Annotations/FileCacheReader.php
@@ -214,6 +214,8 @@ class FileCacheReader implements Reader
             throw new \RuntimeException(sprintf('Unable to create tempfile in directory: %s', $this->dir));
         }
 
+        @chmod($tempfile, 0666 & (~$this->umask));
+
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Annotations/FileCacheReader.php
```diff
@@ -214,14 +214,14 @@ private function saveCacheFile($path, $data)
...
 
+        @chmod($tempfile, 0666 & (~$this->umask));
+
         $written = file_put_contents($tempfile, '<?php return unserialize('.var_export(serialize($data), true).');');
 
         if (false === $written) {
             throw new \RuntimeException(sprintf('Unable to write cached file to: %s', $tempfile));
         }
 
-        @chmod($tempfile, 0666 & (~$this->umask));
```
Ocramius: This should be on the existing file (last doctrine security advisory), not applicable before

mikeSimonson: @Ocramius Does that mean that the issue is present here and that #93 is non fixable ?

Ocramius: Either that, or do chmod twice. At the end of the process, the flags on the file should be `0666 & (~$this->umask)`.

PR #101, opened 2016-10-26
Test annotation parameters containing space

mikeSimonson:
Apparently this used to be a problem but I couldn't
find a test that covered this specific case.


Ocramius:
@mikeSimonson thanks!

Issue #81, opened 2016-03-27
[AnnotationReader] Traits Composed from Traits


guilhermeblanco:
_From @sofakartoffel on June 23, 2015 15:34_

The AnnotationReader does not use the imports from Traits inside of other Traits.

e.g.:
* TraitA imports some custom annotation
* TraitB uses TraitA
* EntityA uses TraitB
* the Imported Annotation in TraitA is not found 

_Copied from original issue: doctrine/common#370_

zakjakub:
Any workaround? :/

Ocramius:
@zakjakub revive https://github.com/doctrine/annotations/pull/102 (rebase it and prepare it for merge)?

zakjakub:
@Ocramius Thanks, I had opened many tabs with same title and overlooked that one! :+1:

DerDu:
Just a quick referencing of issues on the same topic 
- #216 Annotations from Traits are ignored
- #268 Annotation imports are not parsed for nested traits

PR #102, opened 2016-10-27
Fix for issue #81 Traits Composed from Traits

hboomsma:
Fix for issue #81 

Use statements are now also imported for properties
and methods from traits in traits. Conflict resolution
and aliases are respected.

For backwards compatibility reasons the use statements
from the declaring class are still merged. PHP itself 
does not show this behaviour regarding to imported types
in traits.

For the methods in traits I use `\ReflectionMethod::getFileName()`
and then parse the file since this is the only way I can think of to
get the right declaration location in the face of imported traits using
conflict resolution. This same method is also possible for the
properties if you would prefer it.

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -344,22 +344,15 @@ class AnnotationReader implements Reader
     private function getMethodImports(ReflectionMethod $method)
     {
         $class = $method->getDeclaringClass();
-        $classImports = $this->getClassImports($class);
-        if (!method_exists($class, 'getTraits')) {
-            return $classImports;
-        }
+        $classImports  = $this->getClassImports($class);
 
-        $traitImports = array();
+        // Since there is no way to know where a method originates
+        // when conflict resolution for traits has been used we fall
+        // back to using the filename form the \ReflectionMethod
+        // directly.
+        $methodImports = $this->phpParser->parseMethod($method);
 
-        foreach ($class->getTraits() as $trait) {
-            if ($trait->hasMethod($method->getName())
-                && $trait->getFileName() === $method->getFileName()
-            ) {
-                $traitImports = array_merge($traitImports, $this->phpParser->parseClass($trait));
-            }
-        }
-
-        return array_merge($classImports, $traitImports);
+        return array_merge($classImports, $methodImports);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


tommygnr:
It would be great to get this merged if possible. It's a very annoying bug.

tommygnr:
Bump @Ocramius @guilhermeblanco

johnpancoast:
It's been just shy of 2 years since this was submitted. Is there anything we can do to help get this merged?

Ocramius:
I simply completely missed this (among the 186 open PRs I'm mentioned in).

Keeping a tab open, can't promise much.

wadjeroudi:
@Ocramius I think this PR should have a higher priority now, bcz the use of tools like php cs fixer has become democratized. Thx.

Majkl578:
This is what PHPStan is using, we should use something similar or based on that: https://github.com/phpstan/phpstan/blob/0.10.6/src/Reflection/Php/PhpClassReflectionExtension.php#L365-L437

Majkl578:
Ok, this gets REALLY tricky when only using PHP's Reflection:
- It's not possible to reliably detect whether a method actually comes from a trait or was shadowed by another method with same signature in the class. While this would work in most cases, it wouldn't in an edge case when all code is on a single line in the same file: https://3v4l.org/h9Uq1
- There is no way to know where property comes from.

I don't think support for traits is deterministically doable without introspecting the code itself using Nikic's PHP Parser (which is something I would like to avoid, but am open to do since it would also be useful for `use` parser). /cc @Ocramius?

Ocramius:
If we can keep `trait`s out of the game completely, I'm a happy coder anyway.

I realise many folks rely on them, but they are really really messy overall. If we can harden the tests well enough to support all edge cases, then fine, but otherwise we shouldn't ship something half-baked.

Ocramius:
The baseline is: loads and loads of test scenarios for the traits (for properties, methods and class-level annotations), for following scenarios (which are the ones I can think of immediately):

 * class using trait, annotation in class, no collision
 * class using trait, annotation in trait, no collision
 * class using trait, annotation in trait, collision
 * class using trait, annotation in class, aliased import
 * class using trait, annotation in trait, aliased import
 * class using trait using trait, annotation in second trait, no collision
 * class using trait using trait, annotation in second trait, collision
 * class using trait using trait, annotation in second trait, aliased import

Majkl578:
Problem is that we can't even reliably leave traits out of equation. Unless you bail out immediately when a class `use`s there _may_ or _may not_ be trait stuff incorportated (can be shadowed by class or not) and this is not properly detectable by Reflection. :S 
Also like it or not, people use traits for weird stuff and I know a lot of scenarios where people use stuff like `IdentifiedTrait` for ORM entities to define auto-integer `$id`, or `UuidIdentifiedTrait`.

Majkl578:
Current behaviour is IMO even worse: it ignores `use`s from trait files and always only uses `use`s from the class importing trait: https://github.com/doctrine/annotations/blob/master/tests/Doctrine/Tests/Annotations/Fixtures/ControllerWithTrait.php#L5-L6

PR #104, opened 2016-11-14
skip cache writes.

trsteel88:
Cache files aren't used when debug is true. To save IO, this PR skips the write function. It also skips the $this->cache->fetch() execution.

If the cache files aren't being read from, I can't think of a reason why the files should be written.

### DIFF
--- a/lib/Doctrine/Common/Annotations/CachedReader.php
+++ b/lib/Doctrine/Common/Annotations/CachedReader.php
@@ -185,8 +185,12 @@ final class CachedReader implements Reader
      */
     private function fetchFromCache($cacheKey, ReflectionClass $class)
     {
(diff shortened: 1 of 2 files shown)


trsteel88:
After looking into this further, I found that the problem I'm having with excessive IO is because getLastModification is merging the timestamps incorrectly.

The result of the following can't be merged:

```php
dump(array_merge(
    array_map([$this, 'getTraitLastModificationTimes'], $class->getTraits()),
));
```

screen shot 2016-11-14 at 11 29 03 am

The result of this is that it can't work out the max() of the traits.

To resolve this, I updated getTraitLastModificationTimes to always return the max timestamp.

I've moved this to a new PR.

PR #105, opened 2016-11-14
Timestamp returned for trait last modification times in the cached reader should be an integer, is an array

trsteel88:
This PR fixes an issue where the cache is constantly being written - wasting IO.

After looking into this further, I found that the problem I'm having with excessive IO is because getLastModification is merging the timestamps incorrectly.

The result of the following can't be merged:

```php
dump(array_merge(
    array_map([$this, 'getTraitLastModificationTimes'], $class->getTraits()),
));
```

screen shot 2016-11-14 at 11 29 03 am

The result of this is that it can't work out the max() of the traits.

To resolve this, I updated getTraitLastModificationTimes to always return the max timestamp.

Using this PR in favour of another change I made: https://github.com/doctrine/annotations/pull/104

### DIFF
--- a/lib/Doctrine/Common/Annotations/CachedReader.php
+++ b/lib/Doctrine/Common/Annotations/CachedReader.php
@@ -246,13 +246,17 @@ final class CachedReader implements Reader
         ));
     }
 
+    /**
+     * @param ReflectionClass $reflectionTrait
+     * @return int
+     */
     private function getTraitLastModificationTimes(ReflectionClass $reflectionTrait)
     {
         $fileName = $reflectionTrait->getFileName();
 
-        return array_merge(
+        return max(array_merge(
             [$fileName ? filemtime($fileName) : 0],
             array_map([$this, 'getTraitLastModificationTimes'], $reflectionTrait->getTraits())
-        );
+        ));
     }
 }
--- a/lib/Doctrine/Common/Annotations/CachedReader.php
+++ b/lib/Doctrine/Common/Annotations/CachedReader.php
@@ -240,7 +240,7 @@ final class CachedReader implements Reader
 
         return max(array_merge(
             [$filename ? filemtime($filename) : 0],
-            array_map([$this, 'getTraitLastModificationTimes'], $class->getTraits()),
+            array_map([$this, 'getTraitLastModificationTime'], $class->getTraits()),
             array_map([$this, 'getLastModification'], $class->getInterfaces()),
             $parent ? [$this->getLastModification($parent)] : []
         ));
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@trsteel88 Can you add a test case for this ?

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -246,13 +246,17 @@ private function getLastModification(ReflectionClass $class)
         ));
     }
 
+    /**
+     * @param ReflectionClass $reflectionTrait
+     * @return int
+     */
     private function getTraitLastModificationTimes(ReflectionClass $reflectionTrait)
```
stof: the name should be changed, as it does not return several times anymore

Ocramius: This method returns an `int[]`, which is then passed to `max()` in `private function getLastModification()`. If I get it correctly, the fix is different, and it is to change that function's body instead:

```php
         return max(array_merge(
            // notice the three dots:
             ...array_map([$this, 'getTraitLastModificationTimes'], $class->getTraits()),
         ));
```

Ocramius (changes requested):
A test case demonstrating the current issue is also needed.

lib/Doctrine/Common/Annotations/CachedReader.php
trsteel88: The three dots don't work.

Compile Error: Cannot use positional argument after argument unpacking

trsteel88:
@Ocramius @mikeSimonson where should this test go? I can't see any existing tests for the CachedReacher class.

I'm assuming we just need a single test to ensure that getLastModification returns an int (and never an array)?

mikeSimonson:
@trsteel88 The test should go [there](https://github.com/doctrine/annotations/blob/master/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php) with the other tests of cachedReader.

trsteel88:
Thanks @mikeSimonson - I'll add one in.

Also, I noticed that within ***fetchFromCache()*** it has the following:

```
     if (($data = $this->cache->fetch($cacheKey)) !== false) {
            if (!$this->debug || $this->isCacheFresh($cacheKey, $class)) {
                return $data;
            }
        }
```

In dev, debug is always going to be false. This is rendering the cache completely useless. Is this meant to be there?

Ocramius:
In dev mode, caches should really not be hit at all, so that's intentional.

Ocramius:
@trsteel88 merged, I added a test to it :-)

`master`: doctrine/annotations@8ac9494a6b91fca4b51a9e845179afc50d25d634
`1.3.x`: doctrine/annotations@bc1963dbceda634ee91c97d0bd19ccd1fc9a0f98

chosroes:
This caused the error in #117 =/

Ocramius:
@chosroes I still need a test case for that - can't do anything without it, and I'm not bootstrapping a symfony application for that.

PR #110, opened 2016-12-28
setIgnoreNotImportedAnnotations(true) didn’t work for existing classes.

bfanger:
Because FuelPHP has a Package class, which isn’t an @Annotation the following error occurs:

> [Semantical Error] The class "package" is not annotated with @Annotation. Are you sure this class can be used as annotation? If so, then you need to add @Annotation to the _class_ doc comment of "package". If it is indeed no annotation, then you need to add @IgnoreAnnotation("package") to the _class_ doc comment of ...

---

Duplicate of #25, requested by @Ocramius

### DIFF
--- a/tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
@@ -387,11 +387,8 @@ abstract class AbstractReaderTest extends \PHPUnit_Framework_TestCase
         $reader = $this->getReader();
         $ref = new \ReflectionClass('Doctrine\Tests\Common\Annotations\Fixtures\InvalidAnnotationUsageButIgnoredClass');
         $annots = $reader->getClassAnnotations($ref);
-        if ($annots[0] instanceof IgnoreAnnotation) {
-            $this->assertEquals(2, count($annots));
-        } else { // SimpleAnnotationReader doens't include the IgnoreAnnotation in the results.
-            $this->assertEquals(1, count($annots));
-        }
+
+        $this->assertEquals(2, count($annots));
     }
 
     /**
--- a/tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
@@ -387,8 +387,11 @@ abstract class AbstractReaderTest extends \PHPUnit_Framework_TestCase
         $reader = $this->getReader();
         $ref = new \ReflectionClass('Doctrine\Tests\Common\Annotations\Fixtures\InvalidAnnotationUsageButIgnoredClass');
         $annots = $reader->getClassAnnotations($ref);
-
-        $this->assertEquals(2, count($annots));
+        if ($annots[0] instanceof IgnoreAnnotation) {
+            $this->assertEquals(2, count($annots));
+        } else { // SimpleAnnotationReader doens't include the IgnoreAnnotation in the results.
+            $this->assertEquals(1, count($annots));
+        }
     }
 
     /**
--- a/tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
@@ -1049,6 +1049,15 @@ DOCBLOCK;
         $this->assertEquals(0, count($result));
     }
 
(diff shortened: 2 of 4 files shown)


tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
```diff
@@ -387,8 +387,11 @@ public function testInvalidAnnotationUsageButIgnoredClass()
         $reader = $this->getReader();
         $ref = new \ReflectionClass('Doctrine\Tests\Common\Annotations\Fixtures\InvalidAnnotationUsageButIgnoredClass');
         $annots = $reader->getClassAnnotations($ref);
-
-        $this->assertEquals(2, count($annots));
+        if ($annots[0] instanceof IgnoreAnnotation) {
+            $this->assertEquals(2, count($annots));
```
mikeSimonson: Please use assertCount everywhere you count in the assertion.

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
```diff
@@ -387,8 +387,11 @@ public function testInvalidAnnotationUsageButIgnoredClass()
...
+        } else { // SimpleAnnotationReader doens't include the IgnoreAnnotation in the results.
+            $this->assertEquals(1, count($annots));
```
mikeSimonson: assertCount

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -1049,6 +1049,15 @@ public function testNotAnAnnotationClassIsIgnoredWithoutWarning()
         $this->assertEquals(0, count($result));
     }
 
+    public function testNotAnAnnotationClassIsIgnoredWithoutWarningWithoutCheating()
+    {
+        $parser = new DocParser();
+        $parser->setIgnoreNotImportedAnnotations(true);
+        $result = $parser->parse('@PHPUnit_Framework_TestCase');
+
+        $this->assertEquals(0, count($result));
```
mikeSimonson: assertCount

mikeSimonson:
@bfanger Thanks for the PR.

Someone still needs to verify that the logic of this change is sound.

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
bfanger: `grep -r "assertCount" tests/|wc -l` = 13
 `grep -r "assertEquals" tests/|grep "count("|wc -l` = 51
This codebase clearly prefers assertEquals over assertCount.

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
```diff
@@ -387,8 +387,11 @@ public function testInvalidAnnotationUsageButIgnoredClass()
         $reader = $this->getReader();
         $ref = new \ReflectionClass('Doctrine\Tests\Common\Annotations\Fixtures\InvalidAnnotationUsageButIgnoredClass');
         $annots = $reader->getClassAnnotations($ref);
-
-        $this->assertEquals(2, count($annots));
+        if ($annots[0] instanceof IgnoreAnnotation) {
```
bfanger: I have removed if statement from the unittest and moved testing the behaviour of SimpleAnnotationReader into the SimpleAnnotationReaderTest

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
mikeSimonson: You realise that this justification is weak right ?
With that kind of logic why not using assertTrue everywhere ?
Why do you think that we need to conform to use only one assertion in the hundred that exists ?

The reason to use assertCount here is because it will make the error message when the test fails an order of magnitude easier to understand and that why we test.

bfanger: I'm in favour of `assertCount()`, but when creating pull requests it's considered good behaviour to conform to the style of the project. 

Please create a pull request changing all the assertEquals into the better assertCount and when that gets merged first i'll gladly change the assertions.

PS. In my personal project I use `assertSame` and don't use `assertEquals` at all, as it doesn't catch subtle bugs like `$this->assertEquals("1", true);`

Ocramius:
Merged and moved to count-oriented assertions. Will apply further fixes to assertions on `master`

PR #112, opened 2017-01-02
Removes support for PHP 5.x

railto:
Removes support for PHP 5.x from composer.json and .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,4 @@ matrix:
         
 before_script:
     - composer install -n
+    - if [[ $TRAVIS_PHP_VERSION = hhvm* ]]; then echo hhvm.php7.all=1 >> /etc/hhvm/php.ini; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,9 @@ php:
 matrix:
     allow_failures:
         - php: hhvm
+
+before_install:
+    - if [[ $TRAVIS_PHP_VERSION = hhvm* ]]; then echo hhvm.php7.all=1 >> /etc/hhvm/php.ini; fi
         
 before_script:
     - composer install -n
-    - if [[ $TRAVIS_PHP_VERSION = hhvm* ]]; then echo hhvm.php7.all=1 >> /etc/hhvm/php.ini; fi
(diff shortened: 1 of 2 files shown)


Ocramius:
Same as with doctrine/common, HHVM version in travis-ci needs tweaking.

.travis.yml
```diff
@@ -7,10 +7,13 @@ cache:
...
 
 php:
-  - 5.6
   - 7
   - 7.1
   - hhvm
 
+matrix:
+    allow_failures:
+        - php: hhvm
```
Ocramius: Not the correct solution -  HHVM should simply be installed in a PHP7-compatible version

mpalourdio:
@Ocramius @railto What about dropping this kind of checks ?
 
https://github.com/doctrine/annotations/blob/master/lib/Doctrine/Common/Annotations/AnnotationReader.php#L186

I remember introducing them in https://github.com/doctrine/annotations/pull/59

I suppose we don't need them anymore :)

mpalourdio:
Also, I'm curious to see how make hhvm PH7-compatible. Any reference somewhere of a travis config that works this way ?

railto:
This is the same as on https://github.com/doctrine/common/pull/756, will need to find a proper resolution for that as well as this repo.  Unfortunately that is something outside of my expertise as I do not work with travis myself.

mpalourdio:
I just remember a tweeter thread from @fabpot asking for this, but can't find it ATM

mpalourdio:
The thread is here for reference : https://twitter.com/fabpot/status/810179750606016513

Ocramius:
So in `travis.yml`: `- if [[ $TRAVIS_PHP_VERSION = hhvm* ]]; then echo hhvm.php7.all=1 >> /etc/hhvm/php.ini; fi`

lcobucci:
Looks like it's using a wrong version of [HHVM](https://travis-ci.org/doctrine/annotations/jobs/189765758#L147). PHP 7 support was just added on HHVM 3.11.

Can you try to update the matrix to use [hhvm-3.12 and ubuntu trusty](https://docs.travis-ci.com/user/languages/php#HHVM-versions-on-Trusty)?

railto:
Closing as I do not have the ability to get this to work with HHVM tests

PR #118, opened 2017-01-08
Integrating PHPStan

ondrejmirtes:
This is a PR for integrating static analysis tool [PHPStan](https://github.com/phpstan/phpstan) that focuses on finding bugs and running in your CI build. It found a few issues that I put in separate commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,5 +12,17 @@ php:
   - 7.1
   - hhvm
 
+matrix:
+  include:
+    - php: 7
+      env: PHPSTAN=1
+    - php: 7.1
+      env: PHPSTAN=1
+
 before_script:
     - composer install -n
+    - if [[ $PHPSTAN = 1 ]]; then composer require --dev phpstan/phpstan:^0.5.2; fi
+
+script:
+    - vendor/bin/phpunit
+    - if [[ $PHPSTAN = 1 ]]; then vendor/bin/phpstan analyse -c phpstan.neon -l 3 lib tests; fi
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -307,7 +307,8 @@ class AnnotationReader implements Reader
      */
     private function getIgnoredAnnotationNames(ReflectionClass $class)
     {
-        if (isset($this->ignoredAnnotationNames[$name = $class->getName()])) {
+        $name = $class->getName();
+        if (isset($this->ignoredAnnotationNames[$name])) {
             return $this->ignoredAnnotationNames[$name];
         }
 
@@ -325,7 +326,8 @@ class AnnotationReader implements Reader
      */
     private function getClassImports(ReflectionClass $class)
     {
-        if (isset($this->imports[$name = $class->getName()])) {
+        $name = $class->getName();
+        if (isset($this->imports[$name])) {
             return $this->imports[$name];
         }
 
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,14 @@
+parameters:
+    autoload_files:
+        - %currentWorkingDirectory%/tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
+    excludes_analyse:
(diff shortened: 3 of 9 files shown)


lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -307,7 +307,8 @@ public function getMethodAnnotation(ReflectionMethod $method, $annotationName)
      */
     private function getIgnoredAnnotationNames(ReflectionClass $class)
     {
-        if (isset($this->ignoredAnnotationNames[$name = $class->getName()])) {
+        $name = $class->getName();
```
Ocramius: The expression was valid before - consider raising an issue in phpstan, since `$name` would have been defined in any case

lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -325,7 +326,8 @@ private function getIgnoredAnnotationNames(ReflectionClass $class)
      */
     private function getClassImports(ReflectionClass $class)
     {
-        if (isset($this->imports[$name = $class->getName()])) {
+        $name = $class->getName();
```
Ocramius: The expression was valid before - consider raising an issue in phpstan, since `$name` would have been defined in any case

phpstan.neon
```diff
@@ -0,0 +1,14 @@
+parameters:
+	autoload_files:
+		- %currentWorkingDirectory%/tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```
Ocramius: No tabs, please

tests/Doctrine/Tests/Common/Annotations/Fixtures/ClassWithClosure.php
```diff
@@ -26,7 +26,7 @@
     public function methodName(\Closure $callback)
     {
         $self = $this;
-        return function() use ($self, $callback) {
+        return function() use ($callback) {
```
Ocramius: Don't understand this change

.travis.yml
```diff
@@ -12,5 +12,16 @@ php:
...
+      env: PHPSTAN=1
+    - php: 7.1
+      env: PHPSTAN=1
+
 before_script:
     - composer install -n
+
+script:
+  - vendor/bin/phpunit
+  - if [[ $PHPSTAN = 1 ]]; then composer require --dev phpstan/phpstan:^0.5.2 && vendor/bin/phpstan analyse -c phpstan.neon -l 3 lib tests; fi
```
Ocramius: `phpstan/phpstan` installation should be part of `before_script` (also there: to be done conditionally)

tests/Doctrine/Tests/Common/Annotations/Fixtures/ClassWithClosure.php
ondrejmirtes: It's in a separate commit with description - `$self` is unused inside the closure. I could probably also remove the `$self = $this` line. PHPStan also detects unused constructor parameters in a similar fashion - what can be removed without consequences, or was probably forgotten to be used (like injected parameter in the constructor that's not injected into a property).

Ocramius: Alright. Not going to be picky on this one, since it's a fixture :-)

Ocramius:
LGTM 👍

.travis.yml
```diff
@@ -12,5 +12,17 @@ php:
   - 7.1
   - hhvm
 
+matrix:
+  include:
+    - php: 7
+      env: PHPSTAN=1
+    - php: 7.1
+      env: PHPSTAN=1
```
stof: No need to add a second PHP 7 job running both PHPUnit and PHPSTAN IMO. It increases the matrix for nothing.

ondrejmirtes: I am used to putting it separately in case you'd need to add it allow_failures temporarily.

ondrejmirtes: Of course PHPUnit does not need to run in this job, there'd need to be another flag for turning PHPUnit off.

Issue #115, opened 2017-01-04
Missing annotations with the latest composer version


CracyCrazz:
Hey guys,

I have already posted this issue in the doctrine2 Repository as well as the composer repository itself see:

https://github.com/doctrine/doctrine2/issues/6214
https://github.com/composer/composer/issues/6038

Just a short briefing: Since the latest update to the composer version 1.3 some annotations can't be found in our Symfony applications for example the PrePersist Annotation in our entity.

This only happens if we prefix the use statement with a "\" like: 

`use \Doctrine\ORM\Mapping as ORM;`

Once we remove that prefix everything works out fine.

Before composer version 1.3 it worked out fine as composer itself replaced the first character if it was an backslash as an workaround for an old PHP 5.3 issue.

According to Seladek, one of the composer members, this issue is related to the way the parser handles the namespace as it is and not replaces the prefixed slash.

I was about to start a PR for this but then I found a comment here:

https://github.com/doctrine/annotations/blob/b544ea1f2f2e1af294dbd516d4ddecd3db3d7e33/lib/Doctrine/Common/Annotations/DocParser.php#L689

And was wondering if all of this is intended or not?

Would appreciate any answer.

Ocramius:
@CracyCrazz this needs a failing test case: can't really do anything with the description provided so far.

stof:
@CracyCrazz Can you should us how your usage of ``AnnotationRegistry`` looks like (there should be at least a place using it in your project, outside the doctrine/annotations code itself) ?

rommsen:
I cant provide a failing test case because I really dont understand what is happening here but I spent the last 8 hours trying to find a reason why suddenly our JmsJobBundle was not working anymore.

We got many `PHP Warning:  Invalid argument supplied for foreach() in vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/CachedReader.php on line 125` warnings because `$this->getPropertyAnnotations($property)` returns null in some occasions

Furthermore we had this error: ` [Doctrine\ORM\Query\QueryException]                                                                                                                          
  [Semantical Error] line 0, col 118 near 'queue NOT IN(:excludedQueues)': Error: Class JMS\JobQueueBundle\Entity\Job has no field or association named queue`
but the annotation is clearly there: https://github.com/schmittjoh/JMSJobQueueBundle/blob/master/Entity/Job.php#L98

[...]

PR #120, opened 2017-02-09
Missing annotations with the latest composer version

gnat42:
This PR demonstrates the issue with #115 as well as a potential fix.

There's one oddity here however, I can only cause the failure if I *only* run that one test. This is why the test I added has a '@group gnat' annotation.

When I run 
```
phpunit --group gnat
```
Without my fix in place in DocParser it fails, running all tests for some reason doesn't fail. I'm not sure why. I also realize that really I should have a test against DocParser instead, since that is where the issue lies. However this was a proof of concept of the issue. If you comment out my changes in DocParser, the error we're getting in symfony based projects happens.

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -728,6 +728,8 @@ final class DocParser
             }
         }
 
+        $name = ltrim($name,'\\');
+
         if ( ! $this->classExists($name)) {
             throw AnnotationException::semanticalError(sprintf('The annotation "@%s" in %s does not exist, or could not be auto-loaded.', $name, $this->context));
         }
--- a/tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
@@ -2,8 +2,11 @@
 
 namespace Doctrine\Tests\Common\Annotations;
 
+use Doctrine\Common\Annotations\AnnotationException;
 use Doctrine\Common\Annotations\AnnotationReader;
 use Doctrine\Common\Annotations\DocParser;
+use Doctrine\Tests\Common\Annotations\Fixtures\Annotation\SingleUseAnnotation;
+use Doctrine\Tests\Common\Annotations\Fixtures\ClassWithFullPathUseStatement;
 use Doctrine\Tests\Common\Annotations\Fixtures\IgnoredNamespaces\AnnotatedAtClassLevel;
 use Doctrine\Tests\Common\Annotations\Fixtures\IgnoredNamespaces\AnnotatedAtMethodLevel;
 use Doctrine\Tests\Common\Annotations\Fixtures\IgnoredNamespaces\AnnotatedAtPropertyLevel;
@@ -107,5 +110,19 @@ class AnnotationReaderTest extends AbstractReaderTest
         $reader::addGlobalIgnoredNamespace('SomePropertyAnnotationNamespace');
 
         self::assertEmpty($reader->getPropertyAnnotations($ref->getProperty('property')));
+    }
+
+    public function testClassWithFullPathUseStatement()
+    {
+        if (class_exists(SingleUseAnnotation::class, false)) {
+            throw new \LogicException('The SingleUseAnnotation must not be used in other tests for this test to be useful');
+        }
+
+        $reader = $this->getReader();
+        $ref = new \ReflectionClass(ClassWithFullPathUseStatement::class);
+
+        $annotations = $reader->getClassAnnotations($ref);
+
+        self::assertInstanceOf(SingleUseAnnotation::class,$annotations[0]);
     }
 }
--- a/tests/Doctrine/Tests/Common/Annotations/Fixtures/ClassWithFullPathUseStatement.php
+++ b/tests/Doctrine/Tests/Common/Annotations/Fixtures/ClassWithFullPathUseStatement.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Doctrine\Tests\Common\Annotations\Fixtures;
+
+/**
+ * The leading \ is intentional to ensure that code using
+ * this use statement format works see issue #115/PR #120
+ */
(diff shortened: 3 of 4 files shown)


lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -728,6 +728,11 @@ private function Annotation()
             }
         }
 
+        // strip the first \ off if used
+        if( '\\' == $name[0]) {
+            $name = substr($name,1);
```
Ocramius: `ltrim($name, '\\')`

tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
```diff
@@ -108,4 +110,23 @@ public function testPropertyAnnotationIsIgnored()
 
         self::assertEmpty($reader->getPropertyAnnotations($ref->getProperty('property')));
     }
+
+    /**
+     * @group gnat
```
Ocramius: to be removed

tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
```diff
@@ -108,4 +110,23 @@ public function testPropertyAnnotationIsIgnored()
...
+     */
+    public function testClassWithFullPathUseStatement()
+    {
+        try {
+            $reader = $this->getReader();
+            $ref = new \ReflectionClass(ClassWithFullPathUseStatement::class);
+
+            $annotations = $reader->getClassAnnotations($ref);
+        } catch(AnnotationException $exception) {
+            $this->fail('Class with full path use statement unable to load annotation');
```
Ocramius: If you don't expect an exception, let it bubble up

tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
```diff
@@ -108,4 +110,23 @@ public function testPropertyAnnotationIsIgnored()
...
+        }
+
+        $this->assertTrue(true);
```
Ocramius: Please assert for the annotations to be at least of the type you expected

tests/Doctrine/Tests/Common/Annotations/Fixtures/ClassWithFullPathUseStatement.php
```diff
@@ -0,0 +1,22 @@
+<?php
+/**
```
Ocramius: to be removed

tests/Doctrine/Tests/Common/Annotations/Fixtures/ClassWithFullPathUseStatement.php
```diff
@@ -0,0 +1,22 @@
...
+ * Created by PhpStorm.
+ * User: gnat
+ * Date: 09/02/17
+ * Time: 9:39 AM
+ */
+
+namespace Doctrine\Tests\Common\Annotations\Fixtures;
+
+use \Doctrine\Tests\Common\Annotations\Fixtures\Annotation as Annotations;
```
Ocramius: Please add a comment explaining that the leading `\` is intentional

tests/Doctrine/Tests/Common/Annotations/Fixtures/ClassWithFullPathUseStatement.php
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Class ClassWithFullPathUseStatement
```
Ocramius: remove

tests/Doctrine/Tests/Common/Annotations/Fixtures/ClassWithFullPathUseStatement.php
```diff
@@ -0,0 +1,22 @@
...
+ * @package Doctrine\Tests\Common\Annotations\Fixtures
```
Ocramius: remove

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -728,6 +728,11 @@ private function Annotation()
             }
         }
 
+        // strip the first \ off if used
```
Ocramius: Can actually just call `ltrim($name, '\\')` all the time, and make this comment and conditional redundant.

stof:
@gnat42 the bug does not appear when running all tests, because you reused an annotation used in other tests, and so the class is already loaded when reaching your test. But the bug only happens when the class is not yet loaded (classes already loaded are detected using ``class_exists($name, false)``, which goes through the PHP processing of class names and so is not buggy).
to make the tests usable in the full testsuite, use an annotation class created specially for this test. And add ``$this->assertFalse(class_exists(YourAnnotation::class, false), 'The YourAnnotation must not be used in other tests for this test to be useful');`` assertion at the beginning of your test to forbid breaking the test in the future by reusing this annotation by mistake.

Ocramius:
> `$this->assertFalse(class_exists(YourAnnotation::class, false), 'The YourAnnotation must not be used in other tests for this test to be useful');`

very clever! Indeed, please add that one :-)

@stof he's already using `@runInSeparateProcess`

stof:
@Ocramius I prefer using a separate annotation class used only for this test. Much less overhead than forcing a subprocess for the test, for the same result

Ocramius:
Agree, good point. The assertion to be added is obviously needed too.

stof:
thus, I not even sure that ``@runInSeparateProcess`` would solve the issue here. IIRC, PHPUnit forces reloading defined classes when launching the subprocess (at least, PHPUnit 4.x was doing it; I haven't checked whether it changed in 5.x)

stof:
even better than the assertion:

```php
if (class_exists(YourAnnotation::class, false)) {
    throw new LogicException('The YourAnnotation must not be used in other tests for this test to be useful');
}
```

We want to mark the test as an error when someone messes with the testsuite requirement, not as failed. The issue is not in the code but in the testsuite

gnat42:
I think that I've addressed all changes/suggestions. Let me know if I missed any.

mikeSimonson:
@gnat42 Thanks

mikeSimonson:
Fix #115

PR #122, opened 2017-02-17
Allow support for JSON arrays within annotations

rlweb:
As said, added square brackets to complete support for JSON.

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocLexer.php
+++ b/lib/Doctrine/Common/Annotations/DocLexer.php
@@ -51,6 +51,8 @@ final class DocLexer extends AbstractLexer
     const T_TRUE                = 110;
     const T_NULL                = 111;
     const T_COLON               = 112;
+    const T_OPEN_SQR_BRACKETS   = 113;
+    const T_CLOSE_SQR_BRACKETS  = 114;
 
     /**
      * @var array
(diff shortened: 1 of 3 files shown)


tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -422,6 +422,8 @@ public function getAnnotationVarTypeProviderValid()
               // array type
             array('array', '{@AnnotationExtendsAnnotationTargetAll}'),
             array('array', '{@AnnotationExtendsAnnotationTargetAll,@AnnotationExtendsAnnotationTargetAll}'),
+            array('array', '[@AnnotationExtendsAnnotationTargetAll]'),
+            array('array', '[@AnnotationExtendsAnnotationTargetAll,@AnnotationExtendsAnnotationTargetAll]'),
```
Ocramius: Would need a few tests around parsing array elements, their order, and different types of sub-elements

Ocramius: Also nested arrays

lib/Doctrine/Common/Annotations/DocLexer.php
```diff
@@ -51,6 +51,8 @@
     const T_TRUE                = 110;
     const T_NULL                = 111;
     const T_COLON               = 112;
+    const T_OPEN_SQR_BRACKETS   = 113;
```
Ocramius: `SQUARE`

lib/Doctrine/Common/Annotations/DocLexer.php
```diff
@@ -51,6 +51,8 @@
...
+    const T_CLOSE_SQR_BRACKETS  = 114;
```
Ocramius: `SQUARE`

Ocramius (changes requested):
Code looks good, but we need to improve the testing of nested values, optional additional values.

This should be done both in the parser tests, and in the annotation reader afterwards, in my opinion.

@guilhermeblanco any opinion on supporting this?

Majkl578:
This is something that should be relatively easy with the new parser in Annotations 2.0. Milestoning, since there is no relevant issue, but it will require different implementation.

PR #125, opened 2017-03-02
Add `@required` to global ignore list

nicolas-grekas:
Symfony 3.3 is going to make use of the `@required` annotation, but in lower case form, to make it visually visible that this is a global annotation in the same group as `@deprecated` or `@return`, etc. - *vs* local annotations that map to classes, thus usually start with an uppercase.

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -56,7 +56,7 @@ class AnnotationReader implements Reader
         // Annotation tags
         'Annotation' => true, 'Attribute' => true, 'Attributes' => true,
         /* Can we enable this? 'Enum' => true, */
-        'Required' => true,
+        'Required' => true, 'required' => true,
         'Target' => true,
         // Widely used tags (but not existent in phpdoc)
         'fix' => true , 'fixme' => true,


Ocramius:
@nicolas-grekas that's not something commonly used, while it may be commonly used as annotation (forms, validators, etc): simply add the ignored annotation to the list within the annotation registry (at bootstrap) in symfony-standard.

Closing as `wontfix`

nicolas-grekas:
wow that's a pretty quick decision, where is the discussion?

Ocramius:
There is none: `@required` is a relatively generic name that might be already in use by a quite large number of packages (not necessarily open source), while Symfony is taking a decision to use `@Required` and `@required` (URGH, really, that's already a terrible idea) introducing a new convention that isn't already in phpunit, phpDocumentor, etc.

Either define a custom annotation class in symfony, or ignore it in the symfony kernel bootstrapper: it's as simple as that.

By ignoring the annotation by default in the annotations package, you might harm users that were already relying on such annotations to exist.

Ocramius:
Also: you are still in time for NOT creating YET ANOTHER convention, and forcing it onto an unrelated package :-P

nicolas-grekas:
You mean, stop the time and live in already written code? not my job ;)

Ocramius:
Your job as symfony maintainer is to prevent this stuff from happening in *not yet released* code (3.3 is not out there: pull the handbrake and revert course).

nicolas-grekas:
All these statements are your pov, not mines. Anyway, I'll do what you say and register manually on my side.

Issue #126, opened 2017-03-05
Wrong parsing of DocBlocks with quotes in irrelevant text


julienfalque:
In the following docblock:
```
/**
 * @param mixed description with a single " character.
 *
 * @Marker
 *
 * @This is an irrelevant text with a single " character.
 */
```

`@Marker` is not correctly parsed because the lexer sees it as part of a string.

Is this expected? I wrote tests that cover this [here](https://github.com/julienfalque/annotations/tree/quotes) but I'm not sure how to fix this.

Ocramius:
Could you please send the test case as a pull request?

Looks like a legitimate bug, and we should just stop looking for quotes
when we aren't in the context of an annotation.

Ocramius:
Closing, as this was moved to #127

PR #127, opened 2017-03-06
#126 Quotes failing tests - quotes outside an annotation make parsing fail

julienfalque:
See #126.


Ocramius:
@julienfalque I changed the base branch to `master`. Could you please rebase?

julienfalque:
Done.

Ocramius:
@julienfalque thanks! I won't get to it now, but at least this provides a decent base ground for fixing the issue :-)

julienfalque:
I can try to work on it if that would help.

It looks like all lexer tests assume that it only supports reading from the relevant Annotation part of the DocBlock, not the whole DocBlock itself. Is the lexer failing test case valid? Or should this be fixed in the parser only?

Ocramius:
@julienfalque the lexer is probably not where the bug is: just the parser may be affected. Same tokens, different result expected.

julienfalque:
With `@Foo " @Bar " @Baz`, the lexer does return a string token with `" @Bar "` as value, which is a bug to me.

As a workaround, the parser could simply pass `@Foo`, `@Bar` and `@Baz` successively. The parser already stips the beginning of the DocBlock until it finds the first `@` character (which prevents the bug if the quote is before it). This is why I'm not sure this should be fixed in the lexer (and maybe removing the need to strip the DocBlock) or in the parser (stripping all irrelevant parts instead of just the beginning).

greg0ire:
`master` is probably no longer the right branch for this, I believe 1.8 is now.

derrabus:
Closing as there's no progress to see here.

PR #131, opened 2017-04-14
Support for use statement resolution in annotation property types

Majkl578:
This is a rather simple approach to implement #86 / doctrine/common#562, originally reported by @enumag ages ago.
I don't know the internals of doctrine/annotations well so any suggestions are welcome. ☺️

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -59,6 +59,13 @@ final class DocParser
     private $lexer;
 
     /**
+     * The PHP parser.
+     *
+     * @var \Doctrine\Common\Annotations\PhpParser
+     */
+    private $phpParser;
+
+    /**
      * Current target context.
      *
      * @var integer
@@ -94,6 +101,13 @@ final class DocParser
      * @var array
      */
     private $classExists = array();
+
+    /**
+     * This hashmap is used internally to cache results of class uses look-ups.
+     *
+     * @var array
+     */
+    private $classUses = array();
 
     /**
      * Whether annotations that have not been imported should be ignored.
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -486,6 +496,28 @@ public function getAnnotationVarTypeProviderInvalid()
...
+            // unqualified annotation instance
+            array('unqualifiedAnnotation', 'AnnotationTargetAll', '@Name','an instance of Doctrine\Tests\Common\Annotations\Name'),
+
+            // renamed annotation instance
+            array('renamedAnnotation', 'RenamedAnnotationTargetAll', '@Name','an instance of Doctrine\Tests\Common\Annotations\Name'),
+
+            // partially namespaced annotation instance
+            array(
+                'partiallyNamedAnnotation',
+                'Common\Annotations\Fixtures\AnnotationTargetAll',
```
Majkl578: This assumption is strange, it should be the FQCN, not the one used in declaration. It works fine with arrays - see arrayOf* tests below, but for non-array values, it doesn't seem to be correct...

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -627,6 +664,85 @@ private function collectAttributeTypeMetadata(&$metadata, Attribute $attribute)
...
+    /**
+     * Resolves a type of the attribute to its fully qualified variant.
+     *
+     * @param array           $metadata
+     * @param Attribute       $attribute
+     * @param ReflectionClass $class
+     *
+     * @return void
+     */
+    private function resolveUnqualifiedAttributeType(&$metadata, Attribute $attribute, ReflectionClass $class)
```
Ocramius: We should *really* avoid by-ref modification wherever possible

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -627,6 +664,85 @@ private function collectAttributeTypeMetadata(&$metadata, Attribute $attribute)
...
+
+    /**
+     * Resolves a single type to its fully qualified variant.
+     *
+     * @param string|null     $type
+     * @param ReflectionClass $class
+     *
+     * @return void
+     */
+    private function resolveUnqualifiedType(&$type, ReflectionClass $class)
```
Ocramius: Same: by-ref is a no-go here

PR #133, opened 2017-05-25
Missing throws annotation in AnnotationReader

SenseException:
I've just added a missing `@throws` annotation to the constructor of AnnotationReader.

I'm not sure if this wasn't added before, because it only depends on the PHP installation and configuration and makes this information less interesting for developers during their development and maybe spoils their IDE's static code checks.

If that was the reason, you can close this PR.


Ocramius:
@SenseException thanks!

PR #136, opened 2017-07-17
Add trait annotation to class

allan-simon:
all credits to Matthieu Dumoulin https://github.com/doctrine/annotations/pull/58/commits
I simply added tests and some comments


allan-simon:
actually it seems to be a bug in my application code, not in the library which seems to work perfeclty ...., i've just seen there's already code to handle trait in the current code...

PR #142, opened 2017-08-25
#141 `::class` notations with FQCN notation and leading `\` causes `\` to be part of the produced string value

### LINKED ISSUE #141: `::class` notations with FQCN notation and leading `\` causes `\` to be part of the produced string value
See https://github.com/doctrine/doctrine2/pull/6593#discussion_r131360878

In PHP:

```
<?php
echo \SimpleXMLElement::class;
// SimpleXMLElement
```

In annotations:

```php
/** @Something(\SimpleXMLElement::class) */

// ... dump that value somehow, you'll get:

// \SimpleXMLElement
```

mikeSimonson:
When there is a leading backslash in front of a class name
as annotation value, the leading backslash gets removed.

Fixes #141

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -968,10 +968,14 @@ final class DocParser
             }
         }
 
-        // checks if identifier ends with ::class, \strlen('::class') === 7
-        $classPos = stripos($identifier, '::class');
-        if ($classPos === strlen($identifier) - 7) {
-            return substr($identifier, 0, $classPos);
+        /**
+         * Checks if identifier ends with ::class and remove the leading backslash if it exists.
+         */
+        if ($this->identifierEndsWithClassConstant($identifier) && ! $this->identifierStartsWithBackslash($identifier)) {
+            return substr($identifier, 0, $this->getClassConstantPositionInIdentifier($identifier));
+        }
+        if ($this->identifierEndsWithClassConstant($identifier) && $this->identifierStartsWithBackslash($identifier)) {
+            return substr($identifier, 1, $this->getClassConstantPositionInIdentifier($identifier) - 1);
         }
 
         if (!defined($identifier)) {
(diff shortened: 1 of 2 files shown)


Ocramius:
@tPl0ch can you check this and see if it matches the issue you experienced?

tPl0ch:
@Ocramius Looks like the culprit on first glance. Could add the failing behaviour to the doctrine test suite again to verify. Trying this evening.

mikeSimonson:
@tPl0ch (The/A) failing test is in the patch already. You just need to check that this fix your issue.
If it does not then my failing test is not the correct one.

Ocramius:
@mikeSimonson @tPl0ch was experiencing this first in an ORM patch, which is why I asked him to verify it :-)

mikeSimonson:
That's what I meant too, must have phrased it in a weird way.

malarzm:
@mikeSimonson could you rebase and solve the conflict please?

mikeSimonson:
@malarzm Done

Ocramius:
Will need to check this another time - currently locked onto 1.6.1 to fix a BC issue.

lib/Doctrine/Annotations/DocParser.php
```diff
@@ -933,11 +933,16 @@ private function Constant()
...
-        // checks if identifier ends with ::class, \strlen('::class') === 7
+        /* checks if identifier ends with ::class, \strlen('::class') === 7
+         * and remove the leading backslash if it exists.
+         */
         $classPos = stripos($identifier, '::class');
-        if ($classPos === strlen($identifier) - 7) {
+        if ($classPos === strlen($identifier) - 7 && '\\' !== $identifier[0]) {
             return substr($identifier, 0, $classPos);
         }
+        if ($classPos === strlen($identifier) - 7 && '\\' === $identifier[0]) {
```
Ocramius: The duplication here is a bit clunky

mikeSimonson:
@Ocramius What about now ? The build fails for unrelated reasons.

SenseException:
Failing tests are about the coming PHP 7.4 feature about array access on values like `null` and it looks like it is not related to your PR. The phpstan check found some issues in files that aren't part of your PR either.

greg0ire:
I don't think `master` should be target. bugfixes should go to 1.8

PR #150, opened 2017-10-25
Tests now use namespaced TestCase

MarioBlazek:


alcaeus:
You can also bump phpunit to a `^6.4` version in composer.json since we require PHP 7.1 anyways. That may entail some more changed class names, but it's worth doing IMO.

MarioBlazek:
@alcaeus I've update phpunit's constraint. wdyt ?

alcaeus:
@MarioBlazek changes look good. Would you mind taking a look at the risky tests (https://travis-ci.org/doctrine/annotations/jobs/293061728#L510)? That'd be great

MarioBlazek:
Sure.

MarioBlazek:
@alcaeus regarding risky tests. In PHPUnit version 4 and 5 `beStrictAboutTestsThatDoNotTestAnything` configuration option was by default set to `false`. In version 6 this option is by default set to `true` (https://github.com/sebastianbergmann/phpunit/pull/2487/files). This can be very easily addressed in `phpunit.xml.dist`. Should I leave this as is, or should I set configuration value for `beStrictAboutTestsThatDoNotTestAnything` to false ? wdyt ?

alcaeus:
The setting should remain on, but risky tests fixed:
- tests should make an assertion
- if they don't need an assertion, there's a `@doesNotPerformAssertions` annotation

If you're not sure about the tests, I can take a look tomorrow and finish the tests.

MarioBlazek:
@alcaeus please check tests.

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
```diff
@@ -113,7 +114,7 @@ public function testClassWithWithDanglingComma()
     public function testClassWithInvalidAnnotationTargetAtClassDocBlock()
     {
         $reader  = $this->getReader();
-        $reader->getClassAnnotations(new \ReflectionClass(Fixtures\ClassWithInvalidAnnotationTargetAtClass::class));
+        return $reader->getClassAnnotations(new \ReflectionClass(Fixtures\ClassWithInvalidAnnotationTargetAtClass::class));
```
lcobucci: Why are you adding a `return` statement in this file?

MarioBlazek:
With return statement I can use assertions on tests, that extend this one. Otherwise we would have risky tests. @lcobucci I'm open for suggestions.

MarioBlazek:
ping @lcobucci @alcaeus

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
```diff
@@ -305,7 +306,7 @@ public function testImportDetectsNotImportedAnnotation()
     public function testImportDetectsNonExistentAnnotation()
     {
         $reader = $this->getReader();
-        $reader->getPropertyAnnotations(new \ReflectionProperty(TestNonExistentAnnotationClass::class, 'field'));
+        return $reader->getPropertyAnnotations(new \ReflectionProperty(TestNonExistentAnnotationClass::class, 'field'));
```
stof: why adding such return statements in tests which are asserting the exception ?

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -1184,7 +1185,7 @@ class A {
 
         try {
             $parser = $this->createTestParser();
-            $parser->parse($docblock);
+            self::assertEmpty($parser->parse($docblock));
         } catch (\Exception $e) {
```
stof: catching any exception and then failing is not good: it swallows the clean failure exception.

would be much better to remove the try/catch entirely (same in previous test).

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
MarioBlazek: The idea was to use assert in child class then. Currently child class does not run any assertions. What do you suggest @stof ?

alcaeus: Took me a while to figure out how this worked. Instead of adding a `return` statement here, I'd add a `@doesNotPerformAssertions` annotation on the risky tests in `SimpleAnnotationReaderTest`: these tests check for the absence of an exception without asserting anything. Simply asserting for a return type for the sake of making an assertion is useless IMO, we might as well tell PHPUnit that we're not performing an assertion.

As an alternative, you can add appropriate assertions for the tests, but I'd skip that.

MarioBlazek:
@alcaeus tests updated as you suggested.

alcaeus:
Thanks @MarioBlazek 🍻

MarioBlazek:
@alcaeus thank you.

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -1065,7 +1066,7 @@ public function testNotAnAnnotationClassIsIgnoredWithoutWarningWithoutCheating()
     {
         $parser = new DocParser();
         $parser->setIgnoreNotImportedAnnotations(true);
-        $result = $parser->parse('@PHPUnit_Framework_TestCase');
+        $result = $parser->parse('@\PHPUnit\Framework\TestCase');
```
Majkl578: Use `::class`.

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -1055,8 +1056,8 @@ public function testNotAnAnnotationClassIsIgnoredWithoutWarning()
     {
         $parser = new DocParser();
         $parser->setIgnoreNotImportedAnnotations(true);
-        $parser->setIgnoredAnnotationNames(array(\PHPUnit_Framework_TestCase::class => true));
-        $result = $parser->parse('@PHPUnit_Framework_TestCase');
+        $parser->setIgnoredAnnotationNames(array(\PHPUnit\Framework\TestCase::class => true));
+        $result = $parser->parse('@\PHPUnit\Framework\TestCase');
```
Majkl578: Use `::class`.

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -1055,8 +1056,8 @@ public function testNotAnAnnotationClassIsIgnoredWithoutWarning()
     {
         $parser = new DocParser();
         $parser->setIgnoreNotImportedAnnotations(true);
-        $parser->setIgnoredAnnotationNames(array(\PHPUnit_Framework_TestCase::class => true));
-        $result = $parser->parse('@PHPUnit_Framework_TestCase');
+        $parser->setIgnoredAnnotationNames(array(\PHPUnit\Framework\TestCase::class => true));
```
Majkl578: Don't use FQCN here.

tests/Doctrine/Tests/Common/Annotations/SimpleAnnotationReaderTest.php
```diff
@@ -9,6 +9,8 @@ class SimpleAnnotationReaderTest extends AbstractReaderTest
     /**
      * Contrary to the behavior of the default annotation reader, we do just ignore
      * these in the simple annotation reader (so, no expected exception here).
+     *
+     * @doesNotPerformAssertions
```
Majkl578: Tests that do not assert anything sound pretty useless to me.

alcaeus: We've talked at length about these. The original test checks to ensure the annotation read throws an exception. The simple reader doesn't throw an exception, which is why this test overrides the original but doesn't make an assertion (the absence of an exception is all we need to assert). Hence the annotation.

PR #153, opened 2017-10-29
Ignoring @suppress annotation used by Phan

bobvandevijver:
Phan used `@suppress`, which currently throws an exception when used. See https://github.com/phan/phan/wiki/Annotating-Your-Source-Code#suppress.

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -101,7 +101,9 @@ class AnnotationReader implements Reader
         // PlantUML
         'startuml' => true, 'enduml' => true,
         // Symfony 3.3 Cache Adapter
-        'experimental' => true
+        'experimental' => true,
+        // Phan suppress error
+        'suppress' => true
     );
 
     /**


lcobucci:
@bobvandevijver could you please add tests to cover this change? Although is quite simple we do need tests =)

bobvandevijver:
Ah, this was based on #13, which also did not include tests. And actually, a new PR (#147, `@experimental`), also did not include tests.

I can actually not find any test that specifically tests the excluded tags... So I do not think it is in the scope of this PR to add tests for all :)

Ocramius:
Doesn't really matter: new patches come with new tests. This is probably just a missed review comment by @guilhermeblanco or @alcaeus.

In addition to that, aren't we just piling up too many generic exceptions due to every second package defining their own annotations, rather than just using annotations as it should be used?

In this specific case, a bootstrapper for Phan is more appropriate, since:

 * Phan requires a custom extension to run, which means that almost nobody uses it
 * suppression should be performed on non-docblocks, like inline comments, exactly like it is done in IDEs such as Netbeans or PHPStorm
 * `@suppress` is generic naming that might already be used by userland serializers

I know I'm being a pain in the butt, but we need some sort of policy about what makes it in the exclusion map. This is obviously a question for @doctrine/doctrinecore, not your fault.

bobvandevijver:
Sure, but I currently have no clue on where to start with these tests, as there are currently none for any of suppressed tags... 

I can agree with your points about the specific bootstrapper, however, I can not really find any documentation about creating such a bootstrapper (using Symfony full-stack). 

Maybe we should create an issue regarding the policy?

(By the way, I've found another solution which no longer requires me to use the `@suppress` annotation, so if you want, you can close this PR).

mikeSimonson:
@bobvandevijver 

Ocramius is not talking about testing this patch or not.
He is talking about the point of adding / "polluting" the code of a generic annotation library with every other custom project annotation.

In your case your suppress "annotation" should have the form `/* @suppress */`, which is the form used by ide for instance. But it's not supported by phan as long as I know you might want to test that.

The other solution is for you to configure the annotation reader in your project by using `AnnotationReader::addGlobalIgnoredName('suppress');`

Good luck,
By the way, I am interested in knowing if `/* @suppress */` works for phan ?

bobvandevijver:
@mikeSimonson Ocramius was talking about tests for this specific patch, but also added some points on the ignored names. I already stated I agree with the points on the ignored names, but I also stated that there are currently no tests that even touch the current ignored names.

`/* @suppress */` surely doesn't work, as it not a [correct doc block](https://github.com/phan/phan/wiki/Annotating-Your-Source-Code#doc-blocks). 

`/** @suppress */` is a valid doc block but also wont work as an issue type is required [Phan doc](https://github.com/phan/phan/wiki/Annotating-Your-Source-Code#suppress). Secondly, the suppress statement should be placed on "the doc block of the closest function, method or class scope to where the issue is emitted", which means it unfortunately doesn't work when placed directly above the errored line.

I knew about the `addGlobalIgnoredName()` method, but I'm not sure on where to add it. It works in the Bundle constructor, but I find that wonky. However, I no longer require the `@suppress` annotations, as I've fixed the exception with the [`assert`-function](https://github.com/phan/phan/issues/1248).

PR #154, opened 2017-11-02
Ingore PHPstorm's @mixin annotation

wouterj:
PHPstorm supports a `@mixin` annotation, which is currently not ignored by the AnnotationReader.

Jetbrains issue ref: https://youtrack.jetbrains.com/issue/WI-1730

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -96,6 +96,7 @@ class AnnotationReader implements Reader
         'SuppressWarnings' => true,
         // PHPStorm
         'noinspection' => true,
+        'mixin' => true,
         // PEAR
         'package_version' => true,
         // PlantUML


Ocramius:
Same discussion as https://github.com/doctrine/annotations/pull/153 applies here

mikeSimonson:
Hi,

Thanks for your PR.
In this instance we have to evaluate if adding that exception in a generic annotation library make sense.
Every other project has custom annotation and we cannot decide which one we choose to put in the global ignore list so easily. If we include a lot of stuff we will brake the code of some project that use a custom annotation with the same name in their code.

In your case your "annotation" should have the form /* @mixin */.
This way it will be recognized by an ide but not as a valid PHP annotation. (as it's not a valid docblock)

The other solution is for you to configure the annotation reader in your project by using AnnotationReader::addGlobalIgnoredName('mixin');

Good luck

wouterj:
Alright. This error was produced by eloquent using this annotation, but
I'll make sure to mark it as excluded in my bundle then.

Thanks for the tip about IDEs parsing /* */! Didn't know that and might be
usefull some time.

Op do 2 nov. 2017 13:31 schreef mikeSimonson <notifications@github.com>:

PR #155, opened 2017-11-08
fix AnnotationReader preParser set ignoreAnnotationNames bug

pinguo-lidanyang:


pinguo-lidanyang:
with this clode 

`
/**
 * @package Test
 */
class Test
{
      public function get(){}
}

`

when I try to get the annotations of the method 'get', I will get an exception "class package is not exists" in DocParser.php:499 . But the package is the default ignore annotation.

so I find that the preParser does not set the ignore annotation names or namespaces whick causes the bug.

alcaeus:
Please add a test case that fails without your change and passes with it. Thanks!

pinguo-lidanyang:
```
/**
 * Class Test
 * @package TestClass
 */
class TestClass
{
    public function test()
    {

    }
}

$reader = new \Doctrine\Common\Annotations\AnnotationReader();

$ref = new ReflectionMethod(TestClass::class, 'test');

var_dump($reader->getMethodAnnotations($ref));
```


just try this code in php.

Ocramius:
What Andreas asked for is an automated test to be added to the existing
test suite.

pinguo-lidanyang:
Sorry, I made a mistake. I should use this:

```
AnnotationRegistry::registerLoader(function($class){
    if (class_exists($class)) {
        return true;
    }
    return false;
});
```

Thanks for replying. Please close this pr

Ocramius:
You can simplify that to `AnnotationRegistry::
registerLoader('class_exists');`

malarzm:
Closing as per request

Issue #145, opened 2017-09-12
Memory leak in AnnotationRegistry::registerLoader() when called multiple times


TriAnMan:
If we call AnnotationRegistry::registerLoader() multiple times we've got an self::$loaders array growing indefinitely.

Consider to append self::$loaders with a keyed value.

A key for callable can be determined by spl_object_hash() for anonymous functions and by a combination of md5() and spl_object_hash() for other types of callable.

I can provide a non BC breaking PR for this issue if interested.

alcaeus:
@TriAnMan if you are referring to registering the same loader multiple times, this was fixed in #137.

stof:
@Ocramius idea: optimize the case where `AnnotationRegistry::registerLoader('class_exists')` is called repeatedly: if the loader being registered is `class_exists` and the last loader in the array is also `class_exists`, skip the registration (checking only the last loader rather than any loader is here to avoid breaking BC regarding re-registering the loaders in a different order. Registering `class_exists` twice in a row won't have any difference to a single registration, unless your PHP autoloader is accepting to load the class only on the second call, but then your PHP project is already pure madness).

Registering anything else than `class_exists` should be considered deprecated to allow getting rid of this AnnotationRegistry non-sense in 2.0. The registry is here just because popular pre-Composer autoloading libraries were regularly breaking `class_exists` in case of miss, including Doctrine Common (yeah, Doctrine Common was hurting itself during years) and ZF 1. Supporting the case of autoloaders breaking in case of miss is not worth it IMO (having a non-functional `class_exists` function is likely to break other parts of the project anyway).

@alcaeus no. #137 does not prevent the list of loaders to grow. It optimizes the performance of the usage of these loaders.

stof:
Note that registering `class_exists` repeatedly actually happens when running tests in Symfony with a recent PHP version: to hide this AnnotationRegistry non-sense from users, Symfony registers the loader automatically at the time the annotation parser is instantiated by the DI container.
But in the testsuite, each test would run with its own container, and so the registration could happen again when the service gets instantiated by the second test (and the third one, etc...). If you have lots of tests, the `AnnotationRegistry::$loader` variable becomes huge (with the same `class_exists` value many time in it). And this leaks memory.

stof:
Btw, we could make the implementation a bit more generic (avoiding to tie it to the `class_exists` recommendation): make a strict comparison between the last registered loader and the new loader, and skip the registration (the duplicate `class_exists` is just a special case of it then).

[...]

PR #137, opened 2017-07-22
#135 optimise multiple class load attempts in `AnnotationRegistry`

Ocramius:
This PR includes #135 as well as a number of fixes around the `AnnotationRegistry` class (which is mostly a horror show from the pre-composer era).

PR #161, opened 2017-12-04
Prevent loading class_exists multiple times

jrjohnson:
Implementing @stof's suggestion in https://github.com/doctrine/annotations/issues/145#issuecomment-337289305

This is just a pre-2.0 bandaid to help solve the biggest issue where `class_exists` is added as a loader automatically many many times which is a huge performance killer in testing symfony apps (https://github.com/symfony/symfony/issues/24596). While #137 solves this in applications it doesn't help with tests since they maintain the global static state and start up multiple kernels (and registered loaders) at the same time.

In order to avoid a BC break this only checks for duplicate class exists
loaders at the bottom of the stack.


Ocramius:
As already stated on that thread, the correct solution would be to add a new specialised endpoint.

lib/Doctrine/Common/Annotations/AnnotationRegistry.php
```diff
@@ -111,9 +111,11 @@ public static function registerAutoloadNamespaces(array $namespaces) : void
      */
     public static function registerLoader(callable $callable) : void
     {
-        // Reset our static cache now that we have a new loader to work with
-        self::$failedToAutoload   = [];
-        self::$loaders[]          = $callable;
+        if ($callable !== 'class_exists' || !self::isLastLoaderClassExists()) {
```
Majkl578: Please invert this condition and make it early-return. Also space around `!` is required.

lib/Doctrine/Common/Annotations/AnnotationRegistry.php
```diff
@@ -159,4 +161,17 @@ public static function loadAnnotationClass(string $class) : bool
...
     }
+
+    protected static function isLastLoaderClassExists() : bool
+    {
+        $count = count(self::$loaders);
+        if (0 === $count) {
+            return false;
+        }
+        $lastLoader = self::$loaders[$count - 1];
+        if (!is_string($lastLoader)) {
```
Majkl578: Please remove this condition, it's not an optimization and is redundant.

lib/Doctrine/Common/Annotations/AnnotationRegistry.php
```diff
@@ -159,4 +161,17 @@ public static function loadAnnotationClass(string $class) : bool
 
         return false;
     }
+
+    protected static function isLastLoaderClassExists() : bool
+    {
+        $count = count(self::$loaders);
+        if (0 === $count) {
```
Majkl578: Here no yoda used will be.

jrjohnson:
Thanks @Majkl578 I made those changes.

@Ocramius thanks for the feedback. My thinking is that adding a new `registerLoaderIfNotExists` works, but doesn't solve the problem immediately. It requires work across any package that depends on this one. This solution solves the problem immediately with no risk of BC break. Given that the entire registry is not long for this world it doesn't seem like adding a new method and then updating everything to use it is the right choice. It is your project though and going that route isn't more work for me so if you don't want to merge this I will be happy to go that route.

Ocramius:
> It requires work across any package that depends on this one. 

Yes, that is something that was already clear to me - I realize that I'm taking the non-DX solution, but the repeated registration of loaders is something to be fixed elsewhere anyway (like keeping a `static` singleton reference to the loader in the symfony kernel).

Overall, less magic workarounds, especially in a general-purpose library that is basically used everywhere.

jrjohnson:
This isn't about developer experience it is about where the bug is. The performance killing bug is in this library. It should be fixed here. More importantly it *can* be fixed here for a huge number of situations with no risk.

Ocramius:
From my PoV, the bug is calling the register loader multiple times. The
thing is doing exactly what it should do.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

stof:
@Ocramius the bug to me is the fact that this library requires using a global state to register loaders. Having to add a global state in Symfony itself to workaround a performance issue related to the global state of this library looks weird to me.

Ocramius:
That is indeed something that can be fixed in 3.x by simply not requiring a
loader at all, but for now a loader API that specifically de-duplicates its
internal spl autoloader stack is indeed the fix.

Whether symfony leverages it or not later on is a completely different
question.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

On Mon, Dec 4, 2017 at 6:41 PM, Christophe Coevoet <notifications@github.com

stof:
Well, why forcing to migrate to a new API (that will also disappear in 3.x) when you can make all projects out there benefiting from a better performance by just upgrading, in case they follow the recommendation of registering `class_exists` itself as loader (i.e. doing what Doctrine 3 will always do)

Ocramius:
Because the current API is behaving as expected?

PR #162, opened 2017-12-04
Add registerLoaderIfNotExists to AnnotationRegistry

jrjohnson:
This method allows consumers of this library to add loaders without
worrying about the performance implications of adding the same loader
multiple times.

Refs #145 
Refs #161

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationRegistry.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationRegistry.php
@@ -117,6 +117,18 @@ final class AnnotationRegistry
     }
 
     /**
+     * Registers an autoloading callable for annotations, if it is not already registered
+     *
+     * @deprecated this method is deprecated and will be removed in doctrine/annotations 2.0
+     */
+    public static function registerUniqueLoader(callable $callable) : void
+    {
+        if ( ! in_array($callable, self::$loaders, true) ) {
+            self::registerLoader($callable);
+        }
+    }
+
+    /**
      * Autoloads an annotation class silently.
      */
     public static function loadAnnotationClass(string $class) : bool
--- a/tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
@@ -168,4 +168,19 @@ class AnnotationRegistryTest extends TestCase
 
         self::assertSame(2, $failures);
     }
+
+    /**
+     * @runInSeparateProcess
+     */
+    public function testRegisterLoaderIfNotExistsOnlyRegisteresSameLoaderOnce() : void
+    {
+        $className = 'autoloadedClassThatDoesNotExist';
+        AnnotationRegistry::reset();
+        $autoLoader = self::createPartialMock(\stdClass::class, ['__invoke']);
+        $autoLoader->expects($this->once())->method('__invoke');
+        AnnotationRegistry::registerUniqueLoader($autoLoader);
+        AnnotationRegistry::registerUniqueLoader($autoLoader);
+        AnnotationRegistry::loadAnnotationClass($className);
+        AnnotationRegistry::loadAnnotationClass($className);
+    }
 }


lib/Doctrine/Common/Annotations/AnnotationRegistry.php
```diff
@@ -116,6 +116,18 @@ public static function registerLoader(callable $callable) : void
...
     }
 
+    /**
+     * Registers an autoloading callable for annotations, if it is not already registered
+     *
+     * @deprecated this method is deprecated and will be removed in doctrine/annotations 2.0
+     */
+    public static function registerLoaderIfNotExists(callable $callable) : void
+    {
+        if ( !in_array($callable, self::$loaders) ) {
```
lcobucci: Using a strict comparison would be better, right?

lcobucci: Also, is not following our coding standard. Should be:

```php
if ( ! in_array($callable, self::$loaders, true)) {
```

Majkl578:
The name of the method is horrible though, anything better? `registerNewLoader`? `registerUniqueLoader`?

lcobucci:
> The name of the method is horrible though, anything better? `registerNewLoader`? `registerUniqueLoader`? `tryRegisterLoader`?

`registerUniqueLoader` sounds good to me

jrjohnson:
Thanks for the feedback @Majkl578 and @lcobucci. I made these changes. 
@Ocramius had suggested the name of the method originally, but I agree `registerUniqueLoader` is better.

tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
```diff
@@ -168,4 +168,25 @@ public function testResetClearsRegisteredAutoloaderFailures() : void
 
         self::assertSame(2, $failures);
     }
+
+    /**
+     * @runInSeparateProcess
+     */
+    public function testRegisterLoaderIfNotExistsOnlyRegisteresSameLoaderOnce() : void
+    {
+        $className = 'autoloadedClass' . random_int(10, 100000);
```
morozov: Does it really have to be random? It makes it impossible to reproduce the test case consistently. Please use a fixed class name.

tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
```diff
@@ -168,4 +168,25 @@ public function testResetClearsRegisteredAutoloaderFailures() : void
...
+        AnnotationRegistry::reset();
+        $i = 0;
+        $autoLoader = function () use (&$i) : bool {
+            $i += 1;
+            return false;
+        };
```
morozov: Not a big deal but I'd prefer to use PHPUnit mocks where possible instead of adding logic to the test. The mocking framework is tested by its own tests but this logic isn't.

The above could be implemented as:
```php
$autoLoader = self::createPartialMock(stdClass::class, ['__invoke']);
$autoLoader->expects($this->once())->method('__invoke');
```

tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
```diff
@@ -168,4 +168,25 @@ public function testResetClearsRegisteredAutoloaderFailures() : void
...
+        AnnotationRegistry::registerUniqueLoader($autoLoader);
+        AnnotationRegistry::registerUniqueLoader($autoLoader);
+        AnnotationRegistry::registerUniqueLoader($autoLoader);
```
morozov: Is this test about multiple registrations of the same loader or about the loader having returned `false` shouldn't be called again? If it's about the former, then there should be two calls to `AnnotationRegistry::registerUniqueLoader()` and one to  `AnnotationRegistry::loadAnnotationClass()`, not three of each.

tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
stof: well, the fact that the mocking framework is tested does not mean that your own usage performs the right logic with it. And so this offers no guarantee that the test being written tests things properly either.

tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
stof: return false is just because the loader itself does not define the class (and the AnnotationRegistry expects the loader to return false when it does not make the class exist after the call)

tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
jrjohnson: I don't mind changing this, but I copied it from this file in `testStopCallingLoadersAfterClassIsFound`.

tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
jrjohnson: I have no problem changing this, I copied it from other tests in this file. Shouldn't it be the same for consistency?

tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
morozov: The return value of `false` is fine. I'm asking for a minimal sufficient set of method calls. The fact that `AnnotationRegistry::registerUniqueLoader()` is called more than twice and `AnnotationRegistry::loadAnnotationClass()` is called more than once seems redundant. Isn't it?

tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
morozov: Unless there's a reason the test needs to have a random class name, please change it to a non-random one.

tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
morozov: > well, the fact that the mocking framework is tested does not mean that your own usage performs the right logic with it.

True.

> And so this offers no guarantee that the test being written tests things properly either.

True.

It just doesn't look right to implement mocking a callable and counting calls while this is already implemented in a tool which is designed for exactly that purpose.

jrjohnson:
@morozov thanks for the feedback. I've made these changes.

tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php
```diff
@@ -168,4 +168,19 @@ public function testResetClearsRegisteredAutoloaderFailures() : void
...
     }
+
+    /**
+     * @runInSeparateProcess
+     */
+    public function testRegisterLoaderIfNotExistsOnlyRegisteresSameLoaderOnce() : void
+    {
+        $className = 'autoloadedClassThatDoesNotExist';
+        AnnotationRegistry::reset();
+        $autoLoader = self::createPartialMock(\stdClass::class, ['__invoke']);
```
Ocramius: Woah, didn't know this one, awesome!

Ocramius:
:ship:

PR #164, opened 2017-12-10
Use Null Coalesce Operator

carusogabriel:


mikeSimonson:
@carusogabriel And thanks again

PR #171, opened 2018-02-03
Enhancement: Use class keyword

localheinz:
This PR

* [x] uses the `class` keyword instead of using string literals when referencing classes


localheinz:
Maybe fixing with #177.

PR #174, opened 2018-02-03
Enhancement: Use stages on Travis

localheinz:
This PR

* [x] uses build stages on Travis

Blocks #177.

💁‍♂️ For reference, see https://docs.travis-ci.com/user/build-stages/.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,31 @@
-language: php
-
+dist: trusty
 sudo: false
-
-cache:
-  directory:
-    - $HOME/.composer/cache
+language: php
 
 php:
   - 7.1
   - 7.2
   - nightly
 
-matrix:
-  include:
-    - php: 7.1
-      env: PHPSTAN=1
+cache:
+  directories:
+    - $HOME/.composer/cache
 
-before_script:
-    - composer install -n
-    - if [[ $PHPSTAN = 1 ]]; then composer require --dev phpstan/phpstan:^0.7; fi
+before_install:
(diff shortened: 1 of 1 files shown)


Majkl578:
Can you please use same setup as we use in other Doctrine repos where Stages are already introduced? I don't really like having different setups for different projects... Thanks.

localheinz:
@Majkl578 

Sure, using [`doctrine/migrations`](https://github.com/doctrine/migrations/blob/7045dd2c2e5db313de4c0434b09b59a94281254c/.travis.yml) as a basis then, although I have to admit I don't find it very readable.

Majkl578:
> I have to admit I don't find it very readable.

The same could be said about the proposed change here - i.e. I've never seen & or* in the config. I'm not opposed to improvements, but let's keep it consistent if possible and propose changes afterwards.

localheinz:
@Majkl578 

I can recommend http://blog.daemonl.com/2016/02/yaml.html, then!

localheinz:
@Majkl578 

I can recommend http://blog.daemonl.com/2016/02/yaml.html, then! 🤓

.travis.yml
```diff
@@ -1,25 +1,31 @@
...
+  - ./vendor/bin/phpunit
+
+jobs:
+  allow_failures:
+    - php: nightly
+
+  include:
+    - stage: Lint
+      before_script:
+        - travis_retry composer require --dev --prefer-dist --prefer-stable phpstan/phpstan:^0.7
```
lcobucci: We need to upgrade phpstan to `^0.9`!

lcobucci:
@localheinz 🚢 thanks!

localheinz:
Thank you, @lcobucci and @Majkl578!

PR #175, opened 2018-02-03
Fix: Cache configuration

localheinz:
This PR

* [x] fixes the cache configuration for Travis

💁‍♂️ For reference, see https://docs.travis-ci.com/user/caching/#Arbitrary-directories.


localheinz:
Fixing in #174.

PR #176, opened 2018-02-03
Fix: Disable xdebug as early as possible

localheinz:
This PR

* [x] disables `xdebug` as early as possible


localheinz:
Fixing in #174.

PR #187, opened 2018-03-24
Add support for class name constants that resolve to imported classes

andy-shea:
Currently annotations that accept a parameter of a given type can be
used via:

```
use Foo\Bar\SomeClass;

...

@SomeAnnotation(SomeClass::class)
protected $property;
```

This works fine, but as a minor improvement we can drop the requirement
for `::class` and simply resolve the type against the imported classes
that Doctrine already tracks (similar to how annotations are resolved):

```
@SomeAnnotation(SomeClass) protected $property;
```


Ocramius:
This is not really feasible because this introduces ambiguity between class
names and constant names, and also steers away from a well known PHP
feature. Saving few chars here is not a big win.

andy-shea:
Agreed

PR #197, opened 2018-05-02
Point link to new docs location

greg0ire:


README.md
```diff
@@ -10,7 +10,7 @@ Docblock Annotations Parser library (extracted from [Doctrine Common](https://gi
 
 ## Documentation
 
-See the [doctrine-project website](http://docs.doctrine-project.org/projects/doctrine-common/en/latest/reference/annotations.html).
+See the [doctrine-project website](https://www.doctrine-project.org/projects/doctrine-annotations/en/latest/index.html).
```
morozov: Really, in 2018, we still use the `www.` prefix? :smile:

Majkl578: Yes, otherwise CDN doesn't work... Root records cannot contain CNAMEs unfortunately. :|

PR #198, opened 2018-05-06
Fix package name to  doctrine/annotations

jungleran:
There is no doctrine/annotation package, the right package name is
doctrine/annotations.


lcobucci:
@jungleran 🚢 thanks!

PR #199, opened 2018-05-07
Drop SimpleAnnotationReader

### LINKED ISSUE #87: DCOM-295: Deprecate the SimpleAnnotationReader
_From @doctrinebot on August 31, 2015 5:3_

Jira issue originally created by user @ocramius:

The `SimpleAnnotationReader` is often causing confusion and brings little advantage over the more useful and generic `AnnotationReader`.

Getting rid of it will remove some of this confusion and make the mechanisms for loading/reading annotations more clear to `doctrine/annotations` users.

_Copied from original issue: doctrine/common#607_

Majkl578:
closes #87


Ocramius:
Question: do we want to enforce importing annotations (as in "stop looking for same-namespace annotations") to remove the ambiguity in resolved symbols? That was the biggest problem with this reader: too much implicit resolution.

Majkl578:
> stop looking for same-namespace annotations

This may be a bit problematic since same-namespace `use`s are considered superfluous by IDEs and CS.

Honestly I have little experience with this reader in real world since it was always too magical. What I hate most on this reader is the fact that you can register multiple namespaces _in which to look for annotations_. Then comes the WTF - colliding names are ignored, first one is used:
https://github.com/doctrine/annotations/blob/b4fde48ffe28bf766077f8d41b5d23049a0687a8/lib/Doctrine/Common/Annotations/DocParser.php#L699-L706

Do we have any statistics for SimpleAnnotationReader usage? Would this even be reason to keep it?
Imho imported annotations should stay, same-namespace lookup should stay too (for the reason above), but _namespaces to search in_ should be dropped.

Ocramius:
> same-namespace `use`s are considered superfluous by IDEs and CS.

Agreed. I'm mostly talking from experience in dealing with bug reports, where people type in something like `@Collection` and have a `Collection` in their current namespace. 

This is the scope of my original question.

On the other side, this is consistent with PHP's import and name resolution semantics, so I may just be going nowhere with it.

 > you can register multiple namespaces

No more: burn it with napalm.

 > Do we have any statistics for SimpleAnnotationReader usage? Would this even be reason to keep it?

We made the mistake of basing the ORM simplistic setup on it (including many of the docs). The replacement of the annotations can be automated (or at least reported), so we don't need to concern ourselves with it. Let's just kill it.

Given a `SimpleAnnotationReader`, we can build a tool that reports replacements to be performed.

Ocramius (approved):
I think there's nothing else to be done here, and this is good to go from my perspective :+1:

Majkl578:
Good thing is that it's already gone from ORM dev-master. :tada: 

Following PHP resolution rules is the way to go, the least confusing one.
Once we have 2.0 ready for alpha/beta, we can run some survey to see whether missing SimpleAnnotationReader behavior is a blocker for someone or not.

> Given a `SimpleAnnotationReader`, we can build a tool that reports replacements to be performed.

Sounds good to me. :+1:

Majkl578:
Alright, let's move on. :shipit:

PR #206, opened 2018-05-09
Add Scrutinizer CI

Majkl578:
Basically copied from Migrations and CS condition removed (not used in this repo, yet) and added coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,20 +11,47 @@ cache:
     - $HOME/.composer/cache
 
 before_install:
-  - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || echo "xdebug not available"
-  - composer self-update
+  - |
+    if [ "x$COVERAGE" != "xyes" ]; then
+      mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || true
+    fi
+  - travis_retry composer self-update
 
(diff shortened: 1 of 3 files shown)


Majkl578:
Didn't find any way to disable phpcs run (and [Patches](https://scrutinizer-ci.com/g/doctrine/annotations/inspections/ddd90ecc-f03e-427e-b333-9a9729edc511/patches)) so let's just ignore it until we add CS later.

jwage (approved):

.travis.yml
```diff
@@ -11,20 +11,47 @@ cache:
...
+    fi
+  - travis_retry composer self-update
 
 install: travis_retry composer update --prefer-dist
 
+before_script:
+  - |
+    if [ "x$COVERAGE" == "xyes" ] && [[ ! $(php -m | grep -si xdebug) ]]; then
+      echo "xdebug is required for coverage"
+      exit 1
```
stof: using `exit` in a command put in the .travis.yml will exit the whole build script, which will break the Travis output (as reporting for this command will not be done, as well as following tasks).

Majkl578: First time I hear about it. :thinking: What would you suggest as replacement here?

stof: Well, there used to be a warning in the Travis doc about it in the past, but I cannot find it anymore (they reorganized their doc in the meantime).

I thinkk you can use `travis_terminate 1` for that (which is a command which will exit with `1` as exit code, and so become a failure).

PR #209, opened 2018-05-21
Change Annotation::value typehint to mixed

malarzm:
`string` is wrong as it can also be `null` or an `array`. Merging this (and tagging a new bugfix release) would help me with https://github.com/doctrine/mongodb-odm/pull/1803 :)

### DIFF
--- a/lib/Doctrine/Annotations/Annotation.php
+++ b/lib/Doctrine/Annotations/Annotation.php
@@ -15,7 +15,7 @@ class Annotation
     /**
      * Value property. Common among all derived classes.
      *
-     * @var string
+     * @var mixed
      */
     public $value;
 


lib/Doctrine/Annotations/Annotation.php
```diff
@@ -14,8 +14,6 @@ class Annotation
 {
     /**
      * Value property. Common among all derived classes.
-     *
-     * @var string
```
Ocramius: Should be annotated as `mixed` instead of being skipped

alcaeus: Isn't `mixed` implied when it's not further specified?

malarzm: This was also my assumption. Unless we're explicitly saying `mixed` just to show we care enough :D

Ocramius: > Isn't `mixed` implied when it's not further specified?

"implied" is something we no longer really allow. Even our coding standards prevent us from skipping parameters/return types. Possibly a missing CS rule, @carusogabriel?

malarzm:
@Ocramius `mixed` it is then :+1:

lib/Doctrine/Annotations/Annotation.php
carusogabriel: @Ocramius I’ll add some tests in our CS repo and discovered what should be fixed to avoid it :)

PR #211, opened 2018-06-16
[DocParser] Prevent reflection on ignored annotations when namespaces are present.

paul-m:
I'm working on some Drupal core stuff.

We use doctrine/common and doctrine/annotations.

Our classes have lots of annotations like `@param` and `@return` that we want to ignore. However, our systems end up hitting reflection with class names like `Drupal\Something\param`. This is bad because it means we have at least one `file_exists()` per reflection.

In this issue we've written a test to prove that our subsystems do this: https://www.drupal.org/project/drupal/issues/2557433

We also have issues to roll our own replacements for some pieces of the Doctrine annotation system: https://www.drupal.org/project/drupal/issues/2631202

However, we can't proceed because `DocParser` never ignores annotation tags if namespaces are in use.

This PR has a test which fails if `DocParser` tries to autoload a name that's in the ignore list, and then makes this test pass.

Thanks for considering it.

Drupal is currently stuck at doctrine/annotations v1.2.7, because we support PHP 5.5 and up. So if this fix could find its way back in time that would help us immensely. Thanks.

### DIFF
--- a/lib/Doctrine/Annotations/DocParser.php
+++ b/lib/Doctrine/Annotations/DocParser.php
@@ -667,6 +667,9 @@ final class DocParser
             $loweredAlias = strtolower($alias);
 
             if ($this->namespaces) {
+                if (isset($this->ignoredAnnotationNames[$name])) {
+                    return false;
+                }
                 foreach ($this->namespaces as $namespace) {
                     if ($this->classExists($namespace.'\\'.$name)) {
                         $name = $namespace.'\\'.$name;
--- a/tests/Doctrine/Tests/Annotations/DocParserTest.php
+++ b/tests/Doctrine/Tests/Annotations/DocParserTest.php
@@ -1379,11 +1379,16 @@ DOCBLOCK;
 
     }
 
+    /**
+     * Ensure annotations can be ignored when namespaces are present.
+     *
+     * DocParser should never use class_exists() on an ignored annotation,
+     * including cases where namespaces are set.
+     */
     public function testIgnoredAnnotationSkippedBeforeReflection() {
         $annotation = 'neverReflectThis';
         $parser = new DocParser();
         $parser->setIgnoredAnnotationNames([$annotation => TRUE]);
-//        $parser->setIgnoreNotImportedAnnotations(TRUE);
         $parser->addNamespace('\\Arbitrary\\Namespace');
 
         // Register our class loader which will fail if the parser tries to
--- a/tests/Doctrine/Tests/Annotations/DocParserTest.php
+++ b/tests/Doctrine/Tests/Annotations/DocParserTest.php
@@ -1378,6 +1378,29 @@ DOCBLOCK;
         self::assertCount(1, $result);
 
     }
+
+    public function testIgnoredAnnotationSkippedBeforeReflection() {
+        $annotation = 'neverReflectThis';
+        $parser = new DocParser();
+        $parser->setIgnoredAnnotationNames([$annotation => TRUE]);
+//        $parser->setIgnoreNotImportedAnnotations(TRUE);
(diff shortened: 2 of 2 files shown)


alcaeus:
It's absolutely a BC break: I can no longer have an annotation named `Vendor\Project\Param` if I mark `param` as excluded. If we decide that we want to apply the ignore list to namespaced annotations as well, we need to do so in a new major release and provide a forward compatibility layer since it's a significant change.

Given the widespread use of the annotation library, we're pretty much guaranteed to break someones application with a change like this in a minor release, so this approach is a clear 👎  for me.

If we want to make this feature backwards compatible, we need to add an option to mark an ignored annotation name as ignored for namespaces as well instead of applying all existing ignored annotation names to namespaces as well.

Ocramius:
@alcaeus can you think of a scenario we can add as a test case? I think the performance optimisation is valid, but requires the scenario to be factored in to prevent a BC break.

Majkl578:
Ad

> So there's a use-case in favor of the change, already in use. It's just that we're [filling everyone's opcache](https://www.drupal.org/project/drupal/issues/2631202#comment-11209997) with lookups for `\Drupal\Something\Entity\param`.

Note that `class_exists()` misses do not fill OPcache. It only caches PHP code itself.
```php
var_dump(memory_get_peak_usage());
var_dump(opcache_get_status()['memory_usage']['used_memory']);
var_dump(opcache_get_status()['opcache_statistics']);

for ($i = 0; $i < 100000000; $i++) class_exists('invalid' . $i);

var_dump(memory_get_peak_usage());
var_dump(opcache_get_status()['memory_usage']['used_memory']);
var_dump(opcache_get_status()['opcache_statistics']);
```
```
$ php -d opcache.enable_cli=1 -d opcache.memory_consumption=8 -d opcache.interned_strings_buffer=1 /tmp/test.php 
int(532736)
int(3139600)
array(13) {
  ["num_cached_scripts"]=>
  int(1)
  ["num_cached_keys"]=>
  int(1)
  ["max_cached_keys"]=>
  int(16229)
  ["hits"]=>
  int(0)
  ["start_time"]=>
  int(1529516406)
  ["last_restart_time"]=>
  int(0)
  ["oom_restarts"]=>
  int(0)
  ["hash_restarts"]=>
  int(0)
  ["manual_restarts"]=>
  int(0)
  ["misses"]=>
  int(1)
  ["blacklist_misses"]=>
  int(0)
  ["blacklist_miss_ratio"]=>
  float(0)
  ["opcache_hit_rate"]=>
  float(0)
}
int(532736)
int(3139600)
array(13) {
  ["num_cached_scripts"]=>
  int(1)
  ["num_cached_keys"]=>
  int(1)
  ["max_cached_keys"]=>
  int(16229)
  ["hits"]=>
  int(0)
  ["start_time"]=>
  int(1529516406)
  ["last_restart_time"]=>
  int(0)
  ["oom_restarts"]=>
  int(0)
  ["hash_restarts"]=>
  int(0)
  ["manual_restarts"]=>
  int(0)
  ["misses"]=>
  int(1)
  ["blacklist_misses"]=>
  int(0)
  ["blacklist_miss_ratio"]=>
  float(0)
  ["opcache_hit_rate"]=>
  float(0)
}

```

And regarding APC (mentioned in the linked comment) - APC is dead for many years. It was always broken and buggy, leading to unfamous segfaults, crashes etc.
Please don't use it and don't support it. Use and support OPcache. If you need custom cache, then APCu (if race conditions are not an issue for you) or Redis.

Ocramius:
Ok, closing here as `invalid`. In order to make this patch viable, we'll need various tests around `@param` (global annotation), `@param` (imported from a namespace, therefore not ignored) and `@foo\param` (referenced via FQN).

Once those are factored in, we can discuss this on a new patch.

PR #213, opened 2018-06-28
Check if ini_get is not disabled before using it

ruester:
ini_get may often be disabled in production environments, causing an exception to be thrown (since ini_get in this case returns NULL and NULL == 0 is true).
This change checks that ini_get is usable (otherwise function_exists will return FALSE) before using it. If we cannot use ini_get we assume that the extensions are configured correctly.


Ocramius:
This seems extremely unrealistic: why would `ini_get()` be disabled?

I'm more pointed towards letting `ini_get()` crash hard here, because any internal PHP function may be disabled, and that's a slippery slope to not being able to trust even the most basic language features.

ruester:
Well at least `phpinfo`, `ini_get_all` or `ini_set` is often recommended to be disabled for security reasons in production environments. http://sandbox.onlinephpfunctions.com is one example for many disabled functions but thats a different case...
Other frameworks like Smarty or PHP Solarium are also checking if `ini_get` is allowed.

Another solution would be to change the conditon to `ini_get(...) === "0"` since `ini_get` will always return strings (if not disabled). This way there would only be a warning but the doctrine annotations module would not fail.

PR #226, opened 2018-11-07
Add support getConstantAnnotations

Yurunsoft:


Ocramius (changes requested):
Hey @yurunsoft,

In order to consider this patch, we'd first need extensive testing for:

 * Annotations on constants on classes
 * Annotations on constants on interfaces
 * Annotations on constants on traits
 * Verifying that the target of an annotation allows/prevents usage on constants (expect exceptions)

jkufner:
@Ocramius Shouldn't be enough to copy tests regarding properties annotations (and `s/Property/Constant/`)?

Ocramius:
@jkufner sure, send a patch to this branch :-)

jkufner:
See https://github.com/doctrine/annotations/pull/263. I've cherry-picked commits of this pull request onto 1.7 branch, added some tests and few details around. Since it does not introduce BC break, I don't see a reason to wait until 2.0.

alcaeus:
Closing in favour of #263.

PR #230, opened 2018-12-07
Added PHP CodeSniffer tags to globally ignored annotations list.

watari:
Added `@codingStandardsIgnoreStart` and `@codingStandardsIgnoreEnd` from PHP CodeSniffer to globally ignored annotations list.

Sorry for not including this tags in previous PR. At that moment I don't know about this tags :)

### DIFF
--- a/lib/Doctrine/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Annotations/AnnotationReader.php
@@ -86,7 +86,10 @@ class AnnotationReader implements Reader
         // Symfony 3.3 Cache Adapter
         'experimental' => true,
         // Slevomat Coding Standard
-        'phpcsSuppress' => true
+        'phpcsSuppress' => true,
+        // PHP CodeSniffer
+        'codingStandardsIgnoreStart' => true,
+        'codingStandardsIgnoreEnd' => true,
     ];
 
     /**


lib/Doctrine/Annotations/AnnotationReader.php
```diff
@@ -86,7 +86,9 @@ class AnnotationReader implements Reader
         // Symfony 3.3 Cache Adapter
         'experimental' => true,
         // Slevomat Coding Standard
-        'phpcsSuppress' => true
+        'phpcsSuppress' => true,
+        // PHP CodeSniffer
+        'codingStandardsIgnoreStart' => true, 'codingStandardsIgnoreEnd' => true
```
lcobucci: Please break the line here.

watari: @lcobucci , are you sure about this? 
Above my changes tags from one lib is grouped in same way ([example](https://github.com/doctrine/annotations/blob/320af0032a927aece85014c9594badb0df631556/lib/Doctrine/Annotations/AnnotationReader.php#L77))

Majkl578: Not all of them, also it'll be changed once we make it follow our CS. :)

watari: Ok). Done

lcobucci:
@watari thanks :+1:

PR #177, opened 2018-02-03
Enhancement: Use doctrine/coding-standard

localheinz:
❗️ Blocked by #174.

This PR

* [x] requires `doctrine/coding-standard` and adds a `coding standard` stage
* [ ] runs `phpcbf` (so far, **3x**)

PR #235, opened 2019-01-23
Update AnnotationReader's metadata parser to use list of known ignored annotation 

sanmai:
A parser used to collect parsing metadata basically ignores a known list of globally ignored annotation names, goes through loops checking annotations that do not need any checking. Let's fix that by communicating the list during initialization.

E.g. these checks should not be evaluated for known ignored annotations:

https://github.com/doctrine/annotations/blob/386d3bb15db36fd2a381d90fe5ab9287984aba42/lib/Doctrine/Annotations/DocParser.php#L682-L690

Please let me know if this change needs a test. One way to test is to rig an additional autoloader that would fail on attempt to load a class named as one of the commonly-ignored annotations like `@see`.

### DIFF
--- a/tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
+++ b/tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
@@ -408,6 +408,28 @@ abstract class AbstractReaderTest extends TestCase
         self::assertEmpty($reader->getPropertyAnnotations($class->getProperty('foo')));
     }
 
+    public function testGloballyIgnoredAnnotationNotIgnored() : void
+    {
+        $reader = $this->getReader();
+        $class  = new \ReflectionClass(Fixtures\ClassDDC1660::class);
+
+        $testLoader = static function (string $className) : bool {
+            if ($className === 'since') {
+                throw new \InvalidArgumentException('Globally ignored annotation names should never be passed to an autoloader.');
+            }
+
+            return false;
+        };
+
+        spl_autoload_register($testLoader, true, true);
+
+        try {
+            self::assertEmpty($reader->getClassAnnotations($class));
+        } finally {
+            spl_autoload_unregister($testLoader);
+        }
+    }
+
     public function testAnnotationEnumeratorException()
     {
         $reader     = $this->getReader();
--- a/lib/Doctrine/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Annotations/AnnotationReader.php
@@ -181,6 +181,7 @@ class AnnotationReader implements Reader
 
         $this->preParser->setImports(self::$globalImports);
         $this->preParser->setIgnoreNotImportedAnnotations(true);
+        $this->preParser->setIgnoredAnnotationNames(self::$globalIgnoredNames);
 
         $this->phpParser = new PhpParser;
     }


sanmai:
No problem, I'll add a test later. Meanwhile I'm open to suggestions about exact wording.

sanmai:
@Ocramius please see if it is better now.

tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
```diff
@@ -403,9 +403,22 @@ public function testInvalidAnnotationButIgnored()
         $class  = new \ReflectionClass(Fixtures\ClassDDC1660::class);
 
         self::assertTrue(class_exists(Fixtures\Annotation\Version::class));
-        self::assertEmpty($reader->getClassAnnotations($class));
```
Ocramius: This change highlights a major BC break: if the patch cannot be applied without a change of an existing test, then it should likely target the next major release of the library.

sanmai: This patch can be definitely applied without this change, but with a separate test. Just give me a moment.

Ocramius: Thanks!

sanmai: OK, done. Should I squash commits into one? Else a squash merge is fine by me.

Ocramius: If possible, please squash, yes :-)

sanmai: Done.

tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
```diff
@@ -408,6 +408,26 @@ public function testInvalidAnnotationButIgnored()
         self::assertEmpty($reader->getPropertyAnnotations($class->getProperty('foo')));
     }
 
+    public function testGloballyIgnoredAnnotationNotIgnored()
```
lcobucci: ` : void` missing

tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
```diff
@@ -408,6 +408,26 @@ public function testInvalidAnnotationButIgnored()
...
+    {
+        $reader = $this->getReader();
+        $class  = new \ReflectionClass(Fixtures\ClassDDC1660::class);
+
+        spl_autoload_register($testLoader = function ($className) {
```
lcobucci: Let's move the definition of the function outside of `spl_autoload_register()` call, things will be more readable.

lcobucci: Also making it static and add type declaration to it:

```diff
- $testLoader = function ($className) {
+ $testLoader = static function (string $className) : bool {
```

tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
sanmai: I'm not sure how's that going in line with all ither code in this file. Isn't it better to put these things aside? Like there's #177 which is supposed to do exactly what you're asking, but all over all code.

tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
sanmai: Again, there's not a single place in the entire code base of the project where `static function`, and so is true for type declarations. I'm not saying they're not cool or something, just are not the purpose of this PR.

tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
lcobucci: @sanmai I understand your frustration here, these are **cosmetic** things and don't have much impact on the behaviour of the code. However we have an agreement among maintainers to try as much as possible to make sure our CS is applied everywhere, even when the project isn't following it properly (that's why we have automated incremental checks on ORM and DBAL, for example).

We haven't had the chance of applying the incremental checks here and applying things on the entire code base is a bit too much (as explained in #177), hence my comments.

I'm sorry for the confusion and will completely understand if you decide to not apply the suggestions in the comments, I'm just trying to ensure that **we** (Doctrine team) follow **our** agreements. A maintainer is able to edit your commit on your fork (if you hadn't changed the option while creating the PR).

I hope you'll understand the motivation and I thank you for your contribution.

tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
lcobucci: See https://github.com/doctrine/annotations/pull/235#discussion_r

tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
sanmai: Well, that's just me that spend a good amount of time to make sure that my code looks just like every other code, but if there's a consensus to write newer code in a more appropriate style, I'm good with that too. I'd change on my part but I'd like if these new guidelines would be somewhere at least. Then anyone else wouldn't need to spend the time so useless as I did. Are we on the same page with this?

sanmai: @lcobucci anything else?

tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
```diff
@@ -408,6 +408,28 @@ public function testInvalidAnnotationButIgnored()
         self::assertEmpty($reader->getPropertyAnnotations($class->getProperty('foo')));
     }
 
+    public function testGloballyIgnoredAnnotationNotIgnored() : void
+    {
+        $reader = $this->getReader();
+        $class  = new \ReflectionClass(Fixtures\ClassDDC1660::class);
+
+        $testLoader = static function ($className) : bool {
```
greg0ire:
```suggestion
        $testLoader = static function (string $className) : bool {
```

tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
greg0ire: > I'd change on my part but I'd like if these new guidelines would be somewhere, at least.

That one is documented here: https://www.doctrine-project.org/projects/doctrine-coding-standard/en/latest/reference/index.html#project-level-ruleset ("Always add native types where possible")

sanmai: Well, that's not exactly obvious, what do you think? How am I supposed to find it?..

A like to these guidelines from the README.md, or even from CONTRIBUTING.md, would be best.

greg0ire: I think you're right! I might make PRs to change that if you do not beat me to it.

lcobucci:
@sanmai @Ocramius @Majkl578 @greg0ire I've found an interesting thing while backporting this to `1.7`: the test passes even if `$this->preParser->setIgnoredAnnotationNames(self::$globalIgnoredNames);` isn't called :+1: 

Also interesting to mention it fails for `SimpleAnnotationReaderTest`, because it wrongly parses a lowercase annotation (`@version`).

Shall we mark it as an fix to v2.0 only?

tests/Doctrine/Tests/Annotations/AbstractReaderTest.php
sanmai: @greg0ire see #239

sanmai:
Actually I had this issue on a released version. Let me check what it was...

sanmai:
@lcobucci I got this fix backported to 1.6, see #240

PR #240, opened 2019-01-30
Update AnnotationReader's metadata parser to use list of known ignored annotation

sanmai:
A parser used to collect parsing metadata basically ignores a known list of globally ignored annotation names, goes through loops checking annotations that do not need any checking. Let's fix that by communicating the list during initialization.

This is a backported fix from #235.


tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
```diff
@@ -408,6 +408,30 @@ public function testInvalidAnnotationButIgnored()
...
+                throw new \InvalidArgumentException('Globally ignored annotation names should never be passed to an autoloader.');
+            }
+
+            return false;
+        };
+
+        \Doctrine\Common\Annotations\AnnotationRegistry::registerLoader($testLoader);
+
+        try {
+            self::assertNotNull($reader->getClassAnnotations($class));
```
sanmai: Unlike #235, in this case `getClassAnnotations()` returns non-empty array. I guess `testInvalidAnnotationButIgnored` is broken, but fixing it proper should be done in a different PR.

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
```diff
@@ -408,6 +408,30 @@ public function testInvalidAnnotationButIgnored()
         self::assertEmpty($reader->getPropertyAnnotations($class->getProperty('foo')));
     }
 
+    public function testGloballyIgnoredAnnotationNotIgnored() : void
+    {
+        $reader = $this->getReader();
+        $class  = new \ReflectionClass(Fixtures\ClassDDC1660::class);
+
+        $active = true;
```
sanmai: Another way to disable this autoloader would be using a `ReflectionProperty` like in `AnnotationRegistryTest`, not sure if it is any better.

derrabus:
Let's close the PR. I think it's obsolete.

PR #242, opened 2019-02-19
Extract implicitly ignored annotation names to separate class

Majkl578:
* To improve readability and maintainability of this list.
* Categorized, each category ordered alphabetically.
* Will be further used in Annotations 2.0.

### DIFF
--- a/lib/Doctrine/Annotations/ImplicitlyIgnoredAnnotationNames.php
+++ b/lib/Doctrine/Annotations/ImplicitlyIgnoredAnnotationNames.php
@@ -1,0 +1,137 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Annotations;
+
+/**
+ *  A list of annotations that are implicitly ignored during the parsing process.
+ *
+ *  All names are case sensitive.
+ */
+final class ImplicitlyIgnoredAnnotationNames
+{
+    private const Reserved = [
+        'Annotation' => true,
+        'Attribute'  => true,
+        'Attributes' => true,
+        /* Can we enable this? 'Enum' => true, */
+        'Required'   => true,
+        'Target'     => true,
+    ];
+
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Annotations/ImplicitlyIgnoredAnnotationNames.php
```diff
@@ -0,0 +1,137 @@
...
+ *
+ *  All names are case sensitive.
+ */
+final class ImplicitlyIgnoredAnnotationNames
+{
+    private const Reserved = [
+        'Annotation' => true,
+        'Attribute'  => true,
+        'Attributes' => true,
+        /* Can we enable this? 'Enum' => true, */
```
SenseException: What about this question? This is from 5 years ago. I think it can be removed.

Majkl578: Didn't really want to do side-effects during just _extracting_ the annotations. Also in 2.0 these reserved annotations will likely not be needed anymore.

lib/Doctrine/Annotations/ImplicitlyIgnoredAnnotationNames.php
```diff
@@ -0,0 +1,137 @@
...
+        'codingStandardsIgnoreStart' => true,
+        'codingStandardsIgnoreEnd'   => true,
+    ];
+
+    private const SlevomatCodingStandard = [
+        'phpcsSuppress' => true,
+    ];
+
+    public const LIST = self::Reserved
+        + self::WidelyUsedNonStandard
```
Ocramius: Ooooooh, I like this style of merging stuff in constants!

Ocramius:
This can go to `master` into `1.x` as-is.

Ocramius:

PR #243, opened 2019-02-19
fix ignored annotations with namespace alias

dbu:
when i set a namespace to ignored but use a namespace alias, there was no check if the class actually exists. instead of ignoring, we ended up with an exception that the annotation could not be autoloaded.


alcaeus:
@dbu 1.7 is an upcoming releases, the latest stable one is 1.6.0. Can you fix this in 1.6? Also, this PR is lacking some tests for that functionality.

dbu:
sorry for that. i noticed that master is 2.0 and just assumed that 1.7 is released. closing this in favor of #244 where i also add a test (that caught an edge case of the edge case).

PR #244, opened 2019-02-19
fix ignored annotations with namespace alias

dbu:
when i set a namespace to ignored but use a namespace alias, there was no check if the class actually exists. instead of ignoring, we ended up with an exception that the annotation could not be autoloaded.

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -710,10 +710,11 @@ final class DocParser
                     }
                 }
             } elseif (isset($this->imports[$loweredAlias])) {
-                $found = true;
-                $name  = (false !== $pos)
-                    ? $this->imports[$loweredAlias] . substr($name, $pos)
-                    : $this->imports[$loweredAlias];
+                $namespace = ltrim($this->imports[$loweredAlias], '\\');
+                $name = (false !== $pos)
+                    ? $namespace . substr($name, $pos)
+                    : $namespace;
+                $found = $this->classExists($name);
             } elseif ( ! isset($this->ignoredAnnotationNames[$name])
                 && isset($this->imports['__NAMESPACE__'])
                 && $this->classExists($this->imports['__NAMESPACE__'] . '\\' . $name)
--- a/tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
@@ -12,6 +12,7 @@ use Doctrine\Tests\Common\Annotations\Fixtures\ClassWithPhpCsSuppressAnnotation;
 use Doctrine\Tests\Common\Annotations\Fixtures\IgnoredNamespaces\AnnotatedAtClassLevel;
 use Doctrine\Tests\Common\Annotations\Fixtures\IgnoredNamespaces\AnnotatedAtMethodLevel;
 use Doctrine\Tests\Common\Annotations\Fixtures\IgnoredNamespaces\AnnotatedAtPropertyLevel;
+use Doctrine\Tests\Common\Annotations\Fixtures\IgnoredNamespaces\AnnotatedWithAlias;
 
 class AnnotationReaderTest extends AbstractReaderTest
 {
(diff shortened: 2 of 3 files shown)


lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -705,10 +705,11 @@ private function Annotation()
                     }
                 }
             } elseif (isset($this->imports[$loweredAlias])) {
-                $found = true;
                 $name  = (false !== $pos)
                     ? $this->imports[$loweredAlias] . substr($name, $pos)
                     : $this->imports[$loweredAlias];
+                $name = ltrim($name,'\\');
```
dbu: this is duplicated below, but we need it here to make $this->classExists work

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -940,10 +941,10 @@ private function Constant()
                     break;
 
                 case isset($this->imports[$loweredAlias]):
-                    $found     = true;
                     $className = (false !== $pos)
                         ? $this->imports[$loweredAlias] . substr($className, $pos)
                         : $this->imports[$loweredAlias];
+                    $found     = class_exists($className) || interface_exists($className);
```
dbu: similar issue with constants. i found no tests for this but think we should handle it the same way? i can also remove this line again if you prefer.

tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
```diff
@@ -112,6 +113,21 @@ public function testPropertyAnnotationIsIgnored()
         self::assertEmpty($reader->getPropertyAnnotations($ref->getProperty('property')));
     }
 
+    /**
+     * @group 45
```
dbu: not sure what this is for, but copied from the other ignored annotations tests.

dbu:
travis error is unrelated, see #245

tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
Ocramius: This is for the issue ID: this would be 244 in this case

lib/Doctrine/Common/Annotations/DocParser.php
Ocramius: Please avoid overwriting the variable: wrap the previous expression, if possible

lib/Doctrine/Common/Annotations/DocParser.php
Ocramius: Could you rephrase this point? I didn't understand the comment.

Ocramius: This addition requires additional tests for:

 * class not found
 * interface not found
 * class found
 * interface found

dbu: i will revert this change as i am not sure about it. i suspect the same issue to be here for the constants handling as we have for annotations handling, because it also handles namespace alias separately.

tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
dbu: ah, that makes sense :-D 
fixing that number

lib/Doctrine/Common/Annotations/DocParser.php
dbu: did it differently, okay like this?

dbu:
okay, updated to the feedback.

Majkl578:
As per existing branching rules (which are still not formalized), this should not target 1.6. :/

alcaeus:
@Majkl578 we haven't discussed this, and in an effort to do less cherry-picking and more merging I suggested that the PR be opened against 1.6, as #243 was previously targeting master. With that in mind, I'd no longer ask @dbu to retarget the PR but rather take care of this ourselves when merging the PR and preparing for a new release.

dbu:
hey, i don't mind doing the PR against the right branch again, so long as you can agree on which branch that should be ;-)

dbu:
can i do something to get this merged into the appropriate branch?

alcaeus:
@Majkl578 go ahead and merge this to whatever branch you deem appropriate - I'll let you take care of ensuring this ends up in a 1.6 patch release.

greg0ire:
@dbu the tests no longer pass, please have a look.

dbu:
thanks for wrapping this up @alcaeus ! i am currently a bit busy and did not get around to do it.

PR #245, opened 2019-02-19
build with php 7.3

dbu:
and no build with nightly, which is php 8 and not supported by anything yet...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,8 @@ cache:
 php:
   - 7.1
   - 7.2
-  - nightly
+  - 7.3
+  - 7.4snapshot
 
 matrix:
   include:


alcaeus:
FWIW, you can use `7.4snapshot` to test PHP 7.4.

dbu:
added 7.4snapshot in addition to 7.3

alcaeus:
Note: holding off on merging to discuss target branches internally.

Majkl578:
This has already been done in master btw so it should be cherry-picked either way.

dbu:
i can do a new pull request where i cherry-pick the commit from master if it matters. afaik git does not track cherry-picks over new commits so that won't prevent a merge conflict automatically. and if there is anything else in that merge commit we risk troubles. 

or do whatever you think is right :-D i just wanted to help make the build for the currently active version green again...

dbu:
is this still relevant or did you solve it differently?

greg0ire:
It was solved in 166cbdb0

PR #246, opened 2019-02-23
Drop base `Annotation` ancestor

Majkl578:
This PR drops the `Doctrine\Annotations\Annotation` base annotation class.

It was an inheritance abuse that basically existed only for convenience:
* to implicitly define default `$value` property,
* to behave like `stdClass` and dynamically set undefined properties (without validation).


lib/Doctrine/Annotations/Annotation.php
```diff
@@ -1,62 +0,0 @@
...
-
-/**
- * Annotations class.
- *
- * @author Benjamin Eberlei <kontakt@beberlei.de>
- * @author Guilherme Blanco <guilhermeblanco@hotmail.com>
- * @author Jonathan Wage <jonwage@gmail.com>
- * @author Roman Borschel <roman@code-factory.org>
- */
-class Annotation
```
Ocramius: I don't think the base class should be removed, as it makes the BC break muvh bigger than needed: a deprecation would probably be better

Ocramius (changes requested):

lib/Doctrine/Annotations/Annotation.php
Majkl578: I was rather thinking about deprecating it in 1.7 and going with less burdens in 2.0. Mostly because this class is rarely used out there. But I may be wrong, GitHub search is not very helpful in finding usages.

alcaeus: Deprecating it in 1.7 would be the first step before dropping it from 2.0. We can then see what people say and either decide to revert the deprecation or go along with the removal.

Ocramius (approved):
LGTM. The question that remains before merge is if we should stop plans for `1.7.0` and make `master` `2.0.0` instead. I'd delete the milestone and re-assign everything to `2.0.0` then.

alcaeus:
Master is 2.0.x-dev already.

Ocramius:
Right, I'm moving all of 1.7.0 stuff into 2.0.0 then.

alcaeus:
Please don’t - 1.7 will be necessary to deprecate features that will be dropped in 2.0 as per SemVer.

Ocramius:
As per discussion with @alcaeus and @Majkl578, `master` is already `2.0.0`, so this can be merged nao.

PR #247, opened 2019-02-23
Introduce annotation metadata

Majkl578:
Replace nested untyped metadata arrays structure inside DocParser by proper object metadata classes with builder.
This idea is somewhat inspired by #75 but this approach uses separate PropertyMetadata instead of plain untyped array as well as being more strictly typed.

Metadata will later be used by the new parser as well.
This is just an internals refactoring so it's not a BC break. No existing tests were touched.

### DIFF
--- a/lib/Doctrine/Annotations/DocParser.php
+++ b/lib/Doctrine/Annotations/DocParser.php
@@ -3,10 +3,18 @@
 namespace Doctrine\Annotations;
 
 use Doctrine\Annotations\Annotation\Attribute;
+use Doctrine\Annotations\Metadata\AnnotationTarget;
+use Doctrine\Annotations\Metadata\Builder\AnnotationMetadataBuilder;
+use Doctrine\Annotations\Metadata\Builder\PropertyMetadataBuilder;
+use Doctrine\Annotations\Metadata\InternalAnnotations;
+use Doctrine\Annotations\Metadata\MetadataCollection;
+use Doctrine\Annotations\Metadata\TransientMetadataCollection;
 use ReflectionClass;
 use Doctrine\Annotations\Annotation\Enum;
 use Doctrine\Annotations\Annotation\Target;
 use Doctrine\Annotations\Annotation\Attributes;
+use function array_key_exists;
+use function array_keys;
 
 /**
  * A parser for docblock annotations.
@@ -118,93 +126,12 @@ final class DocParser
     /**
      * Hash-map for caching annotation metadata.
      *
-     * @var array
+     * @var MetadataCollection
      */
-    private static $annotationMetadata = [
-        'Doctrine\Annotations\Annotation\Target' => [
-            'is_annotation'    => true,
-            'has_constructor'  => true,
-            'properties'       => [],
-            'targets_literal'  => 'ANNOTATION_CLASS',
-            'targets'          => Target::TARGET_CLASS,
-            'default_property' => 'value',
-            'attribute_types'  => [
-                'value'  => [
-                    'required'  => false,
-                    'type'      =>'array',
-                    'array_type'=>'string',
-                    'value'     =>'array<string>'
-                ]
-             ],
-        ],
-        'Doctrine\Annotations\Annotation\Attribute' => [
-            'is_annotation'    => true,
-            'has_constructor'  => false,
-            'targets_literal'  => 'ANNOTATION_ANNOTATION',
-            'targets'          => Target::TARGET_ANNOTATION,
-            'default_property' => 'name',
-            'properties'       => [
-                'name'      => 'name',
-                'type'      => 'type',
-                'required'  => 'required'
-            ],
-            'attribute_types'  => [
-                'value'  => [
-                    'required'  => true,
-                    'type'      =>'string',
-                    'value'     =>'string'
-                ],
-                'type'  => [
-                    'required'  =>true,
-                    'type'      =>'string',
-                    'value'     =>'string'
-                ],
-                'required'  => [
-                    'required'  =>false,
-                    'type'      =>'boolean',
-                    'value'     =>'boolean'
-                ]
-             ],
-        ],
-        'Doctrine\Annotations\Annotation\Attributes' => [
-            'is_annotation'    => true,
-            'has_constructor'  => false,
-            'targets_literal'  => 'ANNOTATION_CLASS',
-            'targets'          => Target::TARGET_CLASS,
-            'default_property' => 'value',
-            'properties'       => [
-                'value' => 'value'
-            ],
-            'attribute_types'  => [
-                'value' => [
-                    'type'      =>'array',
-                    'required'  =>true,
-                    'array_type'=>'Doctrine\Annotations\Annotation\Attribute',
-                    'value'     =>'array<Doctrine\Annotations\Annotation\Attribute>'
-                ]
-             ],
-        ],
-        'Doctrine\Annotations\Annotation\Enum' => [
-            'is_annotation'    => true,
-            'has_constructor'  => true,
-            'targets_literal'  => 'ANNOTATION_PROPERTY',
-            'targets'          => Target::TARGET_PROPERTY,
-            'default_property' => 'value',
-            'properties'       => [
-                'value' => 'value'
-            ],
-            'attribute_types'  => [
-                'value' => [
-                    'type'      => 'array',
-                    'required'  => true,
-                ],
-                'literal' => [
-                    'type'      => 'array',
-                    'required'  => false,
-                ],
-             ],
-        ],
-    ];
+    private $metadata;
+
+    /** @var array<string, bool> */
+    private $nonAnnotationClasses = [];
 
     /**
      * Hash-map for handle types declaration.
@@ -224,7 +151,8 @@ final class DocParser
      */
     public function __construct()
     {
-        $this->lexer = new DocLexer;
+        $this->lexer    = new DocLexer;
+        $this->metadata = InternalAnnotations::createMetadata();
     }
 
     /**
@@ -468,85 +396,99 @@ final class DocParser
             ]);
         }
 
-        $class      = new \ReflectionClass($name);
+        $class      = new ReflectionClass($name);
         $docComment = $class->getDocComment();
 
-        // Sets default values for annotation metadata
-        $metadata = [
-            'default_property' => null,
-            'has_constructor'  => (null !== $constructor = $class->getConstructor()) && $constructor->getNumberOfParameters() > 0,
-            'properties'       => [],
-            'property_types'   => [],
-            'attribute_types'  => [],
-            'targets_literal'  => null,
-            'targets'          => Target::TARGET_ALL,
-            'is_annotation'    => false !== strpos($docComment, '@Annotation'),
-        ];
-
         // verify that the class is really meant to be an annotation
-        if ($metadata['is_annotation']) {
-            self::$metadataParser->setTarget(Target::TARGET_CLASS);
+        if (strpos($docComment, '@Annotation') === false) {
+            $this->nonAnnotationClasses[$name] = true;
+            return;
+        }
 
-            foreach (self::$metadataParser->parse($docComment, 'class @' . $name) as $annotation) {
-                if ($annotation instanceof Target) {
-                    $metadata['targets']         = $annotation->targets;
-                    $metadata['targets_literal'] = $annotation->literal;
+        $constructor       = $class->getConstructor();
+        $useConstructor    = $constructor !== null && $constructor->getNumberOfParameters() > 0;
+        $annotationBuilder = new AnnotationMetadataBuilder($name);
 
-                    continue;
-                }
+        if ($useConstructor) {
+            $annotationBuilder->withUsingConstructor();
+        }
 
-                if ($annotation instanceof Attributes) {
-                    foreach ($annotation->value as $attribute) {
-                        $this->collectAttributeTypeMetadata($metadata, $attribute);
-                    }
-                }
+        self::$metadataParser->setTarget(Target::TARGET_CLASS);
+
+        foreach (self::$metadataParser->parse($docComment, 'class @' . $name) as $annotation) {
+            if ($annotation instanceof Target) {
+                $annotationBuilder->withTarget(AnnotationTarget::fromAnnotation($annotation));
+
+                continue;
             }
 
-            // if not has a constructor will inject values into public properties
-            if (false === $metadata['has_constructor']) {
-                // collect all public properties
-                foreach ($class->getProperties(\ReflectionProperty::IS_PUBLIC) as $property) {
-                    $metadata['properties'][$property->name] = $property->name;
+            if ($annotation instanceof Attributes) {
+                foreach ($annotation->value as $attribute) {
+                    $propertyBuilder = new PropertyMetadataBuilder($attribute->name);
 
-                    if (false === ($propertyComment = $property->getDocComment())) {
-                        continue;
-                    }
-
-                    $attribute = new Attribute();
-
-                    $attribute->required = (false !== strpos($propertyComment, '@Required'));
-                    $attribute->name     = $property->name;
-                    $attribute->type     = (false !== strpos($propertyComment, '@var') && preg_match('/@var\s+([^\s]+)/',$propertyComment, $matches))
-                        ? $matches[1]
-                        : 'mixed';
-
-                    $this->collectAttributeTypeMetadata($metadata, $attribute);
-
-                    // checks if the property has @Enum
-                    if (false !== strpos($propertyComment, '@Enum')) {
-                        $context = 'property ' . $class->name . "::\$" . $property->name;
-
-                        self::$metadataParser->setTarget(Target::TARGET_PROPERTY);
-
-                        foreach (self::$metadataParser->parse($propertyComment, $context) as $annotation) {
-                            if ( ! $annotation instanceof Enum) {
-                                continue;
-                            }
-
-                            $metadata['enum'][$property->name]['value']   = $annotation->value;
-                            $metadata['enum'][$property->name]['literal'] = ( ! empty($annotation->literal))
-                                ? $annotation->literal
-                                : $annotation->value;
-                        }
-                    }
+                    $this->collectAttributeTypeMetadata($propertyBuilder, $attribute);
+                    $annotationBuilder->withProperty($propertyBuilder->build());
                 }
-
-                // choose the first property as default property
-                $metadata['default_property'] = reset($metadata['properties']);
             }
         }
 
-        self::$annotationMetadata[$name] = $metadata;
+        if ($useConstructor) {
+            $this->metadata->add($annotationBuilder->build());
+
+            return;
+        }
+
+        // if there is no constructor we will inject values into public properties
+
+        // collect all public properties
+        foreach ($class->getProperties(\ReflectionProperty::IS_PUBLIC) as $i => $property) {
+            $propertyBuilder = new PropertyMetadataBuilder($property->getName());
+            $propertyComment = $property->getDocComment();
+
+            if ($i === 0) {
+                $propertyBuilder->withBeingDefault();
+            }
+
+
+            if ($propertyComment === false) {
+                $annotationBuilder->withProperty($propertyBuilder->build());
+
+                continue;
+            }
+
+            $attribute           = new Attribute();
+            $attribute->required = (false !== strpos($propertyComment, '@Required'));
+            $attribute->name     = $property->name;
+            $attribute->type     = (false !== strpos($propertyComment, '@var') && preg_match('/@var\s+([^\s]+)/',$propertyComment, $matches))
+                ? $matches[1]
+                : 'mixed';
+
+            $this->collectAttributeTypeMetadata($propertyBuilder, $attribute);
+
+            // checks if the property has @Enum
+            if (false !== strpos($propertyComment, '@Enum')) {
+                $context = 'property ' . $class->name . "::\$" . $property->name;
+
+                self::$metadataParser->setTarget(Target::TARGET_PROPERTY);
+
+                foreach (self::$metadataParser->parse($propertyComment, $context) as $annotation) {
+                    if ( ! $annotation instanceof Enum) {
+                        continue;
+                    }
+
+                    $propertyBuilder = $propertyBuilder->withEnum([
+                        'value'   => $annotation->value,
+                        'literal' => ( ! empty($annotation->literal))
+                            ? $annotation->literal
+                            : $annotation->value,
+                    ]);
+                }
+            }
+
+            $annotationBuilder->withProperty($propertyBuilder->build());
+        }
+
+        $this->metadata->add($annotationBuilder->build());
     }
 
     /**
@@ -554,11 +496,11 @@ final class DocParser
      *
      * @param array     $metadata
      * @param Attribute $attribute
-     *
-     * @return void
      */
-    private function collectAttributeTypeMetadata(&$metadata, Attribute $attribute)
-    {
+    private function collectAttributeTypeMetadata(
+        PropertyMetadataBuilder $metadata,
+        Attribute $attribute
+    ) : void {
         // handle internal type declaration
         $type = self::$typeMap[$attribute->type] ?? $attribute->type;
 
@@ -567,36 +509,40 @@ final class DocParser
             return;
         }
 
-        // Evaluate type
-        switch (true) {
-            // Checks if the property has array<type>
-            case (false !== $pos = strpos($type, '<')):
-                $arrayType  = substr($type, $pos + 1, -1);
-                $type       = 'array';
-
-                if (isset(self::$typeMap[$arrayType])) {
-                    $arrayType = self::$typeMap[$arrayType];
-                }
-
-                $metadata['attribute_types'][$attribute->name]['array_type'] = $arrayType;
-                break;
-
-            // Checks if the property has type[]
-            case (false !== $pos = strrpos($type, '[')):
-                $arrayType  = substr($type, 0, $pos);
-                $type       = 'array';
-
-                if (isset(self::$typeMap[$arrayType])) {
-                    $arrayType = self::$typeMap[$arrayType];
-                }
-
-                $metadata['attribute_types'][$attribute->name]['array_type'] = $arrayType;
-                break;
+        if ($attribute->required) {
+            $metadata->withBeingRequired();
         }
 
-        $metadata['attribute_types'][$attribute->name]['type']     = $type;
-        $metadata['attribute_types'][$attribute->name]['value']    = $attribute->type;
-        $metadata['attribute_types'][$attribute->name]['required'] = $attribute->required;
+        // Evaluate type
+
+        // Checks if the property has array<type>
+        if (false !== $pos = strpos($type, '<')) {
+            $arrayType = substr($type, $pos + 1, -1);
+
+            $metadata->withType([
+                'type' => 'array',
+                'array_type' => self::$typeMap[$arrayType] ?? $arrayType,
+            ]);
+
+            return;
+        }
+
+        // Checks if the property has type[]
+         if (false !== $pos = strrpos($type, '[')) {
+            $arrayType = substr($type, 0, $pos);
+
+            $metadata->withType([
+                'type'            => 'array',
+                'array_type' => self::$typeMap[$arrayType] ?? $arrayType,
+            ]);
+
+            return;
+        }
+
+        $metadata->withType([
+            'type'  => $type,
+            'value' => $attribute->type,
+        ]);
     }
 
     /**
@@ -710,12 +656,12 @@ final class DocParser
 
 
         // collects the metadata annotation only if there is not yet
-        if ( ! isset(self::$annotationMetadata[$name])) {
+        if (! $this->metadata->has($name) && ! array_key_exists($name, $this->nonAnnotationClasses)) {
             $this->collectAnnotationMetadata($name);
         }
 
         // verify that the class is really meant to be an annotation and not just any ordinary class
-        if (self::$annotationMetadata[$name]['is_annotation'] === false) {
+        if (array_key_exists($name, $this->nonAnnotationClasses)) {
             if ($this->ignoreNotImportedAnnotations || isset($this->ignoredAnnotationNames[$originalName])) {
                 return false;
             }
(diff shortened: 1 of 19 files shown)


lib/Doctrine/Annotations/Metadata/PropertyMetadata.php
```diff
@@ -0,0 +1,75 @@
...
+    /** @var mixed[]|null */
+    private $enum;
+
+    /**
+     * @param array<string, string> $type
+     * @param mixed[]|null          $enum
+     */
+    public function __construct(
+        string $name,
+        ?array $type,
```
Majkl578: Later on `?array` will be replaced by proper `Type`.

lib/Doctrine/Annotations/Metadata/PropertyMetadata.php
```diff
@@ -0,0 +1,75 @@
...
+        bool $required = false,
+        bool $default = false,
+        ?array $enum = null
```
Majkl578: Later on enum info will be merged into Type.

lib/Doctrine/Annotations/Metadata/AnnotationTarget.php
```diff
@@ -0,0 +1,116 @@
...
+use function array_filter;
+use function assert;
+use function implode;
+
+/**
+ * @internal
+ */
+final class AnnotationTarget
+{
+    public const TARGET_CLASS      = 1;
```
Majkl578: If we can come up with other name than _class_, we could drop `TARGET_` prefix.

greg0ire: `TYPE`? `KLASS`?

Majkl578: Type may work, please no non-english words. 😂

greg0ire: Or just `_CLASS` if starting a constant name with an underscore is valid (on my phone rn, hard to check)

Majkl578: I guess we can also use something like `Target::OnClass`/`Target::ON_CLASS`, `Target::InAnnotation`/`Target::IN_ANNOTATION` etc.
Let's leave it as is for now.

lib/Doctrine/Annotations/DocParser.php
```diff
@@ -2,7 +2,14 @@
 
 namespace Doctrine\Annotations;
 
+use function array_key_exists;
```
eugenekurasov: Interesting, but why?

Majkl578: What exactly why? Why `use function` or why `array_key_exists()`? This function will be optimized into specialized opcode in PHP 7.4, but we use `use function` for all functions either way.

eugenekurasov: Maybe `preg_match`, `strpos` we should import too?

And one more question: is the difference between `use function array_key_exists` and `\array_key_exists`

Majkl578: No, those functions are not specialized, only some specific functions are (basically [this](https://github.com/slevomat/coding-standard/blob/5.0.0/SlevomatCodingStandard/Sniffs/PHP/OptimizedFunctionsWithoutUnpackingSniff.php#L28-L61) list).

No, there is no difference, use whichever variant you prefer (although `use function` helps in terms of readability a lot).

Majkl578: > No, those functions are not specialized,

Oh, sorry, I was a bit unclear: They don't need to be imported in order to be optimized by PHP because no such optimization exists. They should however be imported according to Doctrine Coding Standard. But since it's not in effect in this repo at this moment, and that this file is going to be dropped soon anyway, it's not imported.

Ocramius (commented):
Overall looking good, what's the likeliness of this introducing a BC break, and should it target 1.x?

Ocramius:
Also: needs a performance check now that #248 is merged

Majkl578:
This branch (already rebased):
```
\Doctrine\Performance\Annotations\DocLexerPerformanceBench

    benchMethod.............................I4 [μ Mo]/r: 7.067 7.086 (μs) [μSD μRSD]/r: 0.065μs 0.92%
    benchClass..............................I4 [μ Mo]/r: 10.108 10.017 (μs) [μSD μRSD]/r: 0.189μs 1.87%

\Doctrine\Performance\Annotations\DocParsePerformanceBench

    benchMethodParsing......................I4 [μ Mo]/r: 36.381 34.131 (μs) [μSD μRSD]/r: 4.568μs 12.56%
    benchClassParsing.......................I4 [μ Mo]/r: 32.458 31.741 (μs) [μSD μRSD]/r: 1.462μs 4.50%

\Doctrine\Performance\Annotations\PhpParserPerformanceWithoutShortCutBench

    bench...................................I4 [μ Mo]/r: 82.972 82.738 (μs) [μSD μRSD]/r: 0.507μs 0.61%

\Doctrine\Performance\Annotations\PhpParserPerformanceWithShortCutBench

    bench...................................I4 [μ Mo]/r: 88.515 88.243 (μs) [μSD μRSD]/r: 1.227μs 1.39%

\Doctrine\Performance\Annotations\ReadPerformanceBench

    bench...................................I4 [μ Mo]/r: 32.372 32.235 (μs) [μSD μRSD]/r: 0.314μs 0.97%

7 subjects, 35 iterations, 2,900 revs, 0 rejects, 0 failures, 0 warnings
(best [mean mode] worst) = 6.988 [41.410 40.884] 7.162 (μs)
⅀T: 1,449.365μs μSD/r 1.190μs μRSD/r: 3.260%
```

Master:
```
\Doctrine\Performance\Annotations\DocLexerPerformanceBench

    benchMethod.............................I4 [μ Mo]/r: 7.103 7.055 (μs) [μSD μRSD]/r: 0.093μs 1.31%
    benchClass..............................I4 [μ Mo]/r: 10.043 10.020 (μs) [μSD μRSD]/r: 0.044μs 0.44%

\Doctrine\Performance\Annotations\DocParsePerformanceBench

    benchMethodParsing......................I4 [μ Mo]/r: 23.252 23.183 (μs) [μSD μRSD]/r: 0.367μs 1.58%
    benchClassParsing.......................I4 [μ Mo]/r: 23.165 23.099 (μs) [μSD μRSD]/r: 0.148μs 0.64%

\Doctrine\Performance\Annotations\PhpParserPerformanceWithoutShortCutBench

    bench...................................I4 [μ Mo]/r: 83.942 83.336 (μs) [μSD μRSD]/r: 1.254μs 1.49%

\Doctrine\Performance\Annotations\PhpParserPerformanceWithShortCutBench

    bench...................................I4 [μ Mo]/r: 89.248 89.752 (μs) [μSD μRSD]/r: 1.226μs 1.37%

\Doctrine\Performance\Annotations\ReadPerformanceBench

    bench...................................I4 [μ Mo]/r: 23.663 23.359 (μs) [μSD μRSD]/r: 0.622μs 2.63%

7 subjects, 35 iterations, 2,900 revs, 0 rejects, 0 failures, 0 warnings
(best [mean mode] worst) = 7.030 [37.202 37.115] 7.276 (μs)
⅀T: 1,302.083μs μSD/r 0.536μs μRSD/r: 1.352%
```

Majkl578:
> Overall looking good, what's the likeliness of this introducing a BC break, and should it target 1.x?

I wouldn't push it to 1.x at this state. It doesn't change public API (I reverted removal of `Target::$literal`) however it does remove the _internal_ `static` annotations metadata registry - this makes i.e. ORM test suite slower (4.8s -> 5.3s). We could keep it `static` in 1.x though.

Ocramius:
The bench for `benchMethodParsing` is quite worrisome though: 36% slower there.

Is there any hot path that we can inspect, or adding utility API on the builder to reduce hops?

lib/Doctrine/Annotations/DocParser.php
```diff
@@ -468,135 +395,144 @@ private function collectAnnotationMetadata($name)
...
+                'type' => 'array',
+                'array_type' => self::$typeMap[$arrayType] ?? $arrayType,
+            ]);
+        }
 
-                if (isset(self::$typeMap[$arrayType])) {
-                    $arrayType = self::$typeMap[$arrayType];
-                }
+        // Checks if the property has type[]
+         if (false !== $pos = strrpos($type, '[')) {
```
alcaeus: Indentation looks off here.

lib/Doctrine/Annotations/Metadata/AnnotationMetadata.php
```diff
@@ -0,0 +1,97 @@
...
+    private $usesConstructor;
+
+    /** @var PropertyMetadata[] */
+    private $properties;
+
+    /** @var PropertyMetadata|null */
+    private $defaultProperty;
+
+    /**
+     * @param PropertyMetadata[] $properties
```
alcaeus: This is not necessary as the variadic argument already contains a proper type hint.

lib/Doctrine/Annotations/Metadata/AnnotationMetadata.php
```diff
@@ -0,0 +1,97 @@
...
+     */
+    public function __construct(
+        string $name,
+        AnnotationTarget $target,
+        bool $hasConstructor,
+        PropertyMetadata ...$properties
+    ) {
+        $this->name           = $name;
+        $this->target         = $target;
```
alcaeus: Alignment is off.

lib/Doctrine/Annotations/Metadata/AnnotationMetadata.php
```diff
@@ -0,0 +1,97 @@
...
+        $defaultProperties = array_values(
+            array_filter(
+                $properties,
+                static function (PropertyMetadata $property) : bool {
+                    return $property->isDefault();
+                }
+            )
+        );
+
+        assert(count($defaultProperties) <= 1);
```
alcaeus: I wouldn't use `assert` here. To quote from the manual:

> Assertions should not be used for normal runtime operations like input parameter checks. As a rule of thumb your code should always be able to work correctly if assertion checking is not activated.

I agree with that statement and would rather throw a subclass of `InvalidArgumentException` instead of using `assert`: while the code does not fail now if there is more than one default property, this can introduce issues that are difficult to debug at a later stage.

lib/Doctrine/Annotations/Metadata/AnnotationTarget.php
```diff
@@ -0,0 +1,116 @@
...
+        self::TARGET_ANNOTATION => 'ANNOTATION',
+        self::TARGET_ALL        => 'ALL',
+    ];
+
+    /** @var int */
+    private $target;
+
+    public function __construct(int $target)
+    {
+        assert($target >= 0 && $target <= self::TARGET_ALL, 'Target out of bounds.');
```
alcaeus: Same as above, don't use an assertion but rather throw an exception.

lib/Doctrine/Annotations/Metadata/Builder/AnnotationMetadataBuilder.php
```diff
@@ -0,0 +1,65 @@
...
+
+    public function __construct(string $name)
+    {
+        $this->name   = $name;
+        $this->target = AnnotationTarget::all();
+    }
+
+    public function withTarget(AnnotationTarget $target) : self
+    {
+        $new         = clone $this;
```
alcaeus: I get why you would want an immutable object for metadata, but looking at how the builder has to be used I find it cumbersome in this case. The DocParser is filled with repeated assignments to the same variable for something that is thrown away after calling `build()` on it. My suggestion would be to make the builder mutable but reset its state after calling `build()` to avoid issues.

lib/Doctrine/Annotations/Metadata/InternalAnnotations.php
```diff
@@ -0,0 +1,100 @@
...
+            ),
+            new AnnotationMetadata(
+                Attributes::class,
+                AnnotationTarget::class(),
+                false,
+                new PropertyMetadata(
+                    'value',
+                    [
+                        'type'       => 'array',
+                        'array_type' =>'Doctrine\Annotations\Annotation\Attribute',
```
alcaeus: `::class`

lib/Doctrine/Annotations/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,103 @@
...
+
+    public function __construct(AnnotationMetadata ...$metadatas)
+    {
+        $this->add(...$metadatas);
+    }
+
+    public function add(AnnotationMetadata ...$metadatas) : void
+    {
+        foreach ($metadatas as $metadata) {
+            assert(! isset($this[$metadata->getName()]), sprintf('Metadata with name %s already exists.', $metadata->getName()));
```
alcaeus: One more assertion that should be an exception: you really don't want this check to be ignored.

lib/Doctrine/Annotations/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * @param string $name
+     *
+     * @phpcsSuppress SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint
+     */
+    public function offsetGet($name) : AnnotationMetadata
+    {
+        assert(isset($this[$name]), sprintf('Metadata for name %s does not exist', $name));
```
alcaeus: => exception

lib/Doctrine/Annotations/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * @param null               $name
+     * @param AnnotationMetadata $metadata
+     *
+     * @phpcsSuppress SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint
+     */
+    public function offsetSet($name, $metadata) : void
+    {
+        assert($name === null, 'Setting named metadata is not supported.');
```
alcaeus: => exception

lib/Doctrine/Annotations/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * @param string $name
+     *
+     * @phpcsSuppress SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint
+     */
+    public function offsetUnset($name) : void
+    {
+        assert(isset($this[$name]));
```
alcaeus: The `unset` call below doesn't need checking whether the name is set.

lib/Doctrine/Annotations/Metadata/MetadataCollection.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        return count($this->metadata);
+    }
+
+    /**
+     * @return Traversable<AnnotationMetadata>
+     */
+    public function getIterator() : Traversable
+    {
+        yield from array_values($this->metadata);
```
alcaeus: Out of curiosity, why don't we return an iterator with keys?

alcaeus (commented):
Initial feedback below. In general I'd appreciate it if we can enable phpcs in travis-ci build pipeline before moving along with this. I've created #252 to track this.

In general, there are two main things I would like to see changed:
* `assert` is used throughout for sanity checks. Since assertions can (and many times, are) turned off, I don't think this is feasible for checking input parameters. I believe it makes more sense to throw exceptions here.
* The metadata builders are immutable. While I believe that metadata itself needs to be immutable, using the builder in `DocParser` feels cumbersome and would be drastically improved if they weren't immutable. See additional comment below.

alcaeus:
As per https://github.com/doctrine/annotations/pull/253#issuecomment-469672292, we've decided to not apply doctrine/coding-standard for now. @Majkl578 I've marked style comments in my review above as resolved, feel free to ignore that feedback unless it bothers you personally 😉

lib/Doctrine/Annotations/Metadata/MetadataCollection.php
Majkl578: Dropped unsetting.

lib/Doctrine/Annotations/Metadata/MetadataCollection.php
Majkl578: I don't have strong opinion but it feels redundant to getName() and would encourage using `$name => $metadata` pattern.

lib/Doctrine/Annotations/Metadata/Builder/AnnotationMetadataBuilder.php
Majkl578: I agree mutable builder makes stuff simpler and more readable so I removed the immutability. Although it wasn't the source of performance degradation. :)

Majkl578:
Addressed review comments from @alcaeus. Also did two more changes:
- removed ArrayAccess, it was superfluous and didn't fit very well into the API;
- extracted MetadataCollection into interface, it could serve as one of few extension points for consumers (for possible caching).

Metadata:
```
\Doctrine\Performance\Annotations\DocLexerPerformanceBench

    benchMethod.............................I4 [μ Mo]/r: 7.058 7.076 (μs) [μSD μRSD]/r: 0.044μs 0.62%
    benchClass..............................I4 [μ Mo]/r: 10.018 10.060 (μs) [μSD μRSD]/r: 0.059μs 0.59%

\Doctrine\Performance\Annotations\DocParsePerformanceBench

    benchMethodParsing......................I4 [μ Mo]/r: 29.042 29.096 (μs) [μSD μRSD]/r: 0.119μs 0.41%
    benchClassParsing.......................I4 [μ Mo]/r: 28.667 28.728 (μs) [μSD μRSD]/r: 0.131μs 0.46%

\Doctrine\Performance\Annotations\PhpParserPerformanceWithoutShortCutBench

    bench...................................I4 [μ Mo]/r: 83.974 84.338 (μs) [μSD μRSD]/r: 0.527μs 0.63%

\Doctrine\Performance\Annotations\PhpParserPerformanceWithShortCutBench

    bench...................................I4 [μ Mo]/r: 88.739 88.268 (μs) [μSD μRSD]/r: 0.828μs 0.93%

\Doctrine\Performance\Annotations\ReadPerformanceBench

    bench...................................I4 [μ Mo]/r: 27.972 27.578 (μs) [μSD μRSD]/r: 0.822μs 2.94%

7 subjects, 35 iterations, 2,900 revs, 0 rejects, 0 failures, 0 warnings
(best [mean mode] worst) = 6.988 [39.353 39.306] 7.108 (μs)
⅀T: 1,377.349μs μSD/r 0.361μs μRSD/r: 0.940%
```

Master:
```
\Doctrine\Performance\Annotations\DocLexerPerformanceBench

    benchMethod.............................I4 [μ Mo]/r: 7.086 7.047 (μs) [μSD μRSD]/r: 0.063μs 0.89%
    benchClass..............................I4 [μ Mo]/r: 10.058 10.006 (μs) [μSD μRSD]/r: 0.106μs 1.05%

\Doctrine\Performance\Annotations\DocParsePerformanceBench

    benchMethodParsing......................I4 [μ Mo]/r: 22.700 22.633 (μs) [μSD μRSD]/r: 0.211μs 0.93%
    benchClassParsing.......................I4 [μ Mo]/r: 23.049 22.871 (μs) [μSD μRSD]/r: 0.368μs 1.59%

\Doctrine\Performance\Annotations\PhpParserPerformanceWithoutShortCutBench

    bench...................................I4 [μ Mo]/r: 84.378 84.647 (μs) [μSD μRSD]/r: 0.559μs 0.66%

\Doctrine\Performance\Annotations\PhpParserPerformanceWithShortCutBench

    bench...................................I4 [μ Mo]/r: 88.188 87.933 (μs) [μSD μRSD]/r: 0.656μs 0.74%

\Doctrine\Performance\Annotations\ReadPerformanceBench

    bench...................................I4 [μ Mo]/r: 23.430 23.196 (μs) [μSD μRSD]/r: 0.472μs 2.02%

7 subjects, 35 iterations, 2,900 revs, 0 rejects, 0 failures, 0 warnings
(best [mean mode] worst) = 7.016 [36.984 36.905] 7.180 (μs)
⅀T: 1,294.449μs μSD/r 0.348μs μRSD/r: 1.126%
```

Majkl578:
100% coverage in Metadata namespace. 😃

lib/Doctrine/Annotations/DocParser.php
```diff
@@ -468,97 +396,106 @@ private function collectAnnotationMetadata($name)
...
-            foreach (self::$metadataParser->parse($docComment, 'class @' . $name) as $annotation) {
-                if ($annotation instanceof Target) {
-                    $metadata['targets']         = $annotation->targets;
-                    $metadata['targets_literal'] = $annotation->literal;
+        $annotationBuilder = new AnnotationMetadataBuilder($name);
 
-                    continue;
-                }
+        if ($useConstructor) {
+            $annotationBuilder->withUsingConstructor();
```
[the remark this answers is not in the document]
Majkl578: In general I prefer to avoid public properties until PHP adds proper accessors or at least readonly properties. For `@internal` stuff it may be sufficient but since we don't have PHPStan at this moment, I'd leave this to a later stage - once we have PHPStan L7, we can do this -> create a blocker issue for 2.0 for this?

lib/Doctrine/Annotations/DocParser.php
```diff
@@ -728,80 +669,97 @@ private function Annotation()
...
+            if (! isset($metadata->getProperties()[$property])) {
                 if ('value' !== $property) {
-                    throw AnnotationException::creationError(sprintf('The annotation @%s declared on %s does not have a property named "%s". Available properties: %s', $originalName, $this->context, $property, implode(', ', self::$annotationMetadata[$name]['properties'])));
+                    throw AnnotationException::creationError(
+                        sprintf(
+                            'The annotation @%s declared on %s does not have a property named "%s". Available properties: %s',
+                            $originalName,
+                            $this->context,
+                            $property,
+                            implode(', ', array_keys($metadata->getProperties()))
```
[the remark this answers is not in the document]
Majkl578: Good catch, looks like there is no test for this case.

lib/Doctrine/Annotations/Metadata/Exception/InvalidAnnotationTarget.php
```diff
@@ -0,0 +1,16 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Annotations\Metadata\Exception;
+
+use InvalidArgumentException;
+use function sprintf;
+
+final class InvalidAnnotationTarget extends InvalidArgumentException
+{
+    public static function new(int $target) : self
```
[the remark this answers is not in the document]
Majkl578: `fromBitmask(int $invalidBitmask)`?

lib/Doctrine/Annotations/DocParser.php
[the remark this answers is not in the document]
Majkl578: Actually, only wrong docBlock, it was already `array<string, PropertyMetadata>`.

lib/Doctrine/Annotations/DocParser.php
```diff
@@ -468,97 +396,106 @@ private function collectAnnotationMetadata($name)
...
             }
+        }
 
-            // if not has a constructor will inject values into public properties
-            if (false === $metadata['has_constructor']) {
-                // collect all public properties
-                foreach ($class->getProperties(\ReflectionProperty::IS_PUBLIC) as $property) {
-                    $metadata['properties'][$property->name] = $property->name;
+        // if there is no constructor we will inject values into public properties
+        if (! $useConstructor) {
```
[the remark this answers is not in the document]
Majkl578: I flipped the condition and used early-return. I don't think this is worth any more patience as we'll drop DocParser anyway.

lib/Doctrine/Annotations/DocParser.php
```diff
@@ -468,97 +396,106 @@ private function collectAnnotationMetadata($name)
...
+                $annotationBuilder->withTarget(AnnotationTarget::fromAnnotation($annotation));
+
+                continue;
+            }
+
+            if ($annotation instanceof Attributes) {
+                foreach ($annotation->value as $attribute) {
+                    $propertyBuilder = new PropertyMetadataBuilder($attribute->name);
+
+                    $this->collectAttributeTypeMetadata($propertyBuilder, $attribute);
```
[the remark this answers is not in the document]
Majkl578: Let's leave it as is, DocParser is to be dropped anyway.

lib/Doctrine/Annotations/Metadata/Exception/InvalidAnnotationTarget.php
```diff
@@ -0,0 +1,16 @@
...
+    {
+        return new self(sprintf('Annotation target %d is not valid bitmask of allowed targets.', $target));
```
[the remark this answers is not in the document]
Majkl578: As this is mostly an internal exception, how would you do this? List constant names? Integer range?

lib/Doctrine/Annotations/Metadata/Exception/InvalidAnnotationTarget.php
```diff
@@ -0,0 +1,16 @@
...
+namespace Doctrine\Annotations\Metadata\Exception;
+
+use InvalidArgumentException;
+use function sprintf;
+
+final class InvalidAnnotationTarget extends InvalidArgumentException
+{
+    public static function fromInvalidBitmask(int $bitmask) : self
+    {
+        return new self(sprintf('Annotation target %d is not valid bitmask of allowed targets.', $bitmask));
```
alcaeus: I would suggest putting the target into double quotes to better separate them from the rest of the exception message:
```suggestion
        return new self(sprintf('Annotation target "%d" is not valid bitmask of allowed targets.', $bitmask));
```

lib/Doctrine/Annotations/Metadata/Exception/MetadataAlreadyExists.php
```diff
@@ -0,0 +1,17 @@
...
+
+use Doctrine\Annotations\Metadata\AnnotationMetadata;
+use InvalidArgumentException;
+use function sprintf;
+
+final class MetadataAlreadyExists extends InvalidArgumentException
+{
+    public static function new(AnnotationMetadata $metadata) : self
+    {
+        return new self(sprintf('Metadata for annotation %s already exists.', $metadata->getName()));
```
alcaeus:
```suggestion
        return new self(sprintf('Metadata for annotation "%s" already exists.', $metadata->getName()));
```

lib/Doctrine/Annotations/Metadata/Exception/MetadataDoesNotExist.php
```diff
@@ -0,0 +1,16 @@
...
+namespace Doctrine\Annotations\Metadata\Exception;
+
+use RuntimeException;
+use function sprintf;
+
+final class MetadataDoesNotExist extends RuntimeException
+{
+    public static function new(string $name) : self
+    {
+        return new self(sprintf('Metadata for annotation %s does not exist.', $name));
```
alcaeus:
```suggestion
        return new self(sprintf('Metadata for annotation "%s" does not exist.', $name));
```

lib/Doctrine/Annotations/Metadata/Exception/TooManyDefaultProperties.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Doctrine\Annotations\Metadata\Exception;
+
+use LogicException;
+
+final class TooManyDefaultProperties extends LogicException
+{
+    public static function new() : self
+    {
+        return new self('The annotation can only have at most one default property.');
```
alcaeus: This exception message should at least contain the name of the annotation.

It would probably also be helpful if it listed the properties that are currently marked as default - this would help users find the cause of the error.

alcaeus (commented):
Few comments below. A couple of general ideas:

* Do you want to allow constructing exceptions without usage of the factories or should the constructor be private?
* Since there are a lot of final classes, this becomes very dependent on our own classes without the option to properly customise anything. I think we should introduce interfaces for the builders (containing only the `build` method as everything else is an implementation detail) as well as an interface for the metadata classes (like you did with `MetadataCollection`). That way we're not closing down the entire API to the point where people will have to fork it if they want to change a single thing.

Majkl578:
> Do you want to allow constructing exceptions without usage of the factories or should the constructor be private?

Right now we allow it in Migrations and ORM master. Probably not a big deal, note that proper exception structure is yet to be introduced (converting AnnotationException to interface etc.).

> Since there are a lot of final classes, this becomes very dependent on our own classes without the option to properly customise anything.

This is intentional. Most of the parts of Annotations should not be extended and those that could should be carefully considered.
I'd like to hear opinion from @Ocramius here, in my opinion the more we keep closed to midifications, the better, especially for such a widely used library.
Of course we could reconsider this later, before 2.0 or even after - unfinalizing a class should be BC-okay.

lib/Doctrine/Annotations/Metadata/Exception/InvalidAnnotationTarget.php
```diff
@@ -0,0 +1,16 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Annotations\Metadata\Exception;
+
+use InvalidArgumentException;
+use function sprintf;
+
+final class InvalidAnnotationTarget extends InvalidArgumentException
+{
+    public static function fromInvalidBitmask(int $bitmask) : self
```
alcaeus: Tiny nitpick: this is the only factory method not named `new`. However, we can merge this as-is since I assume that we'll have to do a larger refactoring of exceptions to get them all to be consistent with these new exceptions.

alcaeus (dismissed):

lib/Doctrine/Annotations/Metadata/AnnotationMetadata.php
```diff
@@ -0,0 +1,97 @@
...
+
+use Doctrine\Annotations\Metadata\Exception\TooManyDefaultProperties;
+use function array_combine;
+use function array_filter;
+use function array_map;
+use function array_values;
+use function count;
+
+/**
+ * @internal
```
Ocramius: Let's add a line to explain:

 * why this is internal
 * what this metadata represents (it's unclear to the reader, but this is about an annotation's definition, rather than an annotation instance)

lib/Doctrine/Annotations/Metadata/Builder/PropertyMetadataBuilder.php
```diff
@@ -0,0 +1,78 @@
...
+    /** @var string[]|null */
+    private $type;
+
+    /** @var bool */
+    private $required = false;
+
+    /** @var bool */
+    private $default = false;
+
+    /** @var mixed[]|null */
```
Ocramius: Is it possible to have a stricter type declaration here? Specifically, do we allow objects and resources? I think we only allow constant-compatible values here.

lib/Doctrine/Annotations/Metadata/PropertyMetadata.php
```diff
@@ -0,0 +1,75 @@
...
+    /** @var array<string, string>|null */
+    private $type;
+
+    /** @var bool */
+    private $required;
+
+    /** @var bool */
+    private $default;
+
+    /** @var mixed[]|null */
```
Ocramius: Is it possible to have a stricter type declaration here? Specifically, do we allow objects and resources? I think we only allow constant-compatible values here.

lib/Doctrine/Annotations/Metadata/PropertyMetadata.php
```diff
@@ -0,0 +1,75 @@
...
+    private $enum;
+
+    /**
+     * @param array<string, string> $type
+     * @param mixed[]|null          $enum
```
Ocramius: Is it possible to have a stricter type declaration here? Specifically, do we allow objects and resources? I think we only allow constant-compatible values here.

Ocramius (changes requested):

lib/Doctrine/Annotations/Metadata/AnnotationMetadata.php
Majkl578: I decided to remove `@internal` from AnnotationMetadata, PropertyMetadata and AnnotationTarget because since MetadataCollection is an extension point, it should not depend on internal elements.

lib/Doctrine/Annotations/Metadata/Exception/InvalidAnnotationTarget.php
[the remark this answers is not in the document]
Ocramius: Let's skip over this - indeed not very important, and can be improved later on if someone thinks this is frustrating.

Ocramius (approved):

PR #248, opened 2019-02-25
Convert performance tests to PHPBench

Majkl578:
Instead of testing perf with PHPUnit and printing some fancy output, we can use PHPBench instead.

It's not possible to have phpbench/phpbench as dev-dependency due to back-reference to doctrine/annotations 1.x.

I don't have much experience with writing PHPBench suites yet so suggestions are welcomed. :)


.travis.yml
```diff
@@ -56,3 +56,10 @@ jobs:
     - stage: Lint
       install: travis_retry composer require --dev --prefer-dist phpstan/phpstan:^0.7
       script: vendor/bin/phpstan analyse -l 3 -c phpstan.neon lib tests
+
+    - stage: Benchmark
+      install:
+        - travis_retry composer update --prefer-dist
+        - curl -o phpbench.phar https://phpbench.github.io/phpbench/phpbench.phar
```
Ocramius: I know I'm a weirdo, but I'd rather `composer require` with a specific version here. I don't consider `curl` to be a good way to install things...

composer.json
```diff
@@ -34,7 +34,13 @@
         "psr-4": { "Doctrine\\Annotations\\": "lib/Doctrine/Annotations" }
     },
     "autoload-dev": {
-        "psr-4": { "Doctrine\\Tests\\Annotations\\": "tests/Doctrine/Tests/Annotations" }
+        "psr-4": {
+            "Doctrine\\Performance\\Annotations\\": "tests/Doctrine/Performance/Annotations",
+            "Doctrine\\Tests\\Annotations\\": "tests/Doctrine/Tests/Annotations"
+        },
+        "files": [
+            "tests/Doctrine/Tests/Annotations/Fixtures/SingleClassLOC1000.php"
```
Ocramius: Does this need to be in classmap?

phpbench.json.dist
```diff
@@ -0,0 +1,4 @@
+{
+    "bootstrap": "vendor/autoload.php",
```
Ocramius: Possibly not needed if we use a `vendor` version of the CLI

tests/Doctrine/Performance/Annotations/CachedReadPerformanceWithInMemoryBench.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /** @var CachedReader */
+    private $reader;
+
+    /** @var ReflectionMethod */
+    private $method;
+
+    public function initialize() : void
+    {
+        $this->reader = new CachedReader(new AnnotationReader(), new ArrayCache());
```
Ocramius: Should probably use a void cache in this sort of test. Also clashes with `CachedReader` removal - maybe take the chance and get rid of it overall?

Ocramius (changes requested):

.travis.yml
Majkl578: I'd too, but it's not possible, as said above:

> It's not possible to have phpbench/phpbench as dev-dependency due to back-reference to doctrine/annotations 1.x.

Also `curl`ing specific version is currently not possible as versioned PHARs are not published (already discussed with @dantleech to be possible in future).

composer.json
Majkl578: It's an unnamespaced class that was loaded using `require_once` previously.

tests/Doctrine/Performance/Annotations/CachedReadPerformanceWithInMemoryBench.php
Majkl578: #249 😛

.travis.yml
SenseException: Otherwise a phar could be downloaded with phive, but I had a few issues with travis in the past.

Ocramius: Still same issues: would need to be in the dependency graph. Let's tackle the problem once we hit 2.0.0 (stable) then, and a dependency loop *can* be re-introduced, if needed.

Ocramius (approved):
:ship:

Note that this should hit a `1.x` release before we push for `2.x`, as otherwise we don't know if we're making things better or worse.

Ocramius:

PR #253, opened 2019-02-27
Upgrade Doctrine CS, use only for new files

Majkl578:
6.0.x-dev because I am going to use the _new_ syntax for generics and intersections which only works in 6.0-dev.
Not applying to any code from 1.x era to avoid headaches during rebase conflicts. We will do it later for files that survive refactoring.

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -1,0 +1,56 @@
+<?xml version="1.0"?>
+<ruleset>
+    <arg name="basepath" value="."/>
+    <arg name="extensions" value="php"/>
+    <arg name="parallel" value="80"/>
+    <arg name="cache" value=".phpcs-cache"/>
+    <arg name="colors"/>
+
+     <!-- Ignore warnings, show progress of the run and show sniff names -->
+    <arg value="nps"/>
+
+    <rule ref="Doctrine"/>
+
+    <file>lib/</file>
+    <file>tests/</file>
+
+    <exclude-pattern>lib/Doctrine/Annotations/Annotation/*</exclude-pattern>
+    <exclude-pattern>lib/Doctrine/Annotations/AnnotationException.php</exclude-pattern>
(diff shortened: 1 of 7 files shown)


phpcs.xml.dist
```diff
@@ -0,0 +1,56 @@
...
+
+     <!-- Ignore warnings, show progress of the run and show sniff names -->
+    <arg value="nps"/>
+
+    <rule ref="Doctrine"/>
+
+    <file>lib/</file>
+    <file>tests/</file>
+
+    <exclude-pattern>lib/Doctrine/Annotations/Annotation/*</exclude-pattern>
```
Ocramius: Lots and lots of excludes: is this worth doing in first place, if half the lib isn't covered at all?

Ocramius (changes requested):

phpcs.xml.dist
Majkl578: It's for the new files we add, e.g. #247, new parser etc.

alcaeus:
Agree with @Ocramius here: if we're only going to exclude most files, let's do this when we don't need to exclude everything. Just to understand the reasoning, why not wait until #247 is merged and then do this for the entire codebase?

Majkl578:
@alcaeus Already described the situation in OP:
> Not applying to any code from 1.x era to avoid headaches during rebase conflicts. We will do it later for files that survive refactoring.

alcaeus:
After discussion in Slack, we've decided to hold back introducing doctrine/coding-standard until the initial refactoring is done: the current coding standard should not be applied to code that will not survive refactoring, but at the same time we also don't want to apply the coding standard to a small subset of files.

Instead, maintainers and containers are encouraged to make sure their code is formatted in compliance with doctrine/coding-standard. However, code reviews should refrain from enforcing doctrine/coding-standard until it is officially adopted in the repository and checked automatically.

PR #254, opened 2019-03-06
Add all PHPUnit annotations to the globally ignored list

jrjohnson:
Everything from https://phpunit.readthedocs.io/en/8.0/annotations.html


jrjohnson:
I'm happy to add the rest, I didn't want to go overboard. Just let me know what the right approach is.

alcaeus:
The reason why most of them aren’t ignored is because most of them are only relevant in test files, where the annotation parser doesn’t go very often. The code coverage annotations on the other hand are used in entities and other logic classes, so they have to be excluded.

@doctrine/doctrinecore what’s our stance here? Should we ignore all PHPUnit annotations by default or should we expect users to manually exclude them if they use the annotation parser on test classes?

ostrolucky:
It would help if @jrjohnson would tell us his use case for using annotation parser in test files

Majkl578:
> Should we ignore all PHPUnit annotations by default or should we expect users to manually exclude them if they use the annotation parser on test classes?

I think it might fine to include them in Annotations 2.0 where unqualified annotations will no longer be supported, thus also minimized chances for conflicts.

jrjohnson:
A test library I'm using has recently included `@DisableDatabaseCache` as an [annotation in tests](https://github.com/liip/LiipFunctionalTestBundle/pull/484) which does pretty much what it looks like. It's the first time I've seen a parseable annotation in a test myself, but I don't see anything wrong with it as a concept. 

Sounds like I should go ahead an add the rest of the PHPUnit annotations, I'll update this PR with those included.

jrjohnson:
I've added all the annotations found at https://phpunit.readthedocs.io/en/8.0/annotations.html

lib/Doctrine/Annotations/ImplicitlyIgnoredAnnotationNames.php
```diff
@@ -83,10 +83,39 @@ final class ImplicitlyIgnoredAnnotationNames
     ];
 
     private const PHPUnit = [
-        'codeCoverageIgnore'      => true,
-        'codeCoverageIgnoreEnd'   => true,
-        'codeCoverageIgnoreStart' => true,
-        'covers'                  => true,
+        'author'                         => true,
```
Majkl578: `@author` is already part of phpDoc itself (see above).

Majkl578: Technically it's fine to have duplicates so we can even keep it for completeness.

jrjohnson:
No problem @greg0ire commits are no squashed.

Majkl578:
Thank you @jrjohnson.

jrjohnson:
🎊 Happy to help!

PR #256, opened 2019-03-14
Type hiearchy for property types and enums

Majkl578:
This PR introduces a type hiearchy for annotation properties - types and enums.

Summary:
- introduces rather simple type hiearchy with validation and describe support;
- contains support for constant scalar types;
- `describe()` uses phpDoc-friendly format (plus intersections and constant values);
- validation methods return booleans instead of throwing;
- property metadata now use type objects instead of arrays for type and enum information;
- moves validation decisions out of DocParser into type objects themselves;
- moves type info formatting out of DocParser into type objects themselves.

I didn't put much effort into DocParser changes as it's eventually going to disappear.

Future scope:
- merge type and enum into one,
- rework parser for `@var` to properly support unions and unqualified names.

Coverage in Type namespace is 100%. 😃


lib/Doctrine/Annotations/DocParser.php
```diff
@@ -710,28 +753,35 @@ private function Annotation()
...
 
-                // checks if the attribute has array type declaration, such as "array<string>"
-                if (isset($type['array_type'])) {
-                    foreach ($values[$valueName] as $item) {
-                        if (gettype($item) !== $type['array_type'] && !$item instanceof $type['array_type']) {
-                            throw AnnotationException::attributeTypeError($propertyName, $originalName, $this->context, 'either a(n) '.$type['array_type'].', or an array of '.$type['array_type'].'s', $item);
+                $valueType = $type->getValueType();
+
+                if (! $type->validate($values[$valueName])) {
+                    $firstInvalidValue = (static function (array $values) use ($valueType) {
```
Majkl578: This is awful tbh.

jwage (dismissed):

lib/Doctrine/Annotations/Type/Constant/ConstantFloatType.php
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public function getValue() : float
+    {
+        return $this->value;
+    }
+
+    public function describe() : string
+    {
+        return sprintf('%F', $this->value);
```
ostrolucky: I would sugget to use just `(string) $this-value`
```php
>>> sprintf('%F', 3.3)
=> "3.300000"
>>> (string) 3.3
=> "3.3"
```

lib/Doctrine/Annotations/Type/Constant/ConstantIntegerType.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public function getValue() : int
+    {
+        return $this->value;
+    }
+
+    public function describe() : string
+    {
+        return sprintf('%d', $this->value);
```
ostrolucky: Same here. I don't see value of sprintf in these

lib/Doctrine/Annotations/Type/Constant/ConstantStringType.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public function getValue() : string
+    {
+        return $this->value;
+    }
+
+    public function describe() : string
+    {
+        return sprintf('"%s"', addcslashes($this->value, '\"'));
```
ostrolucky: Why not `addslashes`, but `addcslashes` instead?

lib/Doctrine/Annotations/Type/ObjectType.php
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @internal
+ */
+class ObjectType implements Type
+{
+    /** @var string|null */
+    private $name;
+
+    public function __construct(?string $name)
```
ostrolucky: $name -> $className

lib/Doctrine/Annotations/Type/ScalarType.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Annotations\Type;
+
+/**
+ * @internal
+ */
+interface ScalarType extends Type
```
ostrolucky: This will be really confusing when coexisting with real ScalarType class

lib/Doctrine/Annotations/Type/CompositeType.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Annotations\Type;
+
+/**
+ * @internal
+ */
+interface CompositeType extends Type
```
ostrolucky: Should be replaced with abstract class so we stop duplicating logic in Union and Intersection types

tests/Doctrine/Tests/Annotations/Type/ArrayTypeTest.php
```diff
@@ -0,0 +1,222 @@
...
+        ];
+        yield [
+            new ArrayType(
+                new UnionType(new IntegerType(), new ObjectType(null)),
+                new UnionType(
+                    new StringType(),
+                    new IntersectionType(new ObjectType(Traversable::class), new ObjectType(Countable::class))
+                )
+            ),
+            'array<int|object, string|(Traversable&Countable)>',
```
ostrolucky: should we really allow to have possible `object` as array key?

tests/Doctrine/Tests/Annotations/Type/ArrayTypeTest.php
```diff
@@ -0,0 +1,222 @@
...
+    {
+        $type = new ArrayType(new IntegerType(), new StringType());
+
+        self::assertInstanceOf(StringType::class, $type->getValueType());
+    }
+
+    /**
+     * @return (ArrayType|string)[]
+     */
+    public function descriptionProvider() : iterable
```
ostrolucky: There should be single data provider which includes all: definitions, descriptions, valid values, invalid values. Splitting it makes it very difficult to keep in sync and make sure all cases have correlated tests

lib/Doctrine/Annotations/Type/Constant/ConstantFloatType.php
```diff
@@ -0,0 +1,45 @@
...
+use function abs;
+use function is_float;
+use function sprintf;
+
+/**
+ * @internal
+ */
+final class ConstantFloatType extends FloatType implements ConstantScalarType
+{
+    public const EPSILON = 1e-15;
```
ostrolucky: This needs description. `Smallest value available in PHP.`

tests/Doctrine/Tests/Annotations/Type/Constant/ConstantFloatTypeTest.php
```diff
@@ -0,0 +1,40 @@
...
+        self::assertEquals(1.1111111111, (new ConstantFloatType(1.1111111111))->getValue(), '', 1e-10);
+    }
+
+    public function testDescribe() : void
+    {
+        self::assertSame('1.230000', (new ConstantFloatType(1.23))->describe());
+        self::assertSame('1.111111', (new ConstantFloatType(1.1111111111))->describe());
+    }
+
+    public function testValidate() : void
```
ostrolucky: There needs to be testcase proving the need for epsilon

ostrolucky (changes requested):
I like this overall, good job! Couple comments that need to be addressed

lib/Doctrine/Annotations/DocParser.php
```diff
@@ -528,18 +553,36 @@ private function collectAttributeTypeMetadata(
...
-        $metadata->withType([
-            'type'  => $type,
-            'value' => $attribute->type,
-        ]);
+        $metadata->withType($this->createTypeFromName($attribute->type));
+    }
+
+    private function createTypeFromName(string $name) : Type
+    {
+        if ($name === 'bool' || $name === 'boolean' || $name === 'Boolean') {
```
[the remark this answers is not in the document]
Majkl578: Not applicable, this code is soon to be removed, but it's covered by test.

lib/Doctrine/Annotations/DocParser.php
Majkl578: It's a temporary hack. Later on I think using exception in validate would be a better approach.

lib/Doctrine/Annotations/DocParser.php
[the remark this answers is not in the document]
ostrolucky: How is it not applicable?

lib/Doctrine/Annotations/Type/CompositeType.php
Majkl578: It's not duplicated logic, intersecions and unions are opposite patterns. Even then I'd prefer trait.

lib/Doctrine/Annotations/Type/Constant/ConstantFloatType.php
Majkl578: No: https://3v4l.org/s1FrW

lib/Doctrine/Annotations/Type/Constant/ConstantFloatType.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        return sprintf('%F', $this->value);
+    }
+
+    /**
+     * @param mixed $value
+     */
+    public function validate($value) : bool
+    {
+        return is_float($value) && abs($value - $this->value) < self::EPSILON;
```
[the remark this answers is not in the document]
Majkl578: It's obvious from the "epsilon" constant: https://en.wikipedia.org/wiki/Machine_epsilon

lib/Doctrine/Annotations/Type/Constant/ConstantStringType.php
Majkl578: `addslashes` doesn't allow specifying characters.

lib/Doctrine/Annotations/Type/MixedType.php
```diff
@@ -0,0 +1,26 @@
...
+    {
+        return 'mixed';
+    }
+
+    /**
+     * @param mixed $value
+     */
+    public function validate($value) : bool
+    {
+        return ! is_resource($value);
```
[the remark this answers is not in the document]
Majkl578: How would you define a resource value in phpDoc? :)

lib/Doctrine/Annotations/Type/ScalarType.php
Majkl578: No such class exists and I don't currently plan adding it (it's a union after all). If it does, we can rename stuff. It's internal after all.

lib/Doctrine/Annotations/DocParser.php
ostrolucky: I don't want temporary workarounds to be merged. Fix problems reviewers point out. Don't do _I'll just do it later..._

tests/Doctrine/Tests/Annotations/Type/ArrayTypeTest.php
Majkl578: As it is now, yes. Whether it works in parser/language is another thing.
PHPStan does that also btw: https://github.com/phpstan/phpstan/blob/master/src/Type/ArrayType.php

tests/Doctrine/Tests/Annotations/Type/Constant/ConstantFloatTypeTest.php
Majkl578: It's already covered by 2nd assert.

lib/Doctrine/Annotations/Type/CompositeType.php
ostrolucky: So what is this if not duplicated logic

Screenshot 2019-03-31 at 20 27 52

lib/Doctrine/Annotations/Type/MixedType.php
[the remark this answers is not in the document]
ostrolucky: Exactly. So this isn't MixedType. This should just return true, since it's not possible to define it in phpdoc anyway.

lib/Doctrine/Annotations/Type/CompositeType.php
kejwmen: But it's still not a good argument for using inheritance

lib/Doctrine/Annotations/DocParser.php
Majkl578: This code will be dropped completely so I don'zt really care about that tbh. :D And as said, it's covered by test so it won't be forgotten.
Also, I don't think keys were ever validated so it's basically nop in DocParser.

lib/Doctrine/Annotations/Type/CompositeType.php
Majkl578: Describe maybe. validate() is not duplicated logic.

tests/Doctrine/Tests/Annotations/Type/ArrayTypeTest.php
ostrolucky: > As it is now, yes.

Why?

lib/Doctrine/Annotations/DocParser.php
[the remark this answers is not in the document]
Majkl578: DocParser will be dropped completely and the new parser allows case insensitive (currently).

lib/Doctrine/Annotations/Type/MixedType.php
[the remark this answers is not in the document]
Majkl578: It is in context of annotations: resource is not a valid type.

lib/Doctrine/Annotations/DocParser.php
ostrolucky: > And as said, it's covered by test 

No you didn't. Which test is testing for keys here?

tests/Doctrine/Tests/Annotations/Type/Constant/ConstantFloatTypeTest.php
ostrolucky: That doesn't cover it
```php
>>> 1.1111111111 === 1.1111111111000000000000000000000000001
=> true
```

lib/Doctrine/Annotations/Type/Constant/ConstantFloatType.php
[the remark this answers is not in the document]
ostrolucky: That link does not explain this code AT ALL

lib/Doctrine/Annotations/Type/Constant/ConstantStringType.php
ostrolucky: Please add comment explaining why are only these characters escaped and we purposely ignore escaping of single quote and NUL bytes.

lib/Doctrine/Annotations/Type/MixedType.php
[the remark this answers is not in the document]
ostrolucky: Don't leak assumptions over whole codebase. This file shouldn't know what is and what isn't possible to define in docblock. It says MixedType but doesn't accept resource. It's a problem. Fix naming or make it accept anything. Shouldn't be a big deal since you can't normally define resource in annotations as you say.

lib/Doctrine/Annotations/DocParser.php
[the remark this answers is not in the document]
ostrolucky: Until then, comment needs to explain why we have uppercased boolean.

lib/Doctrine/Annotations/Type/Constant/ConstantFloatType.php
ostrolucky: better than
```php
>>> sprintf('%F', 3.33333333333333333333333333333333333333334)
=> "3.333333"`
```

lib/Doctrine/Annotations/Type/CompositeType.php
ostrolucky: validate() is duplicated logic with inverted result. Equivalent to `return ! parent::validate($value)`

I don't care about inheritance but duplicated logic. Fix in one class will require touching both files. Use your favorite composition if you think it's more appropriate, whatever. You can also use traits, I don't care.

lib/Doctrine/Annotations/Type/ScalarType.php
ostrolucky: Fair enough

lib/Doctrine/Annotations/Type/MixedType.php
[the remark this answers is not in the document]
Majkl578: It's part of Annotations library and it's not public API, it's fine to know overall context.

lib/Doctrine/Annotations/Type/Constant/ConstantStringType.php
ostrolucky: On second thought, it's obvious enough because it's enclosed with double quotes, so double quotes need to be escaped only

lib/Doctrine/Annotations/Type/MixedType.php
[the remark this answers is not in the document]
ostrolucky: So why is it caring about resource when this is annotations library?

lib/Doctrine/Annotations/DocParser.php
Majkl578: Right, there is no test as the keys were not validated at all, as seen above:
```diff
- foreach ($values[$valueName] as $item) {
```
Currently keys stay of mixed type so key violations can't occur. With new parser it may/will change, right now there is nothing to test.

lib/Doctrine/Annotations/Type/ArrayType.php
```diff
@@ -0,0 +1,70 @@
...
+ */
+class ArrayType implements Type
+{
+    /** @var Type */
+    private $keyType;
+
+    /** @var Type */
+    private $valueType;
+
+    public function __construct(Type $keyType, Type $valueType)
```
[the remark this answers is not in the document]
Majkl578: Let's not complicate it any further now. We'd also need to consider scalar or union of scalars for example, will implement when we actually start honoring key types.

lib/Doctrine/Annotations/Type/ArrayType.php
```diff
@@ -0,0 +1,70 @@
...
+     * @param mixed $value
+     */
+    public function validate($value) : bool
+    {
+        if (! is_array($value)) {
+            return false;
+        }
+
+        foreach ($value as $key => $innerValue) {
+            if (! $this->keyType->validate($key)) {
```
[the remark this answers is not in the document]
Majkl578: Same reason as https://github.com/doctrine/annotations/pull/256/files#r270665169.

lib/Doctrine/Annotations/Type/CompositeType.php
Majkl578: Extracted describe() into static utility function but keeping validate as is.
`return ! parent::validate($value)` is not valid approach and using such would make the code less understendable. Feel free to come up with follow-up PR if you find nice solution. 👍

lib/Doctrine/Annotations/Type/Constant/ConstantFloatType.php
Majkl578: I'd agree if used for validation, but here we're talking about describe() where simpleness is preferable. So keeping as is.

tests/Doctrine/Tests/Annotations/Type/ArrayTypeTest.php
Majkl578: Explained above in a comment for ArrayType.

tests/Doctrine/Tests/Annotations/Type/Constant/ConstantFloatTypeTest.php
Majkl578: 1.1111111111000009 will be better choice.

tests/Doctrine/Tests/Annotations/Type/ArrayTypeTest.php
Majkl578: I don't like the idea of massive multi-purpose data providers, that forces you to focus on things you really are not testing when using such data provider.
Feel free to follow-up with your PR.

lib/Doctrine/Annotations/Type/Type.php
```diff
@@ -0,0 +1,18 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Annotations\Type;
+
+/**
+ * @internal
+ */
+interface Type
+{
+    public function describe() : string;
```
Ocramius: A description of the intent of this method is needed. What is the "description" you will get from this? Human-readable? Machine-friendly? An identifier? The type FQN?

lib/Doctrine/Annotations/Type/ConstantType.php
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * @internal
+ */
+interface ConstantType extends Type
+{
+    /**
+     * @return mixed
+     */
+    public function getValue();
```
Ocramius: This confuses me: a constant "has" a type, but "isn't" a type. There's some confusion about the naming of these interfaces, IMO

Ocramius: Is `ConstantType` a restriction over possible types? If so, I'd say that an `EnumType` should probably be represented too (as a `ConstantType[]` collection)

Ocramius (changes requested):
Overall good improvement, but some confusion between types and values (mostly a question of naming and documentation).

If we think of the definition of "type" as a "set" of possible values, then it makes sense to have `IntegerType` as a set of all possible `int` passing `Type#validate()`, but for `ConstantType` (a type that is represented by only one possible value, as far as I can see), having `ConstantType#getValue()` is confusing to me, and I couldn't really find usages.

The reason is that a type type doesn't "have" a value: it is logically always a set, sometimes with only ever one item, or no possible type at all, such as `BottomType` (impossible).

lib/Doctrine/Annotations/Type/ConstantType.php
Majkl578: Constant type is a specialization of generic type, e.g. `5` is ConstantIntegerType which is still IntegerType.
There is no EnumType currently - enum is simply interpreted as UnionType of multiple ConstantTypes.

Majkl578:
> for ConstantType (a type that is represented by only one possible value, as far as I can see), having `ConstantType#getValue()` is confusing to me, and I couldn't really find usages.

Technically `getValue()`is not currently needed so can as well be dropped.

Ocramius:
> Technically `getValue()` is not currently needed so can as well be dropped.

That's what I'm mostly sniping at

Majkl578:
> That's what I'm mostly sniping at

Dropped. 🤞

ostrolucky (approved):

Ocramius (approved):

Ocramius:
Nice work! :+1:

PR #257, opened 2019-03-25
Skip parsing annotations containing dashes, such as `@Foo-bar`, or `@Foo-`

Ocramius:
Most likely fixes https://github.com/doctrine/collections/issues/180

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocLexer.php
+++ b/lib/Doctrine/Common/Annotations/DocLexer.php
@@ -79,7 +79,7 @@ final class DocLexer extends AbstractLexer
     );
 
     /**
-     * Whether the token next token starts immediately, of if there were
+     * Whether the next token starts immediately, or if there were
      * non-captured symbols before that
      */
     public function nextTokenIsAdjacent() : bool
--- a/lib/Doctrine/Common/Annotations/DocLexer.php
+++ b/lib/Doctrine/Common/Annotations/DocLexer.php
@@ -51,6 +51,7 @@ final class DocLexer extends AbstractLexer
     const T_TRUE                = 110;
     const T_NULL                = 111;
     const T_COLON               = 112;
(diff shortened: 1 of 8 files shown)


greg0ire:
> Skip parsing annotations containing dashes, such as `@Foo-bar`, or `@Foo`

:confused: 	where is the dash in `Foo`?

Ocramius:
@greg0ire corrected :-)

lib/Doctrine/Common/Annotations/DocLexer.php
```diff
@@ -76,6 +78,17 @@ final class DocLexer extends AbstractLexer
         'null'  => self::T_NULL
     );
 
+    /**
+     * Whether the token next token starts immediately, of if there were
```
greg0ire:
```suggestion
     * Whether the next token starts immediately, of if there were
```

tests/Doctrine/Tests/Common/Annotations/DocLexerTest.php
```diff
@@ -215,4 +215,99 @@ public function testRecognizesDoubleQuotesEscapeSequence()
...
+                [
+                    'value'     => '-56',
+                    'position'  => 7,
+                    'type'      => DocLexer::T_INTEGER,
+                ]
+            ]
+        );
+    }
+
+    /** @return void */
```
greg0ire: Why don't we use a return type hint here?

Ocramius: Good point! I though `1.6` had older PHP version support :)

Ocramius:
Ported to `1.7` via 5cfa972716505d343c1e3fb6f806bdb62604aa82

Ocramius:
Ported to `master` via 6403955bada39e7483be380039c01fae2c177848

PR #258, opened 2019-03-25
Remove imprecise `T_IDENTIFIER` capture groups, leading to simpler token sequences

Ocramius:
Note: I just need a review - will forward-port after that.

Discovered while working on #257 

A `T_IDENTIFIER` represents:

 * a constant
 * a function name
 * a function name prefixed with `\`
 * a class name
 * a class name prefixed with `\`
 * a class constant
 * a class constant, with class name prefixed with `\`

The current capturing regular expression is too simplistic, and it
leads to edge cases where `Foo\42.5` are captured as separate tokens.

In this patch, we deprecate `DocLexer::T_NAMESPACE_SEPARATOR` (which
shouldn't be looked up directly anymore), and tell the parser to work
with wider `T_IDENTIFIER` tokens instead.

Note that a test is also being removed. That is intentional, since the
test verifies a parser issue when parsing `@Foo\3.42`:
 
 * previously, the tokens would be `[@ Foo\3.42]`
 * now they are `[@, Foo, \, 3.42]`

This also means that the above will be parsed as `@Foo`.

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -651,7 +651,7 @@ final class DocParser
             // make sure the @ is followed by either a namespace separator, or
             // an identifier token
             if ((null === $peek = $this->lexer->glimpse())
-                || (DocLexer::T_NAMESPACE_SEPARATOR !== $peek['type'] && !in_array($peek['type'], self::$classIdentifiers, true))
+                || !in_array($peek['type'], self::$classIdentifiers, true)
                 || $peek['position'] !== $this->lexer->lookahead['position'] + 1) {
                 $this->lexer->moveNext();
                 continue;
@@ -988,18 +988,7 @@ final class DocParser
 
         $this->lexer->moveNext();
 
-        $className = $this->lexer->token['value'];
-
-        while ($this->lexer->lookahead['position'] === ($this->lexer->token['position'] + strlen($this->lexer->token['value']))
-                && $this->lexer->isNextToken(DocLexer::T_NAMESPACE_SEPARATOR)) {
-
-            $this->match(DocLexer::T_NAMESPACE_SEPARATOR);
-            $this->matchAny(self::$classIdentifiers);
-
-            $className .= '\\' . $this->lexer->token['value'];
-        }
-
-        return $className;
+        return $this->lexer->token['value'];
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -988,18 +988,7 @@ private function Identifier()
 
         $this->lexer->moveNext();
 
-        $className = $this->lexer->token['value'];
-
-        while ($this->lexer->lookahead['position'] === ($this->lexer->token['position'] + strlen($this->lexer->token['value']))
```
nicolas-grekas: because lookahead can be null here, this line triggers a deprecation on 7.4
that's one of the main remaining blockers we have to make Symfony's tests pass on PHP 7.4 :)

Tobion:
Fixes #273

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -1258,16 +1258,6 @@ public function testSetValuesExeption()
         $this->createTestParser()->parse($docblock, 'some class');
     }
 
-    /**
-     * @expectedException \Doctrine\Common\Annotations\AnnotationException
-     * @expectedExceptionMessage [Syntax Error] Expected Doctrine\Common\Annotations\DocLexer::T_IDENTIFIER or Doctrine\Common\Annotations\DocLexer::T_TRUE or Doctrine\Common\Annotations\DocLexer::T_FALSE or Doctrine\Common\Annotations\DocLexer::T_NULL, got '3.42' at position 5.
-     */
-    public function testInvalidIdentifierInAnnotation()
```
Ocramius: Urgh, I shouldn't have removed this test. It should at least have a test declaring another sort of failure here.

lib/Doctrine/Common/Annotations/DocParser.php
Ocramius: @guilhermeblanco let's separate the concern of fixing 7.4 compat and the issue found with this patch: we should check for the lookahead to exist in this `while()`, and have a test that triggers a deprecation warning on 7.4

guilhermeblanco:
As agreed, we decided to take a simplified approach and only resolve the 7.4 notice instead.
This patch is now considered invalid. Closing.

PR #255, opened 2019-03-11
phpstan updated to 0.11.3

adaamz:

PR #261, opened 2019-03-25
Make phpdoc comments accurate

greg0ire:
This fixes the phpstan job of the build.
Refs #255

### DIFF
--- a/lib/Doctrine/Annotations/Metadata/Builder/PropertyMetadataBuilder.php
+++ b/lib/Doctrine/Annotations/Metadata/Builder/PropertyMetadataBuilder.php
@@ -23,7 +23,7 @@ final class PropertyMetadataBuilder
     /** @var bool */
     private $default = false;
 
-    /** @var array<int|float|string|bool>|null */
+    /** @var array<string, array<int|float|string|bool>|string>|null */
     private $enum;
 
     public function __construct(string $name)
@@ -56,7 +56,7 @@ final class PropertyMetadataBuilder
     }
 
     /**
-     * @param array<int|float|string|bool>|null $enum
+     * @param array<string, array<int|float|string|bool>|string> $enum
      */
     public function withEnum(array $enum) : self
     {
--- a/lib/Doctrine/Annotations/Metadata/PropertyMetadata.php
+++ b/lib/Doctrine/Annotations/Metadata/PropertyMetadata.php
@@ -22,12 +22,12 @@ final class PropertyMetadata
     /** @var bool */
     private $default;
 
-    /** @var array<int|float|string|bool>|null */
+    /** @var array<string, array<int|float|string|bool>|string>|null */
     private $enum;
 
     /**
      * @param array<string, string>             $type
-     * @param array<int|float|string|bool>|null $enum
+     * @param array<string, array<int|float|string|bool>|string>|null $enum
      */
     public function __construct(
         string $name,
@@ -67,7 +67,7 @@ final class PropertyMetadata
     }
 
     /**
-     * @return array<int|float|string|bool>|null
+     * @return array<string, array<int|float|string|bool>|string>|null
      */
     public function getEnum() : ?array
     {
--- a/lib/Doctrine/Annotations/DocParser.php
+++ b/lib/Doctrine/Annotations/DocParser.php
@@ -493,9 +493,6 @@ final class DocParser
 
     /**
      * Collects parsing metadata for a given attribute.
-     *
-     * @param array     $metadata
-     * @param Attribute $attribute
      */
     private function collectAttributeTypeMetadata(
         PropertyMetadataBuilder $metadata,
--- a/lib/Doctrine/Annotations/Metadata/AnnotationMetadata.php
+++ b/lib/Doctrine/Annotations/Metadata/AnnotationMetadata.php
@@ -32,9 +32,6 @@ final class AnnotationMetadata
     /** @var PropertyMetadata|null */
     private $defaultProperty;
 
-    /**
-     * @param PropertyMetadata[] $properties
-     */
     public function __construct(
         string $name,
         AnnotationTarget $target,


lib/Doctrine/Annotations/Metadata/Builder/PropertyMetadataBuilder.php
```diff
@@ -23,7 +23,7 @@ final class PropertyMetadataBuilder
     /** @var bool */
     private $default = false;
 
-    /** @var array<int|float|string|bool>|null */
+    /** @var array<string, array<int|float|string|bool>|string> $enum */
```
greg0ire: Not sure if this is acceptable to you, please tell me what format you prefer (maybe mixed[] + a text description in some format). The actual form of this property is `['value' => [1,2,3], 'literal' => '1,2,3']`

CZechBoY: property can be null - in case when `withEnum` is not called

CZechBoY: and cannot be string, because only array is written to it

lib/Doctrine/Annotations/Metadata/Builder/PropertyMetadataBuilder.php
```diff
@@ -56,7 +56,7 @@ public function withBeingDefault() : self
     }
 
     /**
-     * @param array<int|float|string|bool>|null $enum
+     * @param array<string, array<int|float|string|bool>|string> $enum
```
CZechBoY: Does this fail? String is not subtype of array...

adaamz (changes requested):

CZechBoY:
I think string is not acceptable for $enum.

lib/Doctrine/Annotations/Metadata/Builder/PropertyMetadataBuilder.php
greg0ire: The `|string` is inside the bracket

lib/Doctrine/Annotations/Metadata/Builder/PropertyMetadataBuilder.php
greg0ire: Same reason as above, this is for the `'literal' => '1,2,3'` part

greg0ire:
Me too :P

lib/Doctrine/Annotations/Metadata/Builder/PropertyMetadataBuilder.php
CZechBoY: Ah, I see, you specified the inside array :-) ok then

adaamz (approved):
lgtm

adaamz (changes requested):
I just noticed that in PropertyMetadata is @param for property, there should be @var, please fix.

adaamz (dismissed):

lib/Doctrine/Annotations/Metadata/Builder/PropertyMetadataBuilder.php
```diff
@@ -23,7 +23,7 @@ final class PropertyMetadataBuilder
     /** @var bool */
     private $default = false;
 
-    /** @var array<int|float|string|bool>|null */
+    /** @var ?array<string, array<int|float|string|bool>|string> $enum */
```
Majkl578: `?` is not valid within phpDoc, please use `|null` instead.

lib/Doctrine/Annotations/Metadata/PropertyMetadata.php
```diff
@@ -22,12 +22,12 @@ final class PropertyMetadata
     /** @var bool */
     private $default;
 
-    /** @var array<int|float|string|bool>|null */
+    /** @var ?array<string, array<int|float|string|bool>|string> $enum */
```
Majkl578: `?` is not valid within phpDoc, please use `|null` instead.

lib/Doctrine/Annotations/Metadata/PropertyMetadata.php
```diff
@@ -22,12 +22,12 @@ final class PropertyMetadata
...
     private $enum;
 
     /**
      * @param array<string, string>             $type
-     * @param array<int|float|string|bool>|null $enum
+     * @param ?array<string, array<int|float|string|bool>|string> $enum
```
Majkl578: `?` is not valid within phpDoc, please use `|null` instead.

lib/Doctrine/Annotations/Metadata/PropertyMetadata.php
```diff
@@ -67,7 +67,7 @@ public function isDefault() : bool
     }
 
     /**
-     * @return array<int|float|string|bool>|null
+     * @return ?array<string, array<int|float|string|bool>|string>
```
Majkl578: `?` is not valid within phpDoc, please use `|null` instead.

Majkl578 (changes requested):
Looks ok, besides the `?` part. :)

lib/Doctrine/Annotations/Metadata/PropertyMetadata.php
```diff
@@ -22,12 +22,12 @@ final class PropertyMetadata
     /** @var bool */
     private $default;
 
-    /** @var array<int|float|string|bool>|null */
+    /** @var array<string, array<int|float|string|bool>|string>|null $enum */
```
Majkl578: One more nitpick: properties should not include `$propernyName`.

lib/Doctrine/Annotations/Metadata/Builder/PropertyMetadataBuilder.php
```diff
@@ -23,7 +23,7 @@ final class PropertyMetadataBuilder
     /** @var bool */
     private $default = false;
 
-    /** @var array<int|float|string|bool>|null */
+    /** @var array<string, array<int|float|string|bool>|string>|null $enum */
```
Majkl578: One more nitpick: properties should not include `$propernyName`.

lib/Doctrine/Annotations/Metadata/PropertyMetadata.php
greg0ire: Good catch! Damn copy/paste!

Majkl578 (approved):

Majkl578:
👍

PR #263, opened 2019-03-27
Add Constants Support

jkufner:
Continuation of https://github.com/doctrine/annotations/pull/226. Original commits (excl. merges) cherry-picked onto 1.7 branch.

This pull request adds complete support for constant annotations. It should be the same as property annotations support.

All tests passing.

Example:

`````php
class SomeClass {
    /**
     * @SomeAnnotation
     */
    const SOME_CONSTANT = "foo";
}

$classRefl = new ClassReflection(SomeClass::class);
$constantRefl = $classRefl->getReflectionConstant('SOME_CONSTANT');
$annotations = $annotationReader->getConstantAnnotations($constantRefl);
// $annotations == [ new SomeAnnotation() ]
`````

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -229,7 +229,7 @@ class AnnotationReader implements Reader, ReaderWithConstantsAnnotations
     /**
      * {@inheritDoc}
      */
-    public function getConstantAnnotations(\ReflectionClassConstant $constant)
+    public function getConstantAnnotations(\ReflectionClassConstant $constant): array
     {
         $class   = $constant->getDeclaringClass();
         $context = 'constant ' . $class->getName() . "::" . $constant->getName();
@@ -386,12 +386,8 @@ class AnnotationReader implements Reader, ReaderWithConstantsAnnotations
 
     /**
      * Retrieves imports for constants.
-     *
-     * @param \ReflectionClassConstant $constant
-     *
-     * @return array
      */
-    private function getConstantImports(\ReflectionClassConstant $constant)
+    private function getConstantImports(\ReflectionClassConstant $constant): array
     {
         $class = $constant->getDeclaringClass();
         $classImports = $this->getImports($class);
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -227,6 +227,38 @@ class AnnotationReader implements Reader
     }
 
     /**
+     * {@inheritDoc}
+     */
+    public function getConstantAnnotations(\ReflectionClassConstant $constant)
+    {
+        $class   = $constant->getDeclaringClass();
+        $context = 'constant ' . $class->getName() . "::" . $constant->getName();
+
+        $this->parser->setTarget(Target::TARGET_CONSTANT);
+        $this->parser->setImports($this->getConstantImports($constant));
+        $this->parser->setIgnoredAnnotationNames($this->getIgnoredAnnotationNames($class));
+        $this->parser->setIgnoredAnnotationNamespaces(self::$globalIgnoredNamespaces);
+
+        return $this->parser->parse($constant->getDocComment(), $context);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getConstantAnnotation(\ReflectionClassConstant $constant, $annotationName)
+    {
+        $annotations = $this->getConstantAnnotations($constant);
+
+        foreach ($annotations as $annotation) {
+            if ($annotation instanceof $annotationName) {
+                return $annotation;
+            }
+        }
+
+        return null;
+    }
+
+    /**
      * Gets the annotations applied to a function.
      *
      * @phpstan-return list<object> An array of Annotations.
@@ -352,6 +384,32 @@ class AnnotationReader implements Reader
         return array_merge($classImports, $traitImports);
     }
 
+    /**
+     * Retrieves imports for constants.
+     *
+     * @param \ReflectionClassConstant $constant
+     *
+     * @return array
+     */
+    private function getConstantImports(\ReflectionClassConstant $constant)
+    {
+        $class = $constant->getDeclaringClass();
+        $classImports = $this->getImports($class);
+        if (!method_exists($class, 'getTraits')) {
+            return $classImports;
+        }
+
+        $traitImports = array();
+
+        foreach ($class->getTraits() as $trait) {
+            if ($trait->hasConstant($constant->getName())) {
+                $traitImports = array_merge($traitImports, $this->phpParser->parseClass($trait));
+            }
+        }
+
+        return array_merge($classImports, $traitImports);
+    }
+    
     /**
      * Collects parsing metadata for a given class or function.
      *
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -17,7 +17,7 @@ use function ini_get;
 /**
  * A reader for docblock annotations.
  */
-class AnnotationReader implements Reader
+class AnnotationReader implements Reader, ReaderWithConstantsAnnotations
 {
     /**
      * Global map for imports.
--- a/lib/Doctrine/Common/Annotations/FileCacheReader.php
+++ b/lib/Doctrine/Common/Annotations/FileCacheReader.php
@@ -169,7 +169,7 @@ class FileCacheReader implements Reader, ReaderWithConstantsAnnotations
     /**
      * {@inheritDoc}
      */
-    public function getConstantAnnotations(\ReflectionClassConstant $constant)
+    public function getConstantAnnotations(\ReflectionClassConstant $constant): array
     {
         $class = $constant->getDeclaringClass();
         if ( ! isset($this->classNameHashes[$class->name])) {
--- a/lib/Doctrine/Common/Annotations/FileCacheReader.php
+++ b/lib/Doctrine/Common/Annotations/FileCacheReader.php
@@ -35,7 +35,7 @@ use function var_export;
  *             in version 2.0.0 of doctrine/annotations. Please use the
  *             {@see \Doctrine\Common\Annotations\CachedReader} instead.
  */
-class FileCacheReader implements Reader
+class FileCacheReader implements Reader, ReaderWithConstantsAnnotations
 {
     /** @var Reader */
     private $reader;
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

**Comment by greg0ire** [comment]
I think something needs to be done about the commits. Please fixup commits that fix errors into commits that introduce them, and provide messages in english all the time.

File: lib/Doctrine/Common/Annotations/AnnotationReader.php [hunk 1]
```diff
@@ -292,6 +292,38 @@ public function getMethodAnnotation(ReflectionMethod $method, $annotationName)
         return null;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function getConstantAnnotations(\ReflectionClassConstant $constant)
```
greg0ire: Please add php-7.1-compatible type hints everywhere you can.

lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -378,6 +410,32 @@ private function getPropertyImports(ReflectionProperty $property)
         return array_merge($classImports, $traitImports);
     }
 
+    /**
+     * Retrieves imports for constants.
+     *
+     * @param \ReflectionClassConstant $constant
+     *
+     * @return array
```
greg0ire: Please be more descriptive about the elements of the array (use `Foo[]` if this is an array of foos)

lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -378,6 +410,32 @@ private function getPropertyImports(ReflectionProperty $property)
         return array_merge($classImports, $traitImports);
     }
 
+    /**
+     * Retrieves imports for constants.
+     *
+     * @param \ReflectionClassConstant $constant
```
greg0ire: This one can be removed

lib/Doctrine/Common/Annotations/SimpleAnnotationReader.php
```diff
@@ -83,6 +83,14 @@ public function getPropertyAnnotations(\ReflectionProperty $property)
         return $this->parser->parse($property->getDocComment(), 'property '.$property->getDeclaringClass()->name.'::$'.$property->getName());
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function getConstantAnnotations(\ReflectionClassConstant $constant)
+    {
+        return $this->parser->parse($constant->getDocComment(), 'constant '.$constant->getDeclaringClass()->name.'::'.$constant->getName());
```
greg0ire: Please use linebreaks

lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -292,6 +292,38 @@ public function getMethodAnnotation(ReflectionMethod $method, $annotationName)
         return null;
     }
 
+    /**
+     * {@inheritDoc}
+     */
```
greg0ire, line 231: I don't think any doc is being inherited from anywhere, there is no such method in the interface: https://github.com/doctrine/annotations/blob/6a592da8f416ee23deb419d00ef2f98b6d8fd46a/lib/Doctrine/Common/Annotations/Reader.php#L27

lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -292,6 +292,38 @@ public function getMethodAnnotation(ReflectionMethod $method, $annotationName)
...
+        $this->parser->setImports($this->getConstantImports($constant));
+        $this->parser->setIgnoredAnnotationNames($this->getIgnoredAnnotationNames($class));
+        $this->parser->setIgnoredAnnotationNamespaces(self::$globalIgnoredNamespaces);
+
+        return $this->parser->parse($constant->getDocComment(), $context);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
```
greg0ire, line 247: Same as above

lib/Doctrine/Common/Annotations/FileCacheReader.php
```diff
@@ -159,6 +159,41 @@ public function getPropertyAnnotations(\ReflectionProperty $property)
...
+            && filemtime($path) < filemtime($filename)) {
+            @unlink($path);
+
+            $annot = $this->reader->getConstantAnnotations($constant);
+            $this->saveCacheFile($path, $annot);
+            return $this->loadedAnnotations[$key] = $annot;
+        }
+
+        return $this->loadedAnnotations[$key] = include $path;
+    }
```
greg0ire, line 202: There seems to be a lot of copy/paste in this file, does anybody know why?

jkufner: No idea. But it is beeing removed in 2.0 anyway.

greg0ire: Ok, this might be fine then :)

lib/Doctrine/Common/Annotations/AnnotationReader.php
jkufner: I'm keeping exactly the same API as for properties methods. I believe the consistency is more useful here than perfection. However, this can be improved when porting these changes to the master branch.

lib/Doctrine/Common/Annotations/AnnotationReader.php
jkufner: Returned annotations do not have a common ancestor. The best we can do is object[], but object is not valid in PHP 7.1. Therefore array is the best specification here.

lib/Doctrine/Common/Annotations/AnnotationReader.php
jkufner: Depends on coding style. These useless annotations are at all methods in this file. Should I remove them all?

lib/Doctrine/Common/Annotations/AnnotationReader.php
jkufner: The methods were missing in the interface. Fixed.

jkufner:
Ok, this should do the trick. I've also updated the pull request for 2.0 version (#264).

lib/Doctrine/Common/Annotations/Reader.php
```diff
@@ -86,4 +86,24 @@ function getPropertyAnnotations(\ReflectionProperty $property);
...
+
+    /**
+     * Gets the annotations applied to a constant.
+     *
+     * @param \ReflectionClassConstant $constant The ReflectionClassConstant of the constant
+     *                                           from which the annotations should be read.
+     *
+     * @return array An array of Annotations.
+     */
+    function getConstantAnnotations(\ReflectionClassConstant $constant);
```
greg0ire: Adding methods to an existing interface is a BC-break for implementors. Not sure what the right solution is... Maybe create a new interface, and merge it with this one on master

lib/Doctrine/Common/Annotations/AnnotationReader.php
greg0ire: You could use `mixed[]`

lib/Doctrine/Common/Annotations/AnnotationReader.php
greg0ire: Please remove this one, you can remove the others if you want to, as a bonus, in a separate commit

Majkl578:
I think new features for 1.x are out of question at this point when refactoring for 2.0 is already ongoing.

jkufner:
@Majkl578 Well, I've created two pull requests for both versions. So it can go to 1.7 and refactoring of this feature to 2.0 is already done.

The only differences are in Target enum instead of few constants and in removed readers. Moreover, I did not notice any changes in getMethodAnnotations() and getMethodAnnotation() methods, so I guess the refactoring didn't got too far to cause any troubles here.

alcaeus:
> I think new features for 1.x are out of question at this point when refactoring for 2.0 is already ongoing.

Disagree: with how unclear the timeline for 2.0 is, it makes sense adding this to 1.x as well. I see no problem with releasing a final 1.x release at the same time as 2.0 is released, similar to how Symfony releases a minor release for the previous major along with a new major version.

lib/Doctrine/Common/Annotations/AnnotationReader.php
jkufner: `mixed[]` means exactly the same as `array`.

lib/Doctrine/Common/Annotations/Reader.php
jkufner: Good point. It seems unlikely anybody would implement Reader interface without extending one of the existing implementations, but it is certainly a possibility we need to respect. Therefore, I moved the new methods to a new and deprecated interface `ReaderWithConstantsAnnotations` (which extends `Reader`) and added this interface to all reader implementations. The new interface is already marked as deprecated and removed in the pull request to the 2.0 version. I've intentionally kept redundant `implements Reader` in class declarations to keep the code more readable because the other interface is deprecated and we don't want to confuse anybody.

The v2.0 pull request also contains this interface, but immediately removes it in the next commit, so that it should not cause any trouble if 1.7 branch is ever merged into master.

lib/Doctrine/Common/Annotations/AnnotationReader.php
jkufner: Removed.

lib/Doctrine/Common/Annotations/AnnotationReader.php
greg0ire: It conveys the fact that you did try to see if you could replace `array` with something better but could not, whereas with `array` people might think there is room for improvement

lib/Doctrine/Common/Annotations/AnnotationReader.php
greg0ire: I don't know the option of other maintainers, but IMO, since we are moving towards type hints everywhere, there is no point in making the BC-break that will occur when we finally add them bigger. I think consistency is good, except when it's consistency with a deprecated thing we are trying to move away from.

jkufner: I would break them all at once in 2.0, but as you wish.

jkufner:
I've rebased the MR to the current master (v1.10.x). It seems to work fine as before :)

greg0ire:
@jkufner can you please fix the new conflicts?

jkufner:
@greg0ire I've rebased it to the 1.12.x branch. I hope it still works, it's been a long time.

greg0ire:
>  I hope it still works, it's been a long time.

I'm afraid I have bad news for you :grimacing:

derrabus:
Closing because we won't add new features to this library.

Issue #232, opened 2018-12-18
[meta] Annotations 2.0 (ng)


Majkl578:
# Annotations 2.0 (ng)

This is a meta issue to track Annotations 2.0 milestones/features and aims. It's also a follow-up of #75 which collected some dust over time.

## Backward Compatibility

One of the main goals of the 2.0 version is on minimizing breakages of existing annotations and syntax compatibility with Annotations 1.x. Therefore, when talking about custom annotations and docblocks, _most_ of the user-land stuff should work without any significant effort or none at all. However most of the internal stuff is expected to be changed.

## Namespace change

- [x] The namespace will be changed from `Doctrine\Common\Annotations` to `Doctrine\Annotations`. [#193]

## Dropped features

Some features will be dropped:
- [x] `SimpleAnnotationReader` - _support for so-called unqualified references without `use`s are dropped_ [#199]
- [x] `AnnotationRegistry` - _no more annoying autoloader stuff, annotations are now autoloaded implicitly_ [#205]
- [x] `FileCacheReader` - _dropped in favor of CachedReader (based on doctrine/cache)_ [#202]
- [x] `Doctrine\Common\Annotations\Annotation` - _behavior of this class is redundant with `@Annotation` and `__get()`/`__set()` safety is ouf of scope of this library_ [#246]
- [x] `CachedReader` - _caching is to be done by consumers when needed_ [#249]
- [ ] `DocLexer`, `DocParser` and `doctrine/lexer` dependency - _these will be replaced by new AST-based parser_
---

## New parser

Initial work is split into three different stages and will come as a *new code* without refactoring existing stuff:
1. [ ] Parser - _this is the docblock parser_
   - [ ] PP language - _Doctrine Annotations syntax definition_
   - [ ] Compiler - _docblock to AST converter_
   - [ ] AST structure - _rich domain-specicic AST_
   - [ ] AST builder - _internal AST walker that builds rich AST_
   - [ ] AST walker - _interface for traversing AST_
   - [ ] AST dumper- _utility walker to display AST structure_
   - [ ] Parser parity tests - _compatibility testing between old and new parsers_
2. [ ] Metadata - _this represents an `@Annotation`_
   - [x] Annotation and Property metadata - _information needed to construct annotation objects_ [#247]
   - [x] Property type and enum hiearchy [#256]
   - [ ] AST metadata collector (walker)
   - [ ] Scope - _holds information about subject (origin), active imports, ignored annotations and nesting level_
[...]

PR #271, opened 2019-06-28
Trigger the standard autoloader as the last resort

### LINKED ISSUE #270: AnnotationRegistry is deprecated but has no alternative in v1.6.1
I understand that the AnnotationRegistry is being deprecated and in version 2+ we should just rely on the Composer autoloader, which is awesome.

However, it seems that at least in version 1.6.1, you can't just use the autoloader, you still _must_ use the registry. I think it's wrong to mark the registry members as deprecated when no better alternative exists today. This causes a lot of confusion amongst users and makes downstream deprecation tests somewhat useless, since there's no way to get them to pass today.

I think that version 1.x should be changed to allow using the autoloader for folks that don't want to use the deprecated functions.

sanmai:
As v2 isn't here yet, and looking at #232 it is unclear if it will be coming anytime soon, and as we have to put up with annoying deprecated functions, here I propose to fall back to the standard autoloader if nothing else worked, and if there's no custom loader defined.

Fixes #270

I intentionally did not add any tests as this is a PoC. Please tell me if this makes any sence, and then I'll make a try with tests.

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationRegistry.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationRegistry.php
@@ -64,9 +64,7 @@ final class AnnotationRegistry
     /**
      * Registers file.
      *
-     * @deprecated this method is deprecated and will be removed in doctrine/annotations 2.0
-     *             autoloading should be deferred to the globally registered autoloader by then. For now,
-     *             use @example AnnotationRegistry::registerLoader('class_exists')
+     * @deprecated This method is deprecated and will be removed in doctrine/annotations 2.0. Annotations will be autoloaded in 2.0.
      */
     public static function registerFile(string $file) : void
     {
@@ -83,9 +81,7 @@ final class AnnotationRegistry
      * @param string            $namespace
      * @param string|array|null $dirs
      *
-     * @deprecated this method is deprecated and will be removed in doctrine/annotations 2.0
-     *             autoloading should be deferred to the globally registered autoloader by then. For now,
-     *             use @example AnnotationRegistry::registerLoader('class_exists')
+     * @deprecated This method is deprecated and will be removed in doctrine/annotations 2.0. Annotations will be autoloaded in 2.0.
      */
     public static function registerAutoloadNamespace(string $namespace, $dirs = null) : void
     {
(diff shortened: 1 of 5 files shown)


stof:
shouldn't it check for both `$loaders` and `$autoloadNamespaces` if it want to check that nothing was registered explicitly ?

sanmai:
@stof fair, please see if it is any better now

stof:
:+1: for this, as it means that not registering anything anymore will produce the behavior available in doctrine/annotations 2 (but without forcing anything on existing projects until they remove their explicit registration)

stof:
there is actually still 1 case changing the behavior for existing projects: when only `registerFile` was used.

sanmai:
Well, we can toggle a flag when `registerFile` is used, then proceeding as we do for other methods, but shall we?

sanmai:
Isn't you pre-load a class when you use `AnnotationRegistry::registerFile`? E.g. if you have a class loaded this way, then autoloading won't even trigger. Is there something else I miss here?

stof:
Well, if you used `registerFile`, we might still try to discover a different annotation.

alcaeus:
@sanmai I've rebased this against the new master after making some changes to our branches. Do you still want to work on this patch? If we can get registerFile behaviour sorted out properly, we can include this in the 1.9 release.

sanmai:
Yes, I will at least try to remember what's going on here.

sanmai:
@alcaeus please let me know if this is a passable workaround

alcaeus:
@sanmai this looks good. I've added a couple of tests and updated the deprecation messages. Pending an approval by @stof this is ready for 1.9. Thank you very much for this improvement!

sanmai:
@stof done

alcaeus:
Thanks @sanmai!

sanmai:
Thank you @stof and @alcaeus!

sanmai:
It seems like this PR [didn't get into 1.9](https://github.com/doctrine/annotations/compare/v1.8.0...1.9.0), even though mentioned in the changelog.

Please confirm.

sanmai:
1.9.1 should work just fine IMO.

sanmai:
All clear. Thank you.

Issue #273, opened 2019-07-12
Trying to access array offset on value of type null


garak:
This stopped working on php 7.4, for example on [DocParser class](https://github.com/doctrine/annotations/blob/master/lib/Doctrine/Annotations/DocParser.php#L983) 
See https://twitter.com/nikita_ppv/status/1148931277216800769 for a reference

Ocramius:
Ref: https://github.com/doctrine/annotations/blob/793b91ac63464dff1fa1e949f20214b7eecf6472/lib/Doctrine/Annotations/DocParser.php#L983

rbairwell:
If it helps, the following seems to fix the problem:
`while ($this->lexer->lookahead!==null && $this->lexer->lookahead['position'] === ($this->lexer->token['position'] + strlen($this->lexer->token['value']))`

Tobion:
Symfony also has alot of failures in php 7.4 due to this, ref. https://github.com/symfony/symfony/issues/32844

davidbyoung:
Will the fix be merged into `master` in the near future?  I'm testing 2.0 on PHP 7.4-beta4, and am blocked.  Thanks!

glensc:
RFC that implemented the breakage: https://wiki.php.net/rfc/notice-for-non-valid-array-container

SuperlativeEntity:
Does this still seem to be broken?

ErrorException  : Trying to access array offset on value of type null
vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php:995

`while ($this->lexer->lookahead['position'] === ($this->lexer->token['position'] + strlen($this->lexer->token['value']))`

alcaeus:
@SuperlativeEntity what version are you running? This has been fixed in 1.7.0.

SuperlativeEntity:
Okay, my mistake, I have the wrong version. Thanks for your time.

javiyt:
I'm using version 1.10.2 and its failing in other line:
```
[ErrorException]
 Trying to access array offset on value of type null
in /code/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php:1047
```

SuperlativeEntity:
@javiyt Did you resolve the issue?

javiyt:
No

SuperlativeEntity:
Might be easier to just log a separate issue.

grrenier:
> 
> 
> Does this still seem to be broken?
> 
[...]

PR #274, opened 2019-07-12
Fix access array offset on value of type null

jfcherng:
Fix https://github.com/doctrine/annotations/issues/273.

---

I am using Symfony 4.3 with PHP 7.4.0-dev(nightly snapshot). When I do `bin/console cache:clear -vvv`, there is a warning which will be fixed by this PR.

There is another suspicious one as below.
https://github.com/doctrine/annotations/blob/793b91ac63464dff1fa1e949f20214b7eecf6472/lib/Doctrine/Annotations/DocParser.php#L1030
But it seems not causing a problem in my case. Not sure how it will be generally.

### DIFF
--- a/lib/Doctrine/Annotations/DocParser.php
+++ b/lib/Doctrine/Annotations/DocParser.php
@@ -980,7 +980,8 @@ final class DocParser
 
         $className = $this->lexer->token['value'];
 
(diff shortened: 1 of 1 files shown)


jfcherng:
I see this PR will conflict with https://github.com/doctrine/annotations/pull/258. It's favored hence close this PR.

Issue #277, opened 2019-09-09
@phpcsSuppress annotation is not being ignored.


phpeek:
I'm using `v1.7.0` of this library and seeing this issue.

`@phpcsSuppress` is not being ignored.

I see that PR that's ignoring it has been merged long time ago - https://github.com/doctrine/annotations/pull/229

Ocramius:
Can you write a test case reproducing this, please?

stof:
the PR was merged in master, not in 1.7

phpeek:
@stof @Ocramius Yes, You're right it was targeted at `master`.

I forgot to mention that release notes for 1.7.0 mentions adding this annotation to ignore list.

https://github.com/doctrine/annotations/releases/tag/v1.7.0

I'll try to issue PR with repro test case later.

phpeek:
I've added PR that should fix this targeted at `1.7` branch.

https://github.com/doctrine/annotations/pull/278

Without a test-case for now. Will do that later today.

alcaeus:
As mentioned in the PR, I'll have to double-check why the PR was added to the 1.7 milestone but is not contained in the release. Since the code is already in master, there's no need for an additional PR.

phpeek:
@alcaeus Do we have any info on this issue ? My PR has been canceled since `1.7` is locked for any new features. Since then, I'm using my fork to overcome this and that's totally fine. :) Is there any chance for `1.7.1` release with this fix ?

alcaeus:
Sorry for the delay, and even more sorry for the mishap. It looks like there were three pull requests that were merged to master after we split off a 1.7 branch and were never backported to 1.7, but flagged for the 1.7 milestone. This affects #229, #230, and #235. I'll backport these to the 1.8 branch and tag a new release next week. I might get to that before then, but can't make any guarantees.

If you'd like to backport these yourself, feel free to give it a shot and create separate pull requests to the 1.8 branch for these. I've just pushed a new 1.8 branch, so feel free to add it there.

To avoid this in the future, I've also set the new default branch of the repository to 1.8, as the 2.0 version is still in heavy development, so new features should first target the 1.x release line until we've got a better plan of action for 2.0.

phpeek:
@alcaeus Thank You !

I'll issue a backport PR to `1.8` later today.

alcaeus:
[...]

PR #278, opened 2019-09-09
Added `@phpcsSuppress` to ignore list.

phpeek:
This should fix - https://github.com/doctrine/annotations/issues/277.

I'll push test case later.

This is PR does exactly the same thing that was included in https://github.com/doctrine/annotations/pull/229

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -101,7 +101,9 @@ class AnnotationReader implements Reader
         // PlantUML
         'startuml' => true, 'enduml' => true,
         // Symfony 3.3 Cache Adapter
-        'experimental' => true
+        'experimental' => true,
+        // Slevomat Coding Standard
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -101,7 +101,9 @@ class AnnotationReader implements Reader
         // PlantUML
         'startuml' => true, 'enduml' => true,
         // Symfony 3.3 Cache Adapter
-        'experimental' => true
+        'experimental' => true,
+        // Slevomat Coding Standard
+        'phpcsSuppress' => true
```
Ocramius: Can you please add a trailing comma here? As you can see, forgetting it is causing diff noise here

alcaeus:
Closing here: 1.7.0 is already released and new features aren't accepted. I'll double-check why this is listed in the 1.7 release notes but not contained in the 1.7 branch.

PR #281, opened 2019-09-24
Backporting adding global ignored annotations to DocParser.

phpeek:
Backporting changes made in https://github.com/doctrine/annotations/pull/235 to `1.8` as promised in https://github.com/doctrine/annotations/issues/277.

Not sure if this is needed since tests are passing even without this change.


alcaeus:
Thanks @phpeek!

PR #284, opened 2019-09-28
Fixed DocParser lookahead to work in PHP 7.4

### LINKED ISSUE #283: master is broken in PHP 7.4
This may be considered a duplicate of #273, but 2.0 is currently broken until #276 is merged into `master`.

davidbyoung:
Fixes #283 by explicitly checking if the lookahead position is null before using it (started becoming a problem in PHP 7.4).  Also fixed spl autoload function to have a void return type rather than bool.  Removed an error message from the PHPStan config because it is no longer thrown by the fixed DocParser

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -30,7 +30,3 @@ parameters:
         -
             message: '#^Array has 2 duplicate keys with value .ClassWithConstants\.SOME_KEY. \(\\Doctrine\\Tests\\Annotations\\Fixtures\\ClassWithConstants::SOME_KEY, \\Doctrine\\Tests\\Annotations\\Fixtures\\ClassWithConstants::SOME_KEY\)\.$#'
             path: '%currentWorkingDirectory%/tests\Doctrine\Tests\Annotations\DocParserTest.php'
-        # bug in doctrine/lexer
-        -
-            message: '#^Call to static method PHPUnit\\Framework\\Assert::assertNull\(\) with array will always evaluate to false\.$#'
-            path: '%currentWorkingDirectory%/tests\Doctrine\Tests\Annotations\DocLexerTest.php'
--- a/lib/Doctrine/Annotations/DocParser.php
+++ b/lib/Doctrine/Annotations/DocParser.php
@@ -980,8 +980,11 @@ final class DocParser
 
         $className = $this->lexer->token['value'];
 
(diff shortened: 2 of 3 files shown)


phpstan.neon.dist
```diff
@@ -30,7 +30,3 @@ parameters:
         -
             message: '#^Array has 2 duplicate keys with value .ClassWithConstants\.SOME_KEY. \(\\Doctrine\\Tests\\Annotations\\Fixtures\\ClassWithConstants::SOME_KEY, \\Doctrine\\Tests\\Annotations\\Fixtures\\ClassWithConstants::SOME_KEY\)\.$#'
             path: '%currentWorkingDirectory%/tests\Doctrine\Tests\Annotations\DocParserTest.php'
-        # bug in doctrine/lexer
-        -
-            message: '#^Call to static method PHPUnit\\Framework\\Assert::assertNull\(\) with array will always evaluate to false\.$#'
-            path: '%currentWorkingDirectory%/tests\Doctrine\Tests\Annotations\DocLexerTest.php'
```
greg0ire: Does this related to one of the other 2 commits? If yes, please squash it with that commit, and explain the relationship in the commit message.

davidbyoung: Yes, the error message is no longer reported.  I noticed that PHPStan was throwing an error after my initial commit in [this Travis build](https://travis-ci.org/doctrine/annotations/jobs/590806588).  I'll squash the commits on my end.  Wasn't sure if you were going to squash merge them, or if I should squash them beforehand.

greg0ire: We can do both. Personally, I think it's best if you compose the commit message because you are the one who knows exactly why and how these changes are related.

davidbyoung: I squashed the commits and updated the commit message

malarzm:
We should rather merge https://github.com/doctrine/annotations/pull/276 up as it's doing mostly the same thing and is already in `1.7` branch

greg0ire:
So we should just merge up? Is this repo using a merge up policy?

malarzm:
@greg0ire I assumed it does although I don't know how much master diverged from `1.7`

alcaeus:
I'll merge up soon. I wanted to fix the 1.7 vs. 1.8 issue first (some PRs were marked for 1.7 but only merged to master), then solve the issue by merging up to master.

For what it's worth, we use merge-up strategy here as it requires less work (merging up is generally easier and less error-prone than backporting). As such I'm closing this PR, but I'll leave the related issue open until I've done the merge.

PR #288, opened 2019-10-04
Improve travis settings

BackEndTea:
- `sudo:false` is deprecated and doens't do antying anymore
- `compsoer self-update` is already ran by travis anyways.
- dependencies (and this package) don't declare to be compatible with php 8.0. And most probably wont do so for a while, so this allows to still run tests against that version, instead of producing an error when trying to install the dependencies

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,9 @@ cache:
 before_install:
   - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || echo "xdebug not available"
 
-install: travis_retry composer update --prefer-dist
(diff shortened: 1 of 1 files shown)


greg0ire:
Please move all these insightful explanations to your commit message :pray:

greg0ire:
I asked internally if this should target 1.8, and rewrote the commits, let's wait and see.

greg0ire:
I am going to target 1.8 since the build seems broken on that branch and your changes will definitely help.

greg0ire:
One of your changes was already applied in #299

greg0ire:
Thanks @BackEndTea !

PR #289, opened 2019-10-18
Add support for loading annotations from functions

jvasseur:
I didn't update the `Reader` interface since it would be a BC break to do so.

### DIFF
--- a/lib/Doctrine/Common/Annotations/Annotation/Target.php
+++ b/lib/Doctrine/Common/Annotations/Annotation/Target.php
@@ -25,7 +25,8 @@ final class Target
     public const TARGET_METHOD     = 2;
     public const TARGET_PROPERTY   = 4;
     public const TARGET_ANNOTATION = 8;
-    public const TARGET_ALL        = 15;
+    public const TARGET_FUNCTION   = 16;
+    public const TARGET_ALL        = 31;
 
     /** @var array<string, int> */
     private static $map = [
@@ -33,6 +34,7 @@ final class Target
         'CLASS'      => self::TARGET_CLASS,
         'METHOD'     => self::TARGET_METHOD,
         'PROPERTY'   => self::TARGET_PROPERTY,
+        'FUNCTION'   => self::TARGET_FUNCTION,
         'ANNOTATION' => self::TARGET_ANNOTATION,
     ];
 
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -5,6 +5,7 @@ namespace Doctrine\Common\Annotations;
 use Doctrine\Common\Annotations\Annotation\IgnoreAnnotation;
 use Doctrine\Common\Annotations\Annotation\Target;
 use ReflectionClass;
+use ReflectionFunction;
 use ReflectionMethod;
 use ReflectionProperty;
 
@@ -89,14 +90,14 @@ class AnnotationReader implements Reader
     /**
      * In-memory cache mechanism to store imported annotations per class.
      *
-     * @var array<string, array<string, class-string>>
+     * @psalm-var array<'class'|'function', array<string, array<string, class-string>>>
      */
     private $imports = [];
 
     /**
      * In-memory cache mechanism to store ignored annotations per class.
      *
-     * @var array<string, array<string, true>>
+     * @psalm-var array<'class'|'function', array<string, array<string, true>>>
      */
     private $ignoredAnnotationNames = [];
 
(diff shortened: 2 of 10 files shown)


tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
```diff
@@ -165,4 +165,22 @@ public function testPHPCodeSnifferAnnotationsAreIgnored()
...
         self::assertEmpty($reader->getClassAnnotations($ref));
     }
+
+    public function testFunctionsAnnotation() : void
+    {
+        $reader = $this->getReader();
+        $ref = new \ReflectionFunction('Doctrine\Tests\Common\Annotations\Fixtures\foo');
+
+        $annotations = $reader->getFunctionAnnotations($ref);
+        self::assertInstanceOf(Fixtures\Annotation\Autoload::class, $annotations[0]);
```
SenseException, line 287: Can you please also check for the amount of elements in that annotations array with `assertCount`?

lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -384,15 +426,29 @@ private function getPropertyImports(ReflectionProperty $property)
         return array_merge($classImports, $traitImports);
     }
 
+    private function getFunctionImports(ReflectionFunction $function) : array
```
SenseException: This function shares the same logic as `getClassImports()` except an array key. Can you please outsource the common logic into another private method?

SenseException (changes requested):
Thank you for your contribution. It's too bad that ReflectionClass and ReflectionFunction don't share a common interface for type hints.

lib/Doctrine/Common/Annotations/AnnotationReader.php
jvasseur: I've merged `getClassImports` and `getFunctionImports` into an unique `getImports` method.

SenseException (approved):
Thank you for your contribution.

greg0ire:
I think this feature should be mentioned in the docs.

jvasseur:
@greg0ire I have updated the doc to include mentions of loading annotations for functions

lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -301,41 +302,82 @@ public function getMethodAnnotation(ReflectionMethod $method, $annotationName)
...
+        return $this->parser->parse($function->getDocComment(), $context);
     }
 
     /**
-     * Retrieves imports.
+     * Gets a function annotation.
      *
-     * @param \ReflectionClass $class
+     * @param ReflectionFunction $function       The ReflectionFunction to read the annotations from.
+     * @param string             $annotationName The name of the annotation.
```
greg0ire: I'd remove that line, and maybe also the line before that.

jvasseur: I've added them for consistency with other methods : https://github.com/doctrine/annotations/blob/dc296632db0500c7bd5144161cf96708f61059ff/lib/Doctrine/Common/Annotations/Reader.php#L42-L44

But it's true that they don't bring much value.

SenseException (dismissed):

lib/Doctrine/Common/Annotations/AnnotationReader.php
greg0ire: Then please don't add them, and hopefully the next contributor will try to be consistent with what _you_ did :crossed_fingers:

jvasseur: Done.

jvasseur:
I've rebase the PR to fix the conflicts.

lib/Doctrine/Common/Annotations/Annotation/Target.php
```diff
@@ -25,14 +25,16 @@ final class Target
     public const TARGET_METHOD     = 2;
     public const TARGET_PROPERTY   = 4;
     public const TARGET_ANNOTATION = 8;
-    public const TARGET_ALL        = 15;
+    public const TARGET_FUNCTION   = 16;
+    public const TARGET_ALL        = 31;
```
greg0ire, line 29: That's a BC-break isn't it?

greg0ire:
@jvasseur I rebased your PR but the build fails now, please take a look :pray:

lib/Doctrine/Common/Annotations/Annotation/Target.php
jvasseur: I'm not sure here, these constants are meant to be used as bitmasks and since the new value includes all the bits of the old value it doesn't break BC.

Of course if someone is doing something strange with this value (like comparing directly with 15 instead of using it as a bitmask) it could leads to BC breaks but I don't see a good reason for someone to do that.

jvasseur:
@greg0ire I've fixed the failing tests.

lib/Doctrine/Common/Annotations/Annotation/Target.php
greg0ire: Sounds fair enough, I hadn't noticed that it was the logical OR of all other values

greg0ire (approved):

greg0ire:
Thanks @jvasseur !

PR #291, opened 2019-10-31
Pass context in second argument of annotation constructor

zorn-v:


alcaeus:
Please explain the reason for your patch and add a test for it. Thanks.

zorn-v:
Reason - to show informational errors (e.g. validation in constructor) 
Something like `Parameter "param" is invalid for "@MyAnnotation" in class App\ClassWithAnnotation`

zorn-v:
Test added

greg0ire:
Please kindly squash your commits together. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble. Please include the reason you gave in your commit message

How to do that?

1. `git rebase -i origin/1.9`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already push. Don't forget the `--force` option otherwise git will try to merge both things together.

beberlei:
As we move towards #369 with named arguments in constructor for attribute compatibility I will close this PR, as we shouldn't add new minor features anymore.

PR #293, opened 2019-11-10
addGlobalIgnoredName 'mixin'

HZMarico:
```
use Redis;

/**
 * Class Connector
 * @package Cache
 *
 * @mixin Redis
 */
class Connector
{
    /**
     * @var Redis
     */
    protected $redis;

    /**
     * __call
     *
     * @param string $name
     * @param array $arguments
     *
     * @return mixed
     */
    public function __call(string $name = '', array $arguments = [])
    {
        // call
        return $this->redis->{$name}(...$arguments);
    }
}
```

```
Connector->set('key', 'value');
```

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -96,6 +96,7 @@ class AnnotationReader implements Reader
         'SuppressWarnings' => true,
         // PHPStorm
         'noinspection' => true,
+        'mixin' => true,
         // PEAR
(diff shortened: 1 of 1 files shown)


Ocramius:
I don't think `mixin` should be considered a reserved annotation: too niche, and also an active bad practice

alcaeus:
Agree here. It may be controversial, but if you use this annotations library and a library that parses annotations on its own, you need to manage exceptions yourself or have the other library add an exception. We can’t add exceptions for every tool that’s out there.

PR #295, opened 2019-11-23
Ditch doctrine/cache dependency in favor of PSR-16

greg0ire:
This shows how we could proceed to get rid of `doctrine/cache` entirely

### DIFF
--- a/docs/en/annotations.rst
+++ b/docs/en/annotations.rst
@@ -107,7 +107,22 @@ is very important during development. If you don't set it to true you have to de
 This gives faster performance, however should only be used in production, because of its inconvenience
 during development.
 
-You can also use one of the ``Doctrine\Common\Cache\Cache`` cache implementations to cache the annotations:
+You can use a PSR-16 cache implementation to cache the annotations:
+
+.. code-block:: php
+
+    use Doctrine\Common\Annotations\AnnotationReader;
+    use Doctrine\Common\Annotations\CachedReader;
+
+    $reader = CachedReader::fromPsr16Cache(
+        new AnnotationReader(),
+        new Psr16CacheImplementation(),
+        $debug = true
+    );
+
+You can also use one of the ``Doctrine\Common\Cache\Cache`` cache
+implementations to cache the annotations, but please note that doing so
+is deprecated:
 
 .. code-block:: php
 
--- a/lib/Doctrine/Common/Annotations/CachedReader.php
+++ b/lib/Doctrine/Common/Annotations/CachedReader.php
@@ -20,7 +20,9 @@
 namespace Doctrine\Common\Annotations;
 
 use Doctrine\Common\Cache\Cache;
+use Psr\SimpleCache\CacheInterface;
 use ReflectionClass;
+use Roave\DoctrineSimpleCache\SimpleCacheAdapter;
 
 /**
  * A cache aware annotation reader.
@@ -36,7 +38,7 @@ final class CachedReader implements Reader
     private $delegate;
 
     /**
-     * @var Cache
+     * @var CacheInterface
      */
     private $cache;
 
@@ -51,13 +53,35 @@ final class CachedReader implements Reader
     private $loadedAnnotations = [];
 
     /**
-     * @param bool $debug
+     * @deprecated use fromPsr16Cache instead
+     *
+     * @param Cache|CacheInterface $cache
+     * @param bool                 $debug
      */
-    public function __construct(Reader $reader, Cache $cache, $debug = false)
+    public function __construct(Reader $reader, $cache, $debug = false)
     {
+        if ($cache instanceof Cache) {
+            @trigger_error(sprintf(
+                'Passing a %s instance to %s is deprecated since doctrine/annotations 1.9 and will not be possible anymore in 2.0. Please pass a %s instance',
+                Cache::class,
+                __METHOD__,
+                CacheInterface::class
+            ), E_USER_DEPRECATED);
+            $cache = new SimpleCacheAdapter($cache);
+        }
         $this->delegate = $reader;
-        $this->cache = $cache;
+        $this->cache = (static function (CacheInterface $cache): CacheInterface {
+            return $cache;
+        })($cache);
         $this->debug = (boolean) $debug;
+    }
+
+    public static function fromPsr16Cache(
+        Reader $reader,
+        CacheInterface $cache,
+        $debug = false
+    ): self {
+        return new self($reader, $cache, $debug);
     }
 
     /**
@@ -65,7 +89,7 @@ final class CachedReader implements Reader
      */
     public function getClassAnnotations(ReflectionClass $class)
     {
-        $cacheKey = $class->getName();
+        $cacheKey = strtr($class->getName(), '\\', '.');
 
         if (isset($this->loadedAnnotations[$cacheKey])) {
             return $this->loadedAnnotations[$cacheKey];
@@ -99,7 +123,7 @@ final class CachedReader implements Reader
     public function getPropertyAnnotations(\ReflectionProperty $property)
     {
         $class = $property->getDeclaringClass();
-        $cacheKey = $class->getName().'$'.$property->getName();
+        $cacheKey = strtr($class->getName(), '\\', '.').'$'.$property->getName();
 
         if (isset($this->loadedAnnotations[$cacheKey])) {
             return $this->loadedAnnotations[$cacheKey];
@@ -133,7 +157,7 @@ final class CachedReader implements Reader
     public function getMethodAnnotations(\ReflectionMethod $method)
     {
         $class = $method->getDeclaringClass();
-        $cacheKey = $class->getName().'#'.$method->getName();
+        $cacheKey = strtr($class->getName(), '\\', '.').'#'.$method->getName();
 
         if (isset($this->loadedAnnotations[$cacheKey])) {
             return $this->loadedAnnotations[$cacheKey];
@@ -174,13 +198,13 @@ final class CachedReader implements Reader
     /**
      * Fetches a value from the cache.
      *
-     * @param string $cacheKey The cache key.
+     * @param string $cacheKey A valid PSR-16 cache key.
      *
      * @return mixed The cached value or false when the value is not in cache.
      */
     private function fetchFromCache($cacheKey, ReflectionClass $class)
     {
-        if (($data = $this->cache->fetch($cacheKey)) !== false) {
+        if (($data = $this->cache->get($cacheKey, false)) !== false) {
             if (!$this->debug || $this->isCacheFresh($cacheKey, $class)) {
                 return $data;
             }
@@ -192,23 +216,24 @@ final class CachedReader implements Reader
     /**
      * Saves a value to the cache.
      *
-     * @param string $cacheKey The cache key.
+     * @param string $cacheKey A valid PSR-16 cache key.
      * @param mixed  $value    The value.
      *
      * @return void
      */
     private function saveToCache($cacheKey, $value)
     {
-        $this->cache->save($cacheKey, $value);
+        $this->cache->set($cacheKey, $value);
         if ($this->debug) {
-            $this->cache->save('[C]'.$cacheKey, time());
+            $this->cache->set('[C]'.$cacheKey, time());
         }
     }
 
     /**
      * Checks if the cache is fresh.
      *
-     * @param string $cacheKey
+     * @param string $cacheKey should already be sanitized to be compatible
+     *                         with PSR
      *
      * @return boolean
      */
@@ -219,7 +244,7 @@ final class CachedReader implements Reader
             return true;
         }
 
-        return $this->cache->fetch('[C]'.$cacheKey) >= $lastModification;
+        return $this->cache->get('[C]'.$cacheKey) >= $lastModification;
     }
 
     /**
--- a/tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
@@ -4,6 +4,7 @@ namespace Doctrine\Tests\Common\Annotations;
 
 use Doctrine\Common\Annotations\Annotation;
 use Doctrine\Common\Annotations\AnnotationException;
+use Doctrine\Common\Annotations\Reader;
 use PHPUnit\Framework\TestCase;
 use ReflectionClass, Doctrine\Common\Annotations\AnnotationReader;
 
@@ -470,7 +471,7 @@ abstract class AbstractReaderTest extends TestCase
     }
 
     /**
-     * @return AnnotationReader
+     * @return Reader
      */
     abstract protected function getReader();
 }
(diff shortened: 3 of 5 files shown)


lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -51,16 +53,24 @@ final class CachedReader implements Reader
...
+    public function __construct(Reader $reader, $cache, $debug = false)
     {
+        if ($cache instanceof Cache) {
+            @trigger_error(sprintf(
+                'Passing a %s instance to %s is deprecated since doctrine/annotations 1.9 and will not be possible anymore in 2.0. Please pass a %s instance',
+                Cache::class,
+                __METHOD__,
+                CacheInterface::class
+            ), E_USER_DEPRECATED);
+            $cache = new SimpleCacheAdapter($cache);
```
greg0ire: @Ocramius I need some guidance here, because the adapter will throw if 2 interfaces are not implemented: https://github.com/Roave/DoctrineSimpleCache/blob/165599198031cf2a14b52bc7de6eca970e9dfdf5/src/SimpleCacheAdapter.php#L27-L33 , which means this is a BC break

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -51,16 +53,24 @@ final class CachedReader implements Reader
...
+        }
         $this->delegate = $reader;
-        $this->cache = $cache;
+        $this->cache = (static function (CacheInterface $cache): CacheInterface {
+            return $cache;
+        })($cache);
```
greg0ire: A weird way to not have to generate the error message in case of unsupported type, tell me if it's too weird

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -51,16 +53,24 @@ final class CachedReader implements Reader
...
-    public function __construct(Reader $reader, Cache $cache, $debug = false)
+    public function __construct(Reader $reader, $cache, $debug = false)
     {
+        if ($cache instanceof Cache) {
+            @trigger_error(sprintf(
+                'Passing a %s instance to %s is deprecated since doctrine/annotations 1.9 and will not be possible anymore in 2.0. Please pass a %s instance',
+                Cache::class,
+                __METHOD__,
+                CacheInterface::class
+            ), E_USER_DEPRECATED);
```
greg0ire: Do we have any way to test deprecations currently? I know the phpunit bridge was used on DBAL, and then removed…

greg0ire:
The build fails because the Roave lib require php 7.1, should I drop support for that version of php?

lib/Doctrine/Common/Annotations/CachedReader.php
greg0ire: Maybe these exception could be moved to the methods that require them?

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -51,16 +53,24 @@ final class CachedReader implements Reader
...
-     * @param Reader $reader
-     * @param Cache  $cache
-     * @param bool   $debug
+     * @param Cache|CacheInterface $cache
+     * @param bool                 $debug
      */
-    public function __construct(Reader $reader, Cache $cache, $debug = false)
+    public function __construct(Reader $reader, $cache, $debug = false)
     {
+        if ($cache instanceof Cache) {
```
Ocramius: I'd probably rather deprecate this ctor, and instead add a named ctor for the new cache.

greg0ire: But the named ctor would have to call that deprecated ctor, wouldn't it? Or is there a way around that I'm not aware of?

Ocramius: Correct: we can then make the ctor `private` later on

greg0ire: Sorry, I don't see how this is going to work… on one hand, the named ctor should call the deprecated ctor, on the other hand, the type declaration will make that impossible since it will prevent feeding a PSR-16 cache instance to the deprecated constructor.

Ocramius: Instead of adding a `trigger_error`, we deprecate the public ctor, and add a new one. The type hint has to go regardless

tests/Doctrine/Tests/Common/Annotations/LegacyCachedReaderTest.php
```diff
@@ -0,0 +1,165 @@
...
+namespace Doctrine\Tests\Common\Annotations;
+
+use Doctrine\Common\Cache\Cache;
+use Doctrine\Common\Cache\CacheProvider;
+use Doctrine\Tests\Common\Annotations\Fixtures\Annotation\Route;
+use Doctrine\Common\Annotations\AnnotationReader;
+use Doctrine\Common\Annotations\CachedReader;
+use Doctrine\Common\Cache\ArrayCache;
+
+class LegacyCachedReaderTest extends AbstractReaderTest
```
greg0ire: This is a copy of `CachedReaderTest` before this PR, so that the deprecated code path is still tested and can be easily removed in the next major release.

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
```diff
@@ -470,7 +471,7 @@ public function testWillFailOnAnnotationConstantReferenceContainingDashes()
     }
 
     /**
-     * @return AnnotationReader
+     * @return Reader
```
greg0ire: I noticed this was wrong thanks to phpstan

lib/Doctrine/Common/Annotations/CachedReader.php
alcaeus: I would suggest throwing a deprecation message if the constructor is called with a doctrine/cache instance instead of PSR-16. The general `@deprecated` message on the constructor should stay there as well. This would appropriately trigger the deprecation for people using the Symfony PHPUnit Bridge as well as people using tools like phpstan-deprecation-rules.

greg0ire: That's what I originally did. Restored :)

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -51,19 +53,37 @@ final class CachedReader implements Reader
...
+        }
         $this->delegate = $reader;
-        $this->cache = $cache;
+        $this->cache = (static function (CacheInterface $cache): CacheInterface {
+            return $cache;
+        })($cache);
         $this->debug = (boolean) $debug;
     }
 
+    public static function fromPsr16Cache(
```
stof: should it be `from` or `for` ?

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -178,14 +198,14 @@ public function clearLoadedAnnotations()
...
-     * @param string          $cacheKey The cache key.
-     * @param ReflectionClass $class    The related class.
+     * @param string $cacheKey The cache key.
      *
      * @return mixed The cached value or false when the value is not in cache.
      */
     private function fetchFromCache($cacheKey, ReflectionClass $class)
     {
-        if (($data = $this->cache->fetch($cacheKey)) !== false) {
+        $cacheKey = strtr($cacheKey, '\\', '.');
```
stof: should this be done when generating the cache key rather than altering it here ? I find it confusing that some methods (like `isCacheFresh`) assume a valid PSR-16 cache key while others use the invalid cache key, with the same variable names. Making sure that we have valid PSR-16 cache keys everywhere in variables named `$cacheKey` seems better to me (and AFAICT, we only have 2 places generating them, so that won't duplicate the logic more than your current approach)

tests/Doctrine/Tests/Common/Annotations/LegacyCachedReaderTest.php
```diff
@@ -0,0 +1,165 @@
...
+        $cache->expects(self::never())->method('save');
+
+        $reader = new CachedReader(new AnnotationReader(), $cache, true);
+
+        $this->assertEquals([$route], $reader->getClassAnnotations(new \ReflectionClass($className)));
+    }
+
+    protected function getReader()
+    {
+        $this->cache = new ArrayCache();
```
stof: Do we also have a test running the CachedReader against an actual PSR-16 implementation, or only against the mock ? IMO, having a functional test of the CachedReader would be important (as PSR-16 implementations enforce some rules on the cache keys for instance).

This legacy test actually performs such a functional test for now (as the actual doctrine/cache cache will be wrapped by the actual PSR-16 adapter), but this legacy test is meant to be removed in the future, and we should not loose that functional testing.

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -247,8 +267,7 @@ private function getLastModification(ReflectionClass $class)
     }
 
     /**
-     * @param ReflectionClass $reflectionTrait
-     * @return int
+     * @return int|false
```
stof: wrong phpdoc. `max` return `false` only for empty arrays, and we always pass a non-empty array.

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -204,34 +224,34 @@ private function fetchFromCache($cacheKey, ReflectionClass $class)
...
-        return $this->cache->fetch('[C]'.$cacheKey) >= $lastModification;
+        return $this->cache->get('[C]'.$cacheKey) >= $lastModification;
     }
 
     /**
      * Returns the time the class was last modified, testing traits and parents
      *
-     * @param ReflectionClass $class
-     * @return int
+     * @return int|false
```
stof: wrong phpdoc. `max` return `false` only for empty arrays, and we always pass a non-empty array.

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -204,34 +224,34 @@ private function fetchFromCache($cacheKey, ReflectionClass $class)
...
-     * @param ReflectionClass $class
+     * @param string $cacheKey should already be sanitized to be compatible
+     *                         with PSR
      *
      * @return boolean
      */
     private function isCacheFresh($cacheKey, ReflectionClass $class)
     {
-        if (null === $lastModification = $this->getLastModification($class)) {
+        if (false === $lastModification = $this->getLastModification($class)) {
```
stof: this should actually compare it to `0`, which is what will be returned when reading annotations from built-in classes (classes not coming from a file)

docs/en/annotations.rst
```diff
@@ -106,7 +107,23 @@ is very important during development. If you don't set it to true you have to de
...
 during development.
 
-You can also use one of the ``Doctrine\Common\Cache\Cache`` cache implementations to cache the annotations:
+You can use a PSR-16 cache implementation to cache the annotations:
+
+.. code-block:: php
+
+    use Doctrine\Common\Annotations\AnnotationReader;
+    use Doctrine\Common\Annotations\CachedReader;
+    use Doctrine\Common\Cache\ApcCache;
```
stof: give an example using ApcuCache instead, as ApcCache is deprecated (due to using the legacy extension)

stof: btw, this use statement is actually unused here.

tests/Doctrine/Tests/Common/Annotations/LegacyCachedReaderTest.php
stof: OK, looking at it again, CachedReader does not always use mocks, so it already fits it.

tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php
```diff
@@ -95,34 +98,36 @@ public function testUsesFreshCacheWithTraitsThatUseOtherTraits()
 
     protected function doTestCacheStale($className, $lastCacheModification)
     {
-        $cacheKey = $className;
+        $cacheKey       = strtr($className, '\\', '.');
```
stof: why this extra indentation ?

lib/Doctrine/Common/Annotations/CachedReader.php
greg0ire: The implication is "Build a cached reader from a PSR 16 cache (and a wrapped cache)". We build it for the wrapped reader, from a cache. At least that's my reasoning, it would be great to have a native speaker chime in here.

lib/Doctrine/Common/Annotations/CachedReader.php
greg0ire: Good point, will do 👍

tests/Doctrine/Tests/Common/Annotations/LegacyCachedReaderTest.php
greg0ire: Yes we do, I copy/pasted/adapted this file to ` tests/Doctrine/Tests/Common/Annotations/LegacyCachedReaderTest.php`

lib/Doctrine/Common/Annotations/CachedReader.php
greg0ire: I wasn't so sure but if you are, I will change it.

tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php
greg0ire: Probably a remnant from a previous version of the code, I will remove it.

lib/Doctrine/Common/Annotations/CachedReader.php
stof: well, there is a second case where it can return `false`, but that's when you give it an array of boolean.
The signature of `max` would be described like that:

- `max(empty-array): false` (assuming a non-existent `empty-array` type)
- `max(non-empty-array<T>): T` (using the existent `non-empty-array` type of Psalm)

lib/Doctrine/Common/Annotations/CachedReader.php
lcobucci: Using an error handler as we did on the ORM is the simplest way I can think of.

lib/Doctrine/Common/Annotations/CachedReader.php
lcobucci: All our providers inherit from the abstract class, which implements all three interfaces. Maybe we should expect that abstraction instead of the interface?

lib/Doctrine/Common/Annotations/CachedReader.php
lcobucci: Maybe using an assertion would be simpler, or just an if? Removes a function call which is good in this context

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
lcobucci: Perhaps this could be in a different commit 😁

greg0ire: You're absolutely right 👍

greg0ire:
Closing for now, we are discussing PSR-6 vs PSR-16 internally and need to come up with a final decision.

PR #296, opened 2019-11-30
Non controversional pre cache migration

greg0ire:
I extracted here parts of #295 to make that one more readable. This should also pass the build, contrary to #295

### DIFF
--- a/lib/Doctrine/Common/Annotations/CachedReader.php
+++ b/lib/Doctrine/Common/Annotations/CachedReader.php
@@ -214,7 +214,7 @@ final class CachedReader implements Reader
      */
     private function isCacheFresh($cacheKey, ReflectionClass $class)
     {
-        if (null === $lastModification = $this->getLastModification($class)) {
+        if (0 === $lastModification = $this->getLastModification($class)) {
             return true;
         }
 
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -213,14 +208,13 @@ private function saveToCache($cacheKey, $value)
...
-     * @param string           $cacheKey
-     * @param ReflectionClass $class
+     * @param string $cacheKey
      *
      * @return boolean
      */
     private function isCacheFresh($cacheKey, ReflectionClass $class)
     {
-        if (null === $lastModification = $this->getLastModification($class)) {
+        if (0 === $lastModification = $this->getLastModification($class)) {
```
lcobucci: Let's remove the assignment from the condition and the yoda expression? Two nasty things in a single line is too much for me 🤣

greg0ire: Well I'm not going to be the one who defends Yoda, that's for sure :P https://dev.to/greg0ire/why-using-yoda-conditions-you-should-probably-not

PR #298, opened 2019-12-19
Ignore PHPStan generics annotations

marcortola:
Added @template, @extends and @implements from PHPStan to globally ignored annotations list.

- https://medium.com/@ondrejmirtes/generics-in-php-using-phpdocs-14e7301953
- https://github.com/phpstan/phpstan/issues/2450

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -107,6 +107,8 @@ class AnnotationReader implements Reader
         // PHP CodeSniffer
         'codingStandardsIgnoreStart' => true,
         'codingStandardsIgnoreEnd' => true,
+        // PHPStan
+        'template' => true, 'implements' => true, 'extends' => true,
     ];
 
     /**
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -108,7 +108,7 @@ class AnnotationReader implements Reader
         'codingStandardsIgnoreStart' => true,
         'codingStandardsIgnoreEnd' => true,
         // PHPStan
-        'template' => true, 'implements' => true, 'extends' => true,
+        'template' => true, 'implements' => true, 'extends' => true, 'use' => true,
     ];
 
     /**
--- a/tests/Doctrine/Tests/Common/Annotations/Fixtures/ClassWithPHPStanGenericsAnnotations.php
+++ b/tests/Doctrine/Tests/Common/Annotations/Fixtures/ClassWithPHPStanGenericsAnnotations.php
@@ -20,11 +20,24 @@ class ClassWithPHPStanExtendsAnnotationsGeneric
 
 /**
  * @template T
+ */
+trait GenericPHPStanTrait
+{
+
+}
+
+/**
+ * @template T
  * @implements WithPHPStanExtendsAnnotationsInterface<int>
  * @extends ClassWithPHPStanExtendsAnnotationsGeneric<int>
  */
 class ClassWithPHPStanGenericsAnnotations extends ClassWithPHPStanExtendsAnnotationsGeneric implements WithPHPStanExtendsAnnotationsInterface
 {
+    /**
+     * @use GenericPHPStanTrait<T>
+     */
+    use GenericPHPStanTrait;
+
     /**
      * @var array<T>
      */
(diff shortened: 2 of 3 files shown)


ondrejmirtes:
Do annotations with `-` need to be ignored or not? Like `@phpstan-param` or `@template-covariant`...

ondrejmirtes:
There's also `@use` - reserved for generic traits. (They're not supported yet.)

stof:
@

> Do annotations with `-` need to be ignored or not? Like `@phpstan-param` or `@template-covariant`...

@ondrejmirtes no. They are already ignored automatically, because `-` is not valid in a class name anyway.

alcaeus:
> Do annotations with `-` need to be ignored or not?

This issue was fixed when psalm introduced `@psalm-template`. 

> There's also `@use` - reserved for generic traits. (They're not supported yet.)

Good point. @mortola can you please add `use` to the list? Thanks!

lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -107,6 +107,8 @@ class AnnotationReader implements Reader
         // PHP CodeSniffer
         'codingStandardsIgnoreStart' => true,
         'codingStandardsIgnoreEnd' => true,
+        // PHPStan
+        'template' => true, 'implements' => true, 'extends' => true,
```
alcaeus: `@use` should be excluded as well.

Ocramius: These annotations are too widespread to be ignored, specifically talking about existing applications that rely on things such as `@template("mystuff.twig")`, for example.

Usage of `@phpstan-template` (already ignored by the parser) is to be endorsed instead

Ocramius: These annotations are too widespread to be ignored, specifically talking about existing applications that rely on things such as `@template("mystuff.twig")`, for example.

Usage of `@phpstan-template` (already ignored by the parser) is to be endorsed instead

tests/Doctrine/Tests/Common/Annotations/Fixtures/ClassWithPHPStanGenericsAnnotations.php
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * @template T
+ */
+interface WithPHPStanExtendsAnnotationsInterface
+{
+
+}
+
+/**
+ * @template T
```
Ocramius: `@template()` is still valid downstream, and changing rules around it is a major BC break at this point in time

lib/Doctrine/Common/Annotations/AnnotationReader.php
alcaeus: If the `@template` annotation was autoloaded (which it should be), there's no BC break to be found here.

stof: also, these ignore rules are case sensitive according to the comment above.

Due to PSR-4 autoloading being as case-sensitive as your filesystem (meaning it is, for most servers), the code probably uses `@Template("mystuff.twig")`

alcaeus: Chatted with @nicolas-grekas about this and he doesn't see a problem for Symfony: the `@Template` annotation is "kinda deprecated", and even then it would continue to work as long as users probably autoload this.

Of course, other applications still might be affected, but all the change would do is that we would no longer error if the annotation wasn't autoloader before.

Ocramius: This is a general problem, not framework specific.

If there is really not a problem with it, then I'd need a test showing a working `@template` annotation with this patch applied (or an existing test)

alcaeus: Following up, we have a fixture class that includes multiple `@Template` annotations: https://github.com/doctrine/annotations/blob/346c8bc45a85a53012a29f86e2b11322c0679f21/tests/Doctrine/Tests/Common/Annotations/Fixtures/Controller.php#L16. This file continues to be parsed correctly. Obviously, aliasing any annotation as `@template` would cause this to be an issue, but since the exclude list is case-sensitive I don't think this should be too big a problem.

alcaeus:
Thanks @marcortola!

localheinz:
Any chance this can get a release soon?

alcaeus:
@localheinz my apologies, I got side-tracked and completely forgot about this. I made some changes to the branch structure, but want to revisit #271 in time for 1.9. Please shout at me if we don't have a release within a week 😉

PR #301, opened 2019-12-23
CacheReader::getLastModification calls filemtime too much

### LINKED ISSUE #300: [performance] CacheReader::isCacheFresh cause too much call to filemtime
Context: big project, lof of entities that use annotations and a lot of traits.

The current implementation does something like this

```
foreach $entityClasses as $classes

	getClassAnnotation => x90
		fetchFromCache
			isCacheFresh
				filemtime for the $class, is parents and traits (easily 5 or 6 files)
	getMethodAnnotations
		foreach properties => 803
			fetchFromCache
				isCacheFresh
					filemtime for the $class, is parents and traits (easily 5 or 6 files)
	getPropertyAnnotations
		foreach properties => x1076
			fetchFromCache
				isCacheFresh
					filemtime for the $class, is parents and traits (easily 5 or 6 files)
```

For only 96 files, I'm calling 10951 filemtime.

image

With a simple local cache on `getLastModification` with `$filename` as key (and we also can add on traits, but the cardinality is alreadly low enough), performance are up again \o/.

```php
    /**
     * Returns the time the class was last modified, testing traits and parents
     isCacheFresh*
     * @param ReflectionClass $class
     * @return int
     */
    private function getLastModification(ReflectionClass $class)
    {
        $filename = $class->getFileName();
		$cacheKey = '[T]' . $filename;
        
        if (isset($this->loadedAnnotations[$cacheKey])) {
            return $this->loadedAnnotations[$cacheKey];
        }
        $parent   = $class->getParentClass();

        return $this->loadedAnnotations[$cacheKey] = max(array_merge(
            array_map([$this, 'getTraitLastModificationTime'], $class->getTraits()),
        ));
    }

```

image

image

bastnic:
Fix #300.

### DIFF
--- a/lib/Doctrine/Common/Annotations/CachedReader.php
+++ b/lib/Doctrine/Common/Annotations/CachedReader.php
@@ -51,6 +51,11 @@ final class CachedReader implements Reader
     private $loadedAnnotations = [];
 
     /**
+     * @var int[]
+     */
+    private $loadedFilemtimes = [];
+
+    /**
      * @param bool $debug
      */
     public function __construct(Reader $reader, Cache $cache, $debug = false)
@@ -169,6 +174,7 @@ final class CachedReader implements Reader
     public function clearLoadedAnnotations()
     {
         $this->loadedAnnotations = [];
+        $this->loadedFilemtimes = [];
     }
 
     /**
@@ -230,6 +236,11 @@ final class CachedReader implements Reader
     private function getLastModification(ReflectionClass $class)
     {
         $filename = $class->getFileName();
+
+        if (isset($this->loadedFilemtimes[$filename])) {
+            return $this->loadedFilemtimes[$filename];
+        }
+
         $parent   = $class->getParentClass();
 
         $lastModification =  max(array_merge(
@@ -241,7 +252,7 @@ final class CachedReader implements Reader
 
         assert($lastModification !== false);
 
-        return $lastModification;
+        return $this->loadedFilemtimes[$filename] = $lastModification;
     }
 
     /**
@@ -251,6 +262,10 @@ final class CachedReader implements Reader
     {
         $fileName = $reflectionTrait->getFileName();
 
+        if (isset($this->loadedFilemtimes[$fileName])) {
+            return $this->loadedFilemtimes[$fileName];
+        }
+
         $lastModificationTime = max(array_merge(
             [$fileName ? filemtime($fileName) : 0],
             array_map([$this, 'getTraitLastModificationTime'], $reflectionTrait->getTraits())
@@ -258,6 +273,6 @@ final class CachedReader implements Reader
 
         assert($lastModificationTime !== false);
 
-        return $lastModificationTime;
+        return $this->loadedFilemtimes[$fileName] = $lastModificationTime;
     }
 }
--- a/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php
@@ -7,6 +7,8 @@ use Doctrine\Tests\Common\Annotations\Fixtures\Annotation\Route;
 use Doctrine\Common\Annotations\AnnotationReader;
 use Doctrine\Common\Annotations\CachedReader;
 use Doctrine\Common\Cache\ArrayCache;
+use Doctrine\Tests\Common\Annotations\Fixtures\ClassThatUsesTraitThatUsesAnotherTraitWithMethods;
+use PHPUnit\Framework\MockObject\MockObject;
 
 class CachedReaderTest extends AbstractReaderTest
 {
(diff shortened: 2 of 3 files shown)


bastnic:
Should I also add it here https://github.com/doctrine/annotations/pull/301/files#diff-6c76e9c6aaae35aeb7e3134d2604da9fR258? it avoids calling it too much on Traits.

greg0ire:
Please improve your commit message according to [the contributing guide](https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches).

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -230,6 +230,12 @@ private function isCacheFresh($cacheKey, ReflectionClass $class)
     private function getLastModification(ReflectionClass $class)
     {
         $filename = $class->getFileName();
+        $cacheKey = '[T]'.$filename;
+
+        if (isset($this->loadedAnnotations[$cacheKey])) {
```
greg0ire: I'd use a separate array for this, and type it with `int[]`, that way there will be no need to prefix it and SA tools will be happy.

greg0ire:
Should new tests be added?

lib/Doctrine/Common/Annotations/CachedReader.php
bastnic: @greg0ire I suggest to name it `loadedFilemtimes`. My issue is to purge it, can I reset it in the method names `clearLoadedAnnotations`. Seems counter productive, but I do not see it feasable to add another method to clear things, and it will not be backward compatible.

greg0ire: It sounds reasonable to clear it in the existing method to me

bastnic:
@greg0ire is it better for you?

I can test that the array is filled and cleared. A useful test would be to create a scenario where same file is checked multiple times and assert that filemetime is not called more than once per file.

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -251,13 +262,17 @@ private function getTraitLastModificationTime(ReflectionClass $reflectionTrait)
...
+
         $lastModificationTime = max(array_merge(
             [$fileName ? filemtime($fileName) : 0],
             array_map([$this, 'getTraitLastModificationTime'], $reflectionTrait->getTraits())
         ));
 
         assert($lastModificationTime !== false);
 
-        return $lastModificationTime;
+        return $this->loadedFilemtimes[$fileName] = $lastModificationTime;
```
bastnic: mind the typo/cases: `$fileName` vs `$filename` and `$lastModification` vs `lastModificationTime`.

Not cool to homogenize it in a feature PR I guess.

greg0ire: It is cool as long as you do it in a separate commit

bastnic:
@greg0ire tests added. I needed to use some new fixtures.

morozov:
I'm not familiar with the project but since I was requested to review, why bother about caching modification times? As far as I understand, for a given entry, the annotations should be either read from cache or built from reflection. As soon as the annotations are in memory, it doesn't matter where they were read from and whether the cache is valid.

If we read the same entry more than once, it makes more sense to keep it in memory and just return as is instead of keeping the file modification time in memory and read the cache every time. Doesn't it?

bastnic:
Hi @morozov, I tried to explain everything in the related issue, with the blacfkire gains.

Abstract is, a given file is checked (filemtime) for himself, for each property and for each methods. During the same request. So it's loaded a big big number of time, always more than once per file if you do not have an empty class.
It's even worse with traits. As they are loaded even more. 

I'm trying to prevent it. It does not provide any risk, as it's behind another local cache that has a controlled lifetime, which i'm plugged on.

On a big project, we are talking 10s of ms per request.

morozov:
@bastnic I get it why it's slow, what I don't get is why it needs to hit the cache of a specific file for every element of every parent class and trait? Caching is not free. The data stored in cache may be needed on one case but not needed in others and just waste memory. Maybe instead of the API that returns annotations for a given element, we could have the API that returns them for the whole class or even a hierarchy?

In #300, you attached a profiling screenshot but didn't attach a code example. Can we start with some code that reproduces the issue?

greg0ire:
> I'm not familiar with the project but since I was requested to review

Sorry, I was a bit hasty when looking for people that could review and realised that I pinged everyone just after doing it :man_facepalming: . Anyway, thanks for reviewing @morozov :pray:

bastnic:
@morozov code sample is a simple Symfony project with doctrine entities (quite a few of them). I think the tests added and the description of them show the problem: two methods, isFresh called two times.

> The data stored in cache may be needed in one case but not needed in others and just waste memory
We are talking of one int value for each file watched. It's several order of magnitude less than the loadedAnnotation.

> Maybe instead of the API that returns annotations for a given element, we could have the API that returns them for the whole class or even a hierarchy?
Indeed that was my first though, but not the same type of code than improving performance without any BC, so I made this first patch.

On certain cases where we only need one method's annotation on a big file, it can be less efficient, it's always a tradeoff. 

In symfony/property-info, we load the content only once (and so only once check the filemetime) and then we read from that loaded content on demand, since this PR https://github.com/symfony/symfony/pull/32188.

morozov:
Without the code, it's even harder to assess the validity of the approach. So far, the approach of adding a static cache and a couple of if-statements looks like a band-aid for an improperly designed or improperly used API.

In pseudo-code, this is what I mean. We have a function that returns a tiny portion of the needed data and it's invoked repeatedly to retrieve the entire set:
```
fn get_value(id)
    mysql_connect()
    mysql_query(SELECT value FROM table where id = $id)
    mysql_close()
end

for i in 1..10
    get_value(id)
end
```
Once it became evident that it has poor performance, you're trying to optimize its implementation by caching some internals:
```
mysql_connect → mysql_pconnect
```
However, a better solution might be to have or use another API that is more specialized for this case:
```
fn get_values(min, max)
    mysql_connect()
    mysql_query(SELECT value FROM table where id BETWEEN $min AND $max)
    mysql_close()
end

get_values(1, 10)
```
As you can see, there's no caching done, the function call doesn't have additional memory footprint and yields the same result as the "optimized" one.

That's why I'm asking for code. Without the code and without being familiar with the package, it's hard for me to understand what is being optimized.

bastnic:
I understand what you mean. Maybe we need a new API that get everything for a given class. It will be a big change, for all userland. It will need be on master.
More on this at the end of this message.

This ticket is about what I consider a bug AND a performance issue to hit the disk several time to get the same information as the current api only gives separate methods, properties and class annotations. If you want to get everything, you will hit at least three times its filemtime, and if it had 5 traits, hits three time these same traits. And if you have 100 entities, you will hit 100x3 times to get the filemtime of these traits. Same for all the interfaces and parents classes of course.
As this it totally safe and totally revertable (no new api, changes only in private methods), I purpose to add this in the lowest version possible and be happy with the improvements for everyone.

In the meantime, I will check how Symfony is doing things, certainly prefers getting annotations by batch to avoid 80% of these filemtime. And see if there is room for the potential new api on master and then provide a PR here.

Is it a scenario we can agree on?

Purposed new method that return every annotations for a given class
```
interface Reader {

    /**
     * Gets ALLthe annotations applied to a class.
     * 
     * It will return an array (new object)
     * class => [/* all class annotations */ ],
     * methods => [/* all methods annotations */ ],
     * properties => [/* all properties annotations */ ],
     *
     *
     *
     * @param \ReflectionClass $class The ReflectionClass of the class from which
     *                                the class annotations should be read.
     *
     * @return array An array of array of Annotations.
     */
    function getAnnotations(\ReflectionClass $class);

}
```

I'm still not sure which way around it is:
- this method calls all the others but it provide the filemtime, so no need for them to compute it 
- OR this is calls by default to compute the class only once and then all the other methods use the computed values.
This method is awesome in a stack with warmer like symfony.

There is pros and cons for each.

Note that in debug mode, even if you compute everything at the first pass, ok we will avoid the 41432 calls to check the freshness for each property / methods,... but you still will need to have a cache for parent / traits to avoid hitting them for each files. Something handled by this PR.

morozov:
> Proposed new method that return every annotations for a given class

Yeah, this is something that I believe a batch API should look like. Not necessarily returning an array but rather a DTO.

Please also do not disregard https://github.com/doctrine/annotations/issues/300#issuecomment-568449134. If the code is modified after the annotations were cached, there doesn't seem to be a way to invalidate this cache. And the fact that the caller needs to invalidate the cache manually as opposed to relying on mtime, is a BC break on its own.

bastnic:
> Without the code, it's even harder to assess the validity of the approach. So far, the approach of adding a static cache and a couple of if-statements looks like a band-aid for an improperly designed or improperly used API.

FYI, ALL the calls come from https://github.com/doctrine/orm/blob/ca95b0ee1317cbd5d8a745cbde0980ce3d70447c/lib/Doctrine/ORM/Mapping/Driver/AnnotationDriver.php#L52. We can see that for each class, we ask for its annotation and then it's called for each property / methods.

image

I see that in master branch this code has been refactored, but I can't test it easily.

alcaeus:
Calling this good to go. Even though it might be considered a bug, I'm merging this to 1.9.0 due to the potential impact on long-running services. Thanks for tackling this, @bastnic!

PR #303, opened 2020-01-18
Respect ignored namespaces when imported in use statement

### LINKED ISSUE #302: addGlobalIgnoredNamespace won't work with Annotations imported via use statement. Bug ?
I used the
```
AnnotationReader::addGlobalIgnoredNamespace('JMS');
``` 
to ignore Serializer Annotations. I expected this to ignore Annotations imported via use Statement as well. 
I don't know if this is a Bug or intended Behaviour.

Example: https://github.com/re2bit/annotations/commit/34f5be4fddc2396d9a427166bf9b249d82fd1800

alcaeus:
Fixes #302, ping @re2bit.

When a use statement is found for a relative annotation class, it is always marked as found, which bypasses the ignored namespace check. However, if the class itself does not exist, an exception is thrown even though the namespace is ignored. This PR fixes that issue.


tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
```diff
@@ -114,6 +115,21 @@ public function testPropertyAnnotationIsIgnored()
         self::assertEmpty($reader->getPropertyAnnotations($ref->getProperty('property')));
     }
 
+    /**
+     * @group 45
+     *
+     * @runInSeparateProcess
+     */
+    public function testPropertyAnnotationWithUseStatementIsIgnored()
```
greg0ire:
```suggestion
    public function testPropertyAnnotationWithUseStatementIsIgnored() : void
```

alcaeus:
Closing as this is fixed by #244.

PR #314, opened 2020-03-27
Add ext-tokenizer as requirement to composer.json

### LINKED ISSUE #313: Usage of Tokenizer extension not document
In `TokenParser` a method from the PHP Tokenizer extension (`token_get_all()`) is used, though that this extension is required is not documented. I think it should be added to composer.json or a meaningful exception should be thrown if the method is going to be called, but is missing.

dontub:
Fixes #313

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,7 @@
     ],
     "require": {
         "php": "^7.1",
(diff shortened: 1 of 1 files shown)


greg0ire:
Thanks @dontub !

PR #315, opened 2020-03-31
Ignore pre- and postCondition PHPUnit annotations

rybakit:
See https://github.com/sebastianbergmann/phpunit/blob/7a138d7243009e54c0103a042c761a1ab83f1920/ChangeLog-9.1.md


greg0ire:
I think you should target a lower branch: 1.8 if this is considered a bugfix, 1.9 if this is considered a new feature. Not sure which it should be.

rybakit:
In 1.9, PHPUnit annotations are defined in `AnnotationReader.php`, and there are only three of them. I'm not sure it makes sense to add 2 more to this file if most of them are missing anyway: https://github.com/doctrine/annotations/blob/1.9/lib/Doctrine/Common/Annotations/AnnotationReader.php#L94.

greg0ire:
Yeah but look at https://github.com/doctrine/annotations/pull/242#issuecomment-467469910 … it seems that commit should have been cherry-picked. @Ocramius can you maybe help with this?

alcaeus:
In 1.x, we only ignore the code coverage annotations for PHPUnit, since the assumption is that people won't be reading annotations in their tests folder, where other PHPUnit annotations are commonly found.

As for the branch to target, master was a 2.0.x development branch that has stalled completely. Previous work has been backed up (see https://github.com/alcaeus/annotations/tree/previous-master) and will be revisited once we're ready to start work on 2.0 again. Since the previous master branch contained a number of severe BC breaks that were not covered by a compatibility layer or announced in 1.x releases, this was necessary to ensure a consistent workflow.

If you think that these exceptions should be ignored generally, I invite you to create a new pull request to the master branch. Thanks!

PR #323, opened 2020-04-09
Fix broken autoloading due to Reader calling registerFile

alcaeus:
As pointed out by @lcobucci, since the annotation reader invokes `AnnotationRegistry::registerFile`, we automatically disable the autoload mechanism. This PR fixes this behaviour by exchanging calls to `registerFile` by `class_exists`, invoking the standard autoloader we always assume to be present.


lcobucci:
Looks like a sane approach. Just think that this seems would be a good opportunity to create a functional test for the reader, though.

stof:
@alcaeus Please either write the functional test (if you plan to do it) or merge it and release it

alcaeus:
Will create a test when I have more time on my hands. If anybody wants to contribute this, please do so.

PR #329, opened 2020-05-04
Update DocParser.php

limingxinleo:
When we use the class name of the relative path, it will automatically require the php file. 
examples:
namespace App;
@Inject(Demo::class)

class_exists(Demo::class, false) will return true.

When we use the full path class name, it will not.
examples:
namespace App;
@Inject("App\Demo")

class_exists(Demo::class, false) will return false.

I think it's up to the user to decide whether the file is loaded or not.


stof:
This change won't help at all if you don't use deprecated APIs of the AnnotationRegistry, as the modern behavior of the library is to use `class_exists` in that layer too...

greg0ire:
Closing as per @stof 's comment

PR #330, opened 2020-05-05
Use ::class constants instead of strings to ease future refactorings

jkufner:
The use of ::class constants will make future refactorings easier and enables IDE to be more useful.

This change should have no impact on code bahavior.


alcaeus:
Thanks @jkufner!

PR #332, opened 2020-05-11
DocParser: Improve private typehints

jkufner:
- Improve private typehints and annotations; unify imports
- syntaxError() returns the exception rather than throwing it; it makes the code more readable and IDE less confused

None of these changes should change any behavior.

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -395,7 +395,7 @@ final class DocParser
     private function match(int $token): bool
     {
         if ( ! $this->lexer->isNextToken($token) ) {
-            $this->syntaxError($this->lexer->getLiteral($token));
+            throw $this->syntaxError($this->lexer->getLiteral($token));
         }
 
         return $this->lexer->moveNext();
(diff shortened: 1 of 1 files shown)


greg0ire:
Blocked by #341

greg0ire:
Thanks @jkufner !

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -358,10 +363,8 @@ public function parse($input, $context = '')
      * Finds the first valid annotation
      *
      * @param string $input The docblock string to parse
-     *
-     * @return int|null
      */
-    private function findInitialTokenPosition($input)
+    private function findInitialTokenPosition($input): ?int
```
stof, line 367: if you add the return type, why not adding the argument type too ?

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -421,12 +423,8 @@ private function matchAny(array $tokens)
      *
      * @param string     $expected Expected string.
      * @param array|null $token    Optional token.
```
stof, line 425: the descriptions don't really provide any value here, and the types are not specified through typehints

PR #334, opened 2020-05-25
Allow php 8

greg0ire:
The CI passes, the next logical step would be to get reports from
downstream users. This allows to remove the ignore-platform-reqs flag
when installing dependencies in the CI.
The nightly job is kept in "allow_failures", since we should not create
a sense of urgency about this.

I'm targeting 1.10.x so that this restriction can be lifted ASAP

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ php:
   - 7.2
   - 7.3
   - 7.4
-  - nightly
 
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


alcaeus:
@greg0ire The second commit brings us from this
> Your requirements could not be resolved to an installable set of packages.

To this

> ERRORS!
> Tests: 456, Assertions: 1050, Errors: 9, Skipped: 2.

I would recommend the following:
* merge this into the oldest stable branch (requires no release)
* work on fixing build in oldest stable branch
* once stable, require build to pass in weekly cron builds

We can apply the strategy to all our packages.

PR #337, opened 2020-06-07
Replace "blacklist" terminology with "ignore"

albe:
It's the terminology used in the code, so that just makes sense.


greg0ire:
I'm going to retarget this to the lowest maintained branch, and it will bubble up afterwards.

albe:
Ah yes - sorry, forgot to look closely where to target. Thanks!

PR #341, opened 2020-07-30
Make type in phpdoc resolvable

greg0ire:


greg0ire:
I'm labelling this as a patch since I fix phpdoc in src

PR #343, opened 2020-07-31
Allow using PHPUnit 9.3

greg0ire:
That is the only version that is compatible with PHP 8.
Lowering the minimum-stability on the nightly build is necessary until
9.3.0 is published.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ jobs:
       php: nightly
       before_install:
         - composer config platform.php 7.4.99
+        - composer config minimum-stability dev
 
     - stage: Lint
       before_script:


greg0ire:
@stof the failing test seems indeed to be the one I said:

https://github.com/doctrine/annotations/blob/9674cbc402d6ee39b7f59ee496ffb320100aa131/tests/Doctrine/Tests/Common/Annotations/PhpParserTest.php#L58-L70

greg0ire:
Reported at https://github.com/sebastianbergmann/phpunit/issues/4393

stof:
@greg0ire can you try to set the min stability to `dev` for the PHP nightly job, so that PHPUnit 9.3@dev gets used ?

greg0ire:
@stof it looks like that fixed the issue :)

stof:
I suggest merging the change allowing to use PHP 9.3@dev on nightly then (maybe adding a comment explaining why we change the min stability). This would unlock other PRs working on PHP 8 compat

stof:
I would not merge `--testdox` for the CI. When you don't have fatal errors (and so the PHPUnit process does not stop before the end of its reporting), I find the dot-based progress followed by the summary of errors to be much more readable in the CI logs than the testdox output where failures and passing tests are mixed together.

greg0ire:
Agree on all counts @stof I dropped the commit and reworded the PR

PR #344, opened 2020-07-31
Fix token parsing on latest PHP 8 version, fixes #339

### LINKED ISSUE #339: Aliasing breaks with PHP8
Error:
Uncaught Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The annotation "@ORM\Table" in class Shopware\Models\Shop\Shop was never imported.

```
use Doctrine\ORM\Mapping as ORM;
/**
 * @ORM\Table(name="s_core_shops")
 * @ORM\Entity(repositoryClass="Repository")
 */
```

Without alias it works:

```
use Doctrine\ORM\Mapping as ORM;
/**
 * \Doctrine\ORM\Mapping\Table(name="s_core_shops")
 * \Doctrine\ORM\Mapping\Entity(repositoryClass="Repository")
 */
```

shyim:
This fixes the failing tests  from TokenParser on PHP 8.

There are 10 left Failing tests. But it should be fixed by an PHPUnit Update

```
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

.........................................................E....E  63 / 456 ( 13%)
EEEEEEE........................................................ 126 / 456 ( 27%)
............................................................... 189 / 456 ( 41%)
............................................................... 252 / 456 ( 55%)
............................................................... 315 / 456 ( 69%)
......................................S........................ 378 / 456 ( 82%)
..................E.....................S...................... 441 / 456 ( 96%)
...............                                                 456 / 456 (100%)

Time: 812 ms, Memory: 10.00 MB

There were 10 errors:

1) Doctrine\Tests\Common\Annotations\AnnotationRegistryTest::testRegisterLoaderIfNotExistsOnlyRegisteresSameLoaderOnce
ParseError: syntax error, unexpected token "match", expecting variable

/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/AnnotationRegistryTest.php:183

2) Doctrine\Tests\Common\Annotations\CachedReaderTest::testIgnoresStaleCache
Method ReflectionParameter::getClass() is deprecated

/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:228
/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:22

3) Doctrine\Tests\Common\Annotations\CachedReaderTest::testIgnoresStaleCacheWithParentClass
Method ReflectionParameter::getClass() is deprecated

/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:228
/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:34

4) Doctrine\Tests\Common\Annotations\CachedReaderTest::testIgnoresStaleCacheWithTraits
Method ReflectionParameter::getClass() is deprecated

/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:228
/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:46

5) Doctrine\Tests\Common\Annotations\CachedReaderTest::testIgnoresStaleCacheWithTraitsThatUseOtherTraits
Method ReflectionParameter::getClass() is deprecated

/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:228
/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:60

6) Doctrine\Tests\Common\Annotations\CachedReaderTest::testIgnoresStaleCacheWithInterfacesThatExtendOtherInterfaces
Method ReflectionParameter::getClass() is deprecated

/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:228
/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:76

7) Doctrine\Tests\Common\Annotations\CachedReaderTest::testUsesFreshCacheWithTraitsThatUseOtherTraits
Method ReflectionParameter::getClass() is deprecated

/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:268
/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:93

8) Doctrine\Tests\Common\Annotations\CachedReaderTest::testPurgeLoadedAnnotations
Method ReflectionParameter::getClass() is deprecated

/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:228
/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:109

9) Doctrine\Tests\Common\Annotations\CachedReaderTest::testAvoidCallingFilemtimeTooMuch
Method ReflectionParameter::getClass() is deprecated

/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php:157

10) Doctrine\Tests\Common\Annotations\PhpParserTest::testClassFileDoesNotExist
Error: Call to undefined method ReflectionUnionType::getName()

/var/www/html/annotations/tests/Doctrine/Tests/Common/Annotations/PhpParserTest.php:63

ERRORS!
Tests: 456, Assertions: 1049, Errors: 10, Skipped: 2.
```

fixes #339

### DIFF
--- a/lib/Doctrine/Common/Annotations/TokenParser.php
+++ b/lib/Doctrine/Common/Annotations/TokenParser.php
@@ -99,7 +99,11 @@ class TokenParser
      */
     public function parseUseStatement()
     {
+        return \PHP_VERSION_ID >=80000 ? $this->parseUseStatementV8() : $this->parseUseStatementV7();
+    }
 
+    private function parseUseStatementV7(): array
+    {
         $groupRoot = '';
         $class = '';
         $alias = '';
@@ -115,6 +119,52 @@ class TokenParser
             } else if ($token[0] === T_AS) {
                 $explicitAlias = true;
                 $alias = '';
+            } else if ($token === ',') {
+                $statements[strtolower($alias)] = $groupRoot . $class;
+                $class = '';
+                $alias = '';
+                $explicitAlias = false;
+            } else if ($token === ';') {
+                $statements[strtolower($alias)] = $groupRoot . $class;
+                break;
+            } else if ($token === '{' ) {
+                $groupRoot = $class;
+                $class = '';
+            } else if ($token === '}' ) {
+                continue;
+            } else {
+                break;
+            }
+        }
+
+        return $statements;
+    }
+
+    private function parseUseStatementV8(): array
+    {
+        $groupRoot = '';
+        $class = '';
+        $alias = '';
+        $statements = [];
+        $explicitAlias = false;
+        while (($token = $this->next())) {
+            $isNameToken = $token[0] === T_STRING || $token[0] === T_NS_SEPARATOR;
+
+            if ($token[0] === T_NAME_QUALIFIED || $token[0] === T_NAME_FULLY_QUALIFIED) {
+                $class .= $token[1];
+
+                $classSplit = explode('\\', $token[1]);
+                $alias = $classSplit[count($classSplit) - 1];
+            } else if($token[0] === T_NS_SEPARATOR) {
+                $class .= '\\';
+            } else if ($token[0] === T_AS) {
+                $explicitAlias = true;
+                $alias = '';
+            } else if ($isNameToken && !$explicitAlias) {
+                $class = $token[1];
+                $alias = $token[1];
+            } else if ($isNameToken && $explicitAlias) {
+                $alias = $token[1];
             } else if ($token === ',') {
                 $statements[strtolower($alias)] = $groupRoot . $class;
                 $class = '';
@@ -172,7 +222,7 @@ class TokenParser
     public function parseNamespace()
     {
         $name = '';
-        while (($token = $this->next()) && ($token[0] === T_STRING || $token[0] === T_NS_SEPARATOR)) {
+        while (($token = $this->next()) && ($token[0] === T_STRING || $token[0] === T_NS_SEPARATOR || (\PHP_VERSION_ID >=80000 && $token[0] === T_NAME_QUALIFIED))) {
             $name .= $token[1];
         }
 


lib/Doctrine/Common/Annotations/TokenParser.php
```diff
@@ -82,7 +93,11 @@ public function next($docCommentIsComment = TRUE)
      */
     public function parseUseStatement()
     {
+        return PHP_VERSION_ID >=80000 ? $this->parseUseStatementV8() : $this->parseUseStatementV7();
```
stof, line 96: Please use a fully qualified name for the `PHP_VERSION_ID` constant, so that PHP can resolve that condition at compile time.

lib/Doctrine/Common/Annotations/TokenParser.php
```diff
@@ -48,6 +48,17 @@ public function __construct($contents)
         token_get_all("<?php\n/**\n *\n */");
 
         $this->numTokens = count($this->tokens);
+
+        // Compatibility defines for PHP < 8.0.
```
stof, line 52: Do we even need those now that we have separate methods ?

lib/Doctrine/Common/Annotations/TokenParser.php
```diff
@@ -119,6 +134,52 @@ public function parseUseStatement()
...
+        $explicitAlias = false;
+        while (($token = $this->next())) {
+            $isNameToken = $token[0] === T_STRING || $token[0] === T_NS_SEPARATOR;
+
+            if ($token[0] === T_NAME_QUALIFIED || $token[0] === T_NAME_FULLY_QUALIFIED) {
+                $class .= $token[1];
+
+                $classSplit = explode('\\', $token[1]);
+                $alias = $classSplit[count($classSplit) - 1];
+            } else if($token[0] === T_NS_SEPARATOR) {
```
stof, line 152: Do we still need all that handling in PHP 8 ? wouldn't use statements always be parsed as a single token and so no `T_NS_SEPARATOR` to deal with anymore ?

shyim: Its there only when you have a group

```php
use xxx/{Test, Foo}
```

lib/Doctrine/Common/Annotations/TokenParser.php
shyim: I did use it in https://github.com/doctrine/annotations/pull/344/files#diff-0eaa10dd1d548d163b15102395644457R213
I can split them also in 2 methods, if you want :)

shyim: Also mocking fails, when I remove it

stof: well, that other line of code could be changed to `(\PHP_VERSION_ID >=80000 && $token[0] === T_NAME_QUALIFIED)`, which will avoid having 2 methods and won't have any overhead (as the version check can be resolved at compile time)

shyim: Changed. Thanks :)

greg0ire:
Please target 1.10.x

jkufner:
More importantly, link that RFC from the code, so that when we return it a few years later, we know what happened.

greg0ire:
I'm attempting to address the `at()` warning issue at https://github.com/doctrine/annotations/pull/345

lib/Doctrine/Common/Annotations/TokenParser.php
```diff
@@ -172,7 +222,7 @@ public function parseUseStatements($namespaceName)
     public function parseNamespace()
     {
         $name = '';
-        while (($token = $this->next()) && ($token[0] === T_STRING || $token[0] === T_NS_SEPARATOR)) {
+        while (($token = $this->next()) && ($token[0] === T_STRING || $token[0] === T_NS_SEPARATOR || (\PHP_VERSION_ID >=80000 && $token[0] === T_NAME_QUALIFIED))) {
```
stof, line 225: isn't `T_NAME_FULLY_QUALIFIED` also allowed in a namespace declaration ?

stof:
@greg0ire I don't think we can easily avoid the copy-paste. The common code is the handling of group use statements (cases involving `{`, `,` and `}`) while the handling of other tokens (for the actual segments of a use statement) changed to the different structure of tokens in PHP 8. And we cannot extract the handling of group use statements to a separate method as these cases of the switch are altering the local state (it would of course be possible to extract things, but the result would probably be even harder to read)

stof:
For reference, here is the diff between both methods:

```diff
--- a/v7.php
+++ b/v8.php
@@ -3,7 +3,7 @@ class TokenParser
 {
     // Other code here. this is a file extract
 
-    private function parseUseStatementV7(): array
+    private function parseUseStatementV8(): array
     {
         $groupRoot = '';
         $class = '';
@@ -12,14 +12,22 @@ class TokenParser
         $explicitAlias = false;
         while (($token = $this->next())) {
             $isNameToken = $token[0] === T_STRING || $token[0] === T_NS_SEPARATOR;
-            if (!$explicitAlias && $isNameToken) {
+
+            if ($token[0] === T_NAME_QUALIFIED || $token[0] === T_NAME_FULLY_QUALIFIED) {
                 $class .= $token[1];
-                $alias = $token[1];
-            } else if ($explicitAlias && $isNameToken) {
-                $alias .= $token[1];
+
+                $classSplit = explode('\\', $token[1]);
+                $alias = $classSplit[count($classSplit) - 1];
+            } else if($token[0] === T_NS_SEPARATOR) {
+                $class .= '\\';
             } else if ($token[0] === T_AS) {
                 $explicitAlias = true;
                 $alias = '';
+            } else if ($isNameToken && !$explicitAlias) {
+                $class = $token[1];
+                $alias = $token[1];
+            } else if ($isNameToken && $explicitAlias) {
+                $alias = $token[1];
             } else if ($token === ',') {
                 $statements[strtolower($alias)] = $groupRoot . $class;
                 $class = '';
```

lib/Doctrine/Common/Annotations/TokenParser.php
```diff
@@ -136,6 +140,52 @@ public function parseUseStatement()
...
+                $class .= $token[1];
+
+                $classSplit = explode('\\', $token[1]);
+                $alias = $classSplit[count($classSplit) - 1];
+            } else if($token[0] === T_NS_SEPARATOR) {
+                $class .= '\\';
+            } else if ($token[0] === T_AS) {
+                $explicitAlias = true;
+                $alias = '';
+            } else if ($isNameToken && !$explicitAlias) {
```
stof, line 163: this should use `$token[0] === T_STRING` rather than `$isNameToken`, as the case of `$token[0] === T_NS_SEPARATOR` would never reach that case anyway.
Same for the next condition

stof:
btw, I think the case of `$explicitAlias` could be cleaned for the code for PHP 7 too: `T_NS_SEPARATOR` is not allowed in the alias.

jkufner:
I believe that this is one of the few rare cases when copy&paste is the right thing to do. However, it would be nice to add a few comments to mark the differences (and group them together). Also, the conditions should be as similar to the original as possible, e.g., keeping the order of `$isNameToken` and `$explicitAlias`. It should be easy to do the diff just by looking at the code. The guy reading it when PHP9 appears will thank you.

stof:
> The guy reading it when PHP9 appears will thank you.

At that time, I'm almost sure that the maintained version of doctrine/annotation will not support PHP 7 anyway so the duplicate code would be gone.
Thus, I doubt PHP 9 would change the token structure of use statements again...

greg0ire:
You're right, I didn't realize the diff was so important. A good commit message, or, as @jkufner pointed out, comments in the code would help.

stof:
@greg0ire as I said, cleaning the PHP 7 code could allow reducing it (for instance, we cannot have `T_NS_SEPARATOR` inside a explicit alias and we can only have one `T_STRING` so the branch of `$explicitAlias === true` should look the same in both versions if implemented cleanly in both)

stof:
I created a separate PR in #347 to keep a single method for both PHP 7 and PHP8.

stof:
I'm closing this PR in favor of #347

PR #345, opened 2020-08-01
Remove calls to TestCase::at()

greg0ire:
It has been deprecated as of PHPUnit 9
See https://github.com/sebastianbergmann/phpunit/issues/4297

Note that we still rely on the call order here, I am not the original
author of this test but the order looks like it is pretty important for
this test, or maybe at least for understand it.
Also note that we no longer check that save() is called after all calls
to fetch().

### DIFF
--- a/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php
@@ -157,53 +157,39 @@ class CachedReaderTest extends AbstractReaderTest
         /* @var $cache Cache|MockObject */
         $cache = $this->createMock('Doctrine\Common\Cache\Cache');
 
-        // first pass => cache ok for method 1
-        // we load annotations AND filemtimes for this file
         $cache
-            ->expects($this->at(0))
+            ->expects($this->exactly(6))
             ->method('fetch')
-            ->with($this->equalTo($cacheKeyMethod1))
-            ->will($this->returnValue([$route1])); // Result was cached, but there was an annotation;
-        $cache
-            ->expects($this->at(1))
-            ->method('fetch')
-            ->with($this->equalTo('[C]' . $cacheKeyMethod1))
-            ->will($this->returnValue($cacheTime));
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php
```diff
@@ -157,53 +157,39 @@ public function testAvoidCallingFilemtimeTooMuch() : void
...
-            ->will($this->returnValue([$route1])); // Result was cached, but there was an annotation;
-        $cache
-            ->expects($this->at(1))
-            ->method('fetch')
-            ->with($this->equalTo('[C]' . $cacheKeyMethod1))
-            ->will($this->returnValue($cacheTime));
+            ->withConsecutive(
+                // first pass => cache ok for method 1
+                // we load annotations AND filemtimes for this file
+                [$this->equalTo($cacheKeyMethod1)],
```
greg0ire, line 166: Does somebody know the difference between using this and using `[$cacheKeyMethod1]`?

greg0ire:
>  the order looks like it is pretty important for this test, or maybe at least for understand it.

I noticed transient tests that have to do with `fetch` before making this PR, and I just noticed another one, so maybe we should in fact get completely rid of the order? :thinking:

greg0ire:
Cc @bastnic in case you have ideas

greg0ire:
> What if we use a functional test about the fetch -> save relation?

Maybe that would help. What would that look like?

SenseException:
> Maybe that would help. What would that look like?

It could be a test that is using a real cache implementation like a custom one or maybe the `ArrayCache` and then check afterwards the expected content in the `Cache` class. I haven't checked if there are already tests like that. A custom test double implementing `Cache` could contain the checks for order, but I'm more in favour for already existing classes.

greg0ire:
Well… I restarted the build several times tonight and could not reproduce the issue, so I guess it was more transient than I thought, meaning it's probably not related to my changes. @SenseException , I made the changes you requested, please review again.

PR #347, opened 2020-08-10
Add support for the new PHP 8 tokens for use statements

### LINKED ISSUE #339: Aliasing breaks with PHP8

stof:
Fixes #339
Replaces #344

This PR achieves the same goal than #344 but without forking the whole method. This is done by first refactoring the existing code to differentiate between `T_STRING` and `T_NS_SEPARATOR` as they actually have different needs (`T_NS_SEPARATOR` won't appear in a explicit alias for instance) and then adding the support for the new PHP 8 tokens in this cleaned code.
For the second commit, I marked @shyim as a co-author to credit them for the good work done in #344 to make the feature work.

### DIFF
--- a/lib/Doctrine/Common/Annotations/TokenParser.php
+++ b/lib/Doctrine/Common/Annotations/TokenParser.php
@@ -106,12 +106,14 @@ class TokenParser
         $statements = [];
         $explicitAlias = false;
         while (($token = $this->next())) {
-            $isNameToken = $token[0] === T_STRING || $token[0] === T_NS_SEPARATOR;
-            if (!$explicitAlias && $isNameToken) {
+            if (!$explicitAlias && $token[0] === T_STRING) {
                 $class .= $token[1];
                 $alias = $token[1];
-            } else if ($explicitAlias && $isNameToken) {
-                $alias .= $token[1];
+            } else if ($explicitAlias && $token[0] === T_STRING) {
+                $alias = $token[1];
+            } else if ($token[0] === T_NS_SEPARATOR) {
+                $class .= '\\';
+                $alias = '';
             } else if ($token[0] === T_AS) {
                 $explicitAlias = true;
                 $alias = '';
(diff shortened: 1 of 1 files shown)


greg0ire:
Triggering a new build now that #345 was merged

lib/Doctrine/Common/Annotations/TokenParser.php
```diff
@@ -172,7 +179,7 @@ public function parseUseStatements($namespaceName)
     public function parseNamespace()
     {
         $name = '';
-        while (($token = $this->next()) && ($token[0] === T_STRING || $token[0] === T_NS_SEPARATOR)) {
+        while (($token = $this->next()) && ($token[0] === T_STRING || $token[0] === T_NS_SEPARATOR || (\PHP_VERSION_ID >= 80000 && ($token[0] === T_NAME_QUALIFIED || $token[0] === T_NAME_FULLY_QUALIFIED)))) {
```
greg0ire, line 182: A linebreak would be welcome here, it will mean less work if we want to enforce use our own CS one day (not sure why we don't yet)

stof: what would be the expected formatting ?

greg0ire: Several of them are valid I think. I would go with
```suggestion
        while (($token = $this->next()) && ($token[0] === T_STRING || $token[0] === T_NS_SEPARATOR || (
            \PHP_VERSION_ID >= 80000 &&
            ($token[0] === T_NAME_QUALIFIED || $token[0] === T_NAME_FULLY_QUALIFIED)
        ))) {
```

stof: done

greg0ire:
Thanks @stof @shyim !

PR #350, opened 2020-08-11
Abide by the automated part of our own coding standard

greg0ire:
This is the first step of the plan described at https://github.com/doctrine/annotations/issues/349

PR #352, opened 2020-08-12
Remove unneeded rule exclusion

greg0ire:
It was added before fixing another cs issue (usage of else if) that is
not supported by phpcs and causes an unwanted code change.

See https://github.com/slevomat/coding-standard/issues/1067


SenseException:
Is this PR a successor of #350 and you had issues with that sniff? Of do you want to remove it because it has the mentioned issues in the linked PR?

greg0ire:
It is the successor, and I want to enable the rule again because it's only buggy if you use else if, which we forbid with a rule I applied later on

PR #355, opened 2020-08-15
Manual cs fixes

greg0ire:
I am going to add more fixes, but would love reviews in case I made bad choices on the way

EDIT: I'm done! Please review.

### DIFF
--- a/lib/Doctrine/Common/Annotations/Annotation/Target.php
+++ b/lib/Doctrine/Common/Annotations/Annotation/Target.php
@@ -45,7 +45,7 @@ final class Target
     public const TARGET_ANNOTATION = 8;
     public const TARGET_ALL        = 15;
 
-    /** @var array */
+    /** @var array<string, int> */
     private static $map = [
         'ALL'        => self::TARGET_ALL,
         'CLASS'      => self::TARGET_CLASS,
@@ -54,7 +54,7 @@ final class Target
         'ANNOTATION' => self::TARGET_ANNOTATION,
     ];
 
-    /** @var array */
+    /** @var list<string> */
     public $value;
 
     /**
@@ -72,7 +72,7 @@ final class Target
     public $literal;
 
     /**
-     * @param array $values
+     * @param array{value?: string|list<string>} $values
      *
      * @throws InvalidArgumentException
      */
--- a/lib/Doctrine/Common/Annotations/Annotation/Target.php
+++ b/lib/Doctrine/Common/Annotations/Annotation/Target.php
@@ -54,7 +54,7 @@ final class Target
         'ANNOTATION' => self::TARGET_ANNOTATION,
     ];
 
-    /** @var list<string> */
+    /** @phpstan-var list<string> */
     public $value;
 
     /**
@@ -72,9 +72,9 @@ final class Target
     public $literal;
 
     /**
-     * @param array{value?: string|list<string>} $values
-     *
      * @throws InvalidArgumentException
+     *
+     * @phpstan-param array{value?: string|list<string>} $values
      */
     public function __construct(array $values)
     {
--- a/lib/Doctrine/Common/Annotations/CachedReader.php
+++ b/lib/Doctrine/Common/Annotations/CachedReader.php
@@ -45,7 +45,7 @@ final class CachedReader implements Reader
     /** @var bool */
     private $debug;
 
-    /** @var array */
+    /** @var array<string, array<object>> */
     private $loadedAnnotations = [];
 
     /** @var int[] */
@@ -72,7 +72,8 @@ final class CachedReader implements Reader
             return $this->loadedAnnotations[$cacheKey];
         }
 
-        if (($annots = $this->fetchFromCache($cacheKey, $class)) === false) {
+        $annots = $this->fetchFromCache($cacheKey, $class);
+        if ($annots === false) {
             $annots = $this->delegate->getClassAnnotations($class);
             $this->saveToCache($cacheKey, $annots);
         }
@@ -106,7 +107,8 @@ final class CachedReader implements Reader
             return $this->loadedAnnotations[$cacheKey];
         }
 
-        if (($annots = $this->fetchFromCache($cacheKey, $class)) === false) {
+        $annots = $this->fetchFromCache($cacheKey, $class);
+        if ($annots === false) {
             $annots = $this->delegate->getPropertyAnnotations($property);
             $this->saveToCache($cacheKey, $annots);
         }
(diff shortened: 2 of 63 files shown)


lib/Doctrine/Common/Annotations/Annotation/Target.php
```diff
@@ -54,7 +54,7 @@ final class Target
         'ANNOTATION' => self::TARGET_ANNOTATION,
     ];
 
-    /** @var array */
+    /** @var list<string> */
```
greg0ire, line 57: I tried something more advanced here by using `list<'ALL'|'CLASS'|'METHOD'|'PROPERTY'|'ANNOTATION'>` but phpcs then said `Property \Doctrine\Common\Annotations\Annotation\Target::$value does not have @var annotation for its value.`. I could add it as a prefixed annotation, but I am torn: we use `psalm` annotations in other packages, but we don't use Psalm in this package, so maybe I should prefix them with `phpstan`? A `psalm` prefix would work for Phpstan, it recognizes it.

lib/Doctrine/Common/Annotations/Annotation/Target.php
```diff
@@ -72,7 +72,7 @@ final class Target
     public $literal;
 
     /**
-     * @param array $values
+     * @param array{value?: string|list<string>} $values
```
greg0ire, line 75: I tried advanced syntax here but then got a warning that the line was too long and couldn't find a multiline syntax that would not make phpcs crash with `An error occurred during processing; checking has been aborted. The error message was: Uninitialized string offset: 0 in
    |       |     /home/greg/dev/doctrine-annotations/vendor/squizlabs/php_codesniffer/src/Standards/Squiz/Sniffs/Commenting/FunctionCommentSniff.php on line 381`
or complain that `Missing parameter name (Squiz.Commenting.FunctionComment.MissingParamName)`

lib/Doctrine/Common/Annotations/Annotation/Target.php
SenseException: Yeah, I'm torn by this in projects of mine too, but I guess we have to use the annotations of the tool we use.

greg0ire: Ok, I will do a final pass to prefix those with `phpstan` when I'm done.

greg0ire: @SenseException do you know in which case we should use a prefix, and in which case we shouldn't? `list`? `array<>`? `array{}`? All three? None of them?

ostrolucky: list and array{} at least IMHO

greg0ire: Ok let's start with those 👍

greg0ire: @ostrolucky addressed

greg0ire:
Ping @SenseException @ostrolucky please review

lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -243,8 +245,12 @@ private function getLastModification(ReflectionClass $class)
 
         $lastModification =  max(array_merge(
             [$filename ? filemtime($filename) : 0],
-            array_map([$this, 'getTraitLastModificationTime'], $class->getTraits()),
-            array_map([$this, 'getLastModification'], $class->getInterfaces()),
```
ostrolucky, line 247: What made us do this kind of change? It's missing from list provided at https://github.com/doctrine/annotations/pull/350#issuecomment-672222095. I don't like this unnecessary verbose syntax, what was wrong about previous one?

greg0ire: It is in the list:

>  259 | ERROR | Class CachedReader contains unused private method getTraitLastModificationTime(). (SlevomatCodingStandard.Classes.UnusedPrivateElements.UnusedMethod)

It's a bit verbose indeed, but [this thread](https://github.com/slevomat/coding-standard/issues/159) convinced me to go ahead.

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
```diff
@@ -501,38 +602,56 @@ class TestIgnoresNonAnnotationsClass
 
 class TestTopLevelAnnotationClass
 {
-    /** @\TopLevelAnnotation */
-    private $field;
+    /**
+     * @var mixed
```
SenseException, line 606: Aren't these `@var` annotations testing the ignored annotations again? In case someone breaks the ignored-annotations functionality, wouldn't all the additions you've added break more tests in such a case than necessary?

SenseException (changes requested):

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -1214,6 +1311,8 @@ private function Arrayx(): array
      *
      * @throws AnnotationException
      * @throws ReflectionException
+     *
+     * @phpstan-return array{mixed, mixed}
```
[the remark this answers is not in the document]
greg0ire: Maybe, but my goal here isn't to infer types with 100% accuracy. I saw `$key = $this->Constant()`, and had a look at `Constant`, which is annotated with `@return mixed`. So if I change this, I guess I will also have to change `Constant`, and this feels like a rabbit hole for another PR, do you agree?

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
greg0ire: I guess they are, and would, but you have to choose between that and ending up with a lot more ignore rules in the phpcs config files, so I figured we would be better off that way than potentially ignoring a lot more violations that we mean to. Getting properly-annotated properties in actual test classes (as opposed to fixtures) helps static analysis a lot, and can avoid a lot of mistakes.

SenseException: This is a "pick your poison" situation. Considering that we won't check future PRs which will introduce new classes with the need to extend the phpcs ignore rules, the lesser evil would be keeping the annotations.

SenseException (approved):

ostrolucky (approved):

lib/Doctrine/Common/Annotations/Annotation/Target.php
SenseException: I use `array{}` only for specific structure, so I think your choices are fine.

tests/Doctrine/Tests/Common/Annotations/AbstractReaderTest.php
greg0ire: > This is a "pick your poison" situation.

:100: Wouldn't have phrased it better!

greg0ire:
I'm preparing the merge up

PR #366, opened 2020-09-28
Add PHP 8.0 to CI

beberlei:


stof:
IMO, there is no reason to check coding standards on 2 PHP versions. You are not impacting the CI running tests here

beberlei:
Yeah just playing around :)

PR #367, opened 2020-09-29
Ignore additional generics annotations

sanmai:
Including for [Phan](https://github.com/phan/phan), [Psalm](https://github.com/vimeo/psalm), and [Infection](https://github.com/infection/infection).

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -137,6 +137,10 @@ class AnnotationReader implements Reader
         'implements' => true,
         'extends' => true,
         'use' => true,
+        // Phan, Psalm, Infection
+        'phan' => true,
+        'psalm' => true,
+        'infection' => true,
     ];
 
     /**


lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -137,6 +137,10 @@ class AnnotationReader implements Reader
         'implements' => true,
         'extends' => true,
         'use' => true,
+        // Phan, Psalm, Infection
+        'phan' => true,
+        'psalm' => true,
+        'infection' => true,
```
greg0ire, line 143: Sorry, what are we ignoring exactly here? Do this ignore rule ignore annotations prefixed with `psalm`? I'm asking because I'm familiar with `@psalm-*`, but not with `@psalm` if that's a thing.

sanmai: No, it is exactly about `@psalm-*` thing. But looking at  #298 it might be not required any more.

Can you confirm that dashed annotations like `@psalm-return` are automatically ignored?

stof: yes, dashed annotations are automatically ignored, as they cannot be valid doctrine annotations anyway (as dashes are forbidden in class names)

sanmai: Thank you for the confirmation. Then this PR is indeed unnecessary.

PR #368, opened 2020-10-02
Allow usage of self:: accessor for constants

### LINKED ISSUE #269: Referencing self:: constants in annotations
At the moment constant parsing doesn't read `self::` constants and I have a use case where that would be the most practical solution:
I would like to pass serialization groups for a property that's defined in a trait. Since not all entities will have the same serialization groups for the properties in the traits, the only currently available options that I see is to either repeat the property definition in each entity or repeat the groups serialization definition for each property and each entity in yaml/xml.

I would have created a PR instead of an issue for this, but from what I see in the DocParser class, the only way to get the original class name would be to parse the context string, which seems like a bad idea. I'm hoping that someone more familiar with the codebase could do it in a cleaner way. If there isn't a better way, please let me know and I'll make a PR.

bertrandseurot:
ex: `@Annotation(self::VALUE)`
Fixes #269

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -331,7 +331,10 @@ class AnnotationReader implements Reader
         $this->imports[$name] = array_merge(
             self::$globalImports,
             $this->phpParser->parseClass($class),
-            ['__NAMESPACE__' => $class->getNamespaceName()]
+            [
+                '__NAMESPACE__' => $class->getNamespaceName(),
+                'self' => $name,
+            ]
         );
 
         $this->ignoredAnnotationNames[$name] = $ignoredAnnotationNames;
--- a/tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
@@ -78,6 +78,67 @@ class AnnotationReaderTest extends AbstractReaderTest
         self::assertEquals([], $annotations);
     }
 
+    public function testClassAnnotationSupportsSelfAccessorForConstants(): void
+    {
+        $reader = $this->getReader();
+        $ref    = new ReflectionClass(Fixtures\ClassWithAnnotationWithSelfConstantReference::class);
+
+        $annotations = $reader->getClassAnnotations($ref);
+
+        self::assertCount(1, $annotations);
+
+        $annotation = $annotations[0];
+        self::assertInstanceOf(Fixtures\AnnotationWithConstants::class, $annotation);
+        self::assertEquals(
+            $annotation->value,
+            Fixtures\ClassWithAnnotationWithSelfConstantReference::VALUE_FOR_CLASS
+        );
+    }
+
+    public function testPropertyAnnotationSupportsSelfAccessorForConstants(): void
+    {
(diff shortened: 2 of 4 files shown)


stof:
I think `self` should be forbidden when parsing an annotation on a trait, as `self::` would resolve to the class using the trait in PHP, and we don't know that in the current code.

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -1005,6 +1005,20 @@ public function testSupportClassConstants(string $docblock, $expected): void
         self::assertEquals($expected, $annotation->value);
     }
 
+    public function testSupportSelfAccessorForClassConstants(): void
```
stof, line 1008: this test is not actually testing your patch. It copies your (current) patch in a test. I could revert the AnnotationReader change and still have a green test.

As this feature is implemented at the AnnotationReader level, tis must be tested at that level too.

stof (dismissed):

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
bertrandseurot: You're right, I added some test.

stof: please also remove this test as it does not actually test the feature but a copy of it

bertrandseurot: Done, thank you for your time.

bertrandseurot:
> 
> 
> I think `self` should be forbidden when parsing an annotation on a trait, as `self::` would resolve to the class using the trait in PHP, and we don't know that in the current code.

Seems to work in traits too. Am I missing something ?

stof:
@bertrandseurot if you use `self::` in the code of a trait, it refers to the class using the trait, not to the trait itself, as trait are basically compiler-based copy/paste. See https://3v4l.org/BSPC3 for the proof.

So supporting `self::` in annotations on traits would require that it works the same. If that's the case, this is fine.

bertrandseurot:
@stof This will fails only if you try to directly parse a trait out of a class context : 

```php
$reflectionTrait = new \ReflectionClass( RandomTrait::class );
$reader = new AnnotationReader();
$reader->getMethodAnnotations( reflectionTrait->getMethod( 'randomMethod')  );
```

It coul'd be annoying if you wanted to do that in order to get the traits annotation cached, and use this cache later when parsing classes using this trait. But since the php Reflection API does not not allow to know if a property/method of a trait is overriden in the class, I think this strategy is currently possible.

I confirm that it works in other cases (tested in `testMethodAnnotationSupportsSelfAccessorForConstants()` and `testPropertyAnnotationSupportsSelfAccessorForConstants`)

bertrandseurot:
Because of a force-push, this PR seems stuck in a "changes requested" status. Sorry for that, it's my first PR. Should I open another or it is not a problem ?

greg0ire:
It's not a problem and it's not because of a force-push. You need to get another review from @stof , or to get somebody from the team to dismiss his review.

bertrandseurot:
OK, thank you !

beberlei (approved):
This looks good and brings annotations in line with how PHP 8 attributes work.

PR #369, opened 2020-10-11
Add compatibility with named parameters and constructor property promotion

beberlei:
With PHP 8 the combination of named parameters and constructor property
promotion will be very cool for the new Attributes feature, but also for
Doctrine's Annotation feature. Especially if classes are to be used
for both Attributes and Annotations, then the DocParser needs
a compatible support for named parameters in annotation class constructors.

For backwards compatibility this is done with a marker interface that
uses a different instantiation logic than the current one. Currently
all annotations properties are passed as a single array argument
$values.

This patch uses reflection to match annotation values to class
constructor parameters. For PHP 8 the integration of named arguments
with spread operator is used.

A PHP 8 class that could be read using Doctrine Annotations would
become:

```
/** @Annotation */
class Table implements NamedArgumentConstructorAnnotation
{
    public function __construct(
        public string $name,
        public array $indexes = []
    ) {}
}
/** @Table(name="users") */
/** @Table(name="users", indexes={"foo"}) */
```

### DIFF
--- a/docs/en/custom.rst
+++ b/docs/en/custom.rst
@@ -53,6 +53,52 @@ values into public properties directly:
         public $bar;
     }
 
+Optional: Constructors with Named Parameters
+--------------------------------------------
+
+Starting with Annotations v1.11 a new annotation instantiation strategy
+is available that aims at compatibility of Annotation classes with the PHP 8
+attribute feature.
+
+You can implement the
+``Doctrine\Common\Annotations\NamedArgumentConstructorAnnotation`` interface
+and then declare a constructor with regular parameter names that are matched
+from the named arguments in the annotation syntax.
+
+.. code-block:: php
+
+    namespace MyCompany\Annotations;
+
+    use Doctrine\Common\Annotations\NamedArgumentConstructorAnnotation;
+
+    /** @Annotation */
+    class Bar implements NamedArgumentConstructorAnnotation
+    {
+        private $foo;
+
+        public function __construct(string $foo)
+        {
+            $this->foo = $foo;
+        }
+    }
+
+    /** Useable with @Bar(foo="baz") */
+
+In combination with PHP 8s constructor property promotion feature
+you can simplify this to:
+
+.. code-block:: php
+
+    namespace MyCompany\Annotations;
+
+    use Doctrine\Common\Annotations\NamedArgumentConstructorAnnotation;
+
+    /** @Annotation */
+    class Bar implements NamedArgumentConstructorAnnotation
+    {
+        public function __construct(private string $foo) {}
+    }
+
 Annotation Target
 -----------------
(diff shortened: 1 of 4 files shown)


greg0ire (approved):
I think this should target `master`, since it's not a bugfix.
Also, should docs be added regarding the new feature and how to migrate to it?

greg0ire (commented):
I think this should target `master`, since it's not a bugfix.
Also, should docs be added regarding the new feature and how to migrate to it?

lib/Doctrine/Common/Annotations/NamedArgumentConstructorAnnotation.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Doctrine\Common\Annotations;
+
+/**
+ * Marker interface for PHP7/PHP8 compatible support
+ * for named arguments (and constructor property promotion).
```
greg0ire, line 7: Should constructor property promotion really be mentioned here? You did not write a test for it, so I believe this means it does not really come into play here: it will work the same regardless of whether you use that feature or not.

beberlei:
@greg0ire right, rebased on master. I will document this as soon as we decide that this makes sense.

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -1537,6 +1538,55 @@ public function testWillParseAnnotationSucceededByANonImmediateDash(): void
         self::assertCount(1, $result);
         self::assertInstanceOf(SomeAnnotationClassNameWithoutConstructorAndProperties::class, $result[0]);
     }
+
+    public function testNamedArgumentsConstructorAnnotation(): void
+    {
+        $result = $this
+            ->createTestParser()
+            ->parse('/** @NamedAnnotation(bar=2222, foo="baz") */');
```
SenseException, line 1546: Additionally I would also expect that another test guarantees that a different order of the parameters isn't breaking the build for this feature. `/** @NamedAnnotation(foo="baz", bar=2222) */`

SenseException (changes requested):

docs/en/custom.rst
```diff
@@ -53,6 +53,52 @@ values into public properties directly:
...
+
+        public function __construct(string $foo)
+        {
+            $this->foo = $foo;
+        }
+    }
+
+    /** Useable with @Bar(foo="baz") */
+
+In combination with PHP 8s constructor property promotion feature
```
greg0ire, line 87:
```suggestion
In combination with PHP 8's constructor property promotion feature
```

greg0ire (approved):

SenseException (approved):

PR #372, opened 2020-10-17
GitHub actions

beberlei:
This replaces Travis CI with Github Actions to run the PHPUnit Testsuite for Doctrine Annotations.


greg0ire:
The current travis file is based on `composer update`, so you will probably have to account for that like I did in `doctrine/persistence`

.travis.yml
```diff
@@ -1,44 +0,0 @@
...
-    - stage: Test
-      php: nightly
-      before_install:
-        - composer config platform.php 7.4.99
-
-    - stage: Lint
-      before_script:
-        - travis_retry composer require --dev --prefer-dist --prefer-stable phpstan/phpstan:^0.12.20
-      script: vendor/bin/phpstan analyse -l 3 -c phpstan.neon lib tests
-      php: 7.4
```
greg0ire, line 36: This one is missing, you can copy this file (and remove the Psalm part for now): https://github.com/doctrine/.github/blob/main/workflow-templates/static-analysis.yml

.travis.yml
```diff
@@ -1,44 +0,0 @@
...
-
-    - stage: Benchmark
-      install:
-        - travis_retry composer update --prefer-dist
-        - curl -o phpbench.phar https://phpbench.github.io/phpbench/phpbench.phar
-        - curl -o phpbench.phar.pubkey https://phpbench.github.io/phpbench/phpbench.phar.pubkey
-      script: php phpbench.phar run
-      php: 7.4
```
greg0ire, line 44: That part is missing too

beberlei: Its not really asserting anything or? I don't think its necessary to run this in the CI.

beberlei:
@greg0ire huh, do you have a clue what the phpstan failure is? cant reproduce that locally. Its obviously not an error, because its about testing that these dont exist in the test. Adding a rule to ignore the error in phpstan.neon makes my local build fail, because the ignore rule was not matched.

lcobucci (approved):

.travis.yml
greg0ire: It's not performing assertions I think, but people can have a look at it and see if there were any regressions performance-wise.

greg0ire:
> @greg0ire huh, do you have a clue what the phpstan failure is? cant reproduce that locally. Its obviously not an error, because its about testing that these dont exist in the test. Adding a rule to ignore the error in phpstan.neon makes my local build fail, because the ignore rule was not matched.

I don't see a failure, I suppose this means you figured it out?

beberlei:
@greg0ire yes :)

beberlei:
@greg0ire https://github.com/doctrine/annotations/commit/372939358a083c484b350f2cdd9c197220eefc22

PR #380, opened 2020-11-22
Remove "--no-suggest" option from Composer commands

kubawerlos:
As it is deprecated in Composer 2:

> ```You are using the deprecated option "--no-suggest". It has no effect and will break in Composer 3.```


greg0ire (approved):
Can you please also contribute this to https://github.com/doctrine/.github ? :pray:

kubawerlos:
Done: https://github.com/doctrine/.github/pull/13

PR #382, opened 2020-11-22
Fix badges

kubawerlos:
Fixes:
 - replaced build badge for Travis with CI one for Github Actions
 - removed both versioneye badges as they are not working anymore
 - replaced badge for the stable to use shields.io one as it shows full tag name (including `v`)


greg0ire (approved):

greg0ire:
Thanks @kubawerlos !

PR #383, opened 2020-11-22
Fix link to Coding Standard Guidelines

kubawerlos:
As the old one was not working.


greg0ire (approved):
Regardless of whether latest is working or not, it makes more sense to link to the current branch than to the latest one, since that one is not necessarily released.

PR #391, opened 2021-01-01
Introduced annotation NamedArgumentConstructor

### LINKED ISSUE #379: Reconsider NamedArgumentConstructorAnnotation
I would like to use the named constructor arguments feature in Symfony for some of the annotations there. We already have altered the constructors of most annotation classes in order to enable them to be used as attributes with named arguments as well. By leveraging this new way of constructing annotation classes, I expect to simplify those constructors again.

However, in order to do so, I have to add this marker interface `NamedArgumentConstructorAnnotation` to my annotation class. That‘s kind of a first because so far, we can construct instances of our annotation classes, even if the `doctrine/annotations` library is not installed. This becomes even more important if one fine day we wanted to build an application that uses PHP Attributes instead of Doctrine annotations everywhere.

Because of that, I wonder if we could find an alternative way to flag a class for named argument construction. Maybe we could achieve this with a `@NamedArgumentConstructor` annotation that has the same effect as the marker interface.

derrabus:
Fixes #379

This PR enables annotations to provide a constructor to be called with named arguments without requiring to implement a marker interface.

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -145,14 +145,14 @@ final class DocParser
      */
     private static $annotationMetadata = [
         Annotation\Target::class => [
-            'is_annotation'         => true,
-            'has_constructor'       => true,
-            'has_named_constructor' => false,
-            'properties'            => [],
-            'targets_literal'       => 'ANNOTATION_CLASS',
-            'targets'               => Target::TARGET_CLASS,
-            'default_property'      => 'value',
-            'attribute_types'       => [
+            'is_annotation'                  => true,
+            'has_constructor'                => true,
+            'has_named_argument_constructor' => false,
+            'properties'                     => [],
+            'targets_literal'                => 'ANNOTATION_CLASS',
+            'targets'                        => Target::TARGET_CLASS,
+            'default_property'               => 'value',
+            'attribute_types'                => [
                 'value'  => [
                     'required'   => false,
                     'type'       => 'array',
@@ -162,18 +162,18 @@ final class DocParser
             ],
         ],
         Annotation\Attribute::class => [
-            'is_annotation'         => true,
-            'has_constructor'       => false,
-            'has_named_constructor' => false,
-            'targets_literal'       => 'ANNOTATION_ANNOTATION',
-            'targets'               => Target::TARGET_ANNOTATION,
-            'default_property'      => 'name',
-            'properties'            => [
+            'is_annotation'                  => true,
+            'has_constructor'                => false,
+            'has_named_argument_constructor' => false,
+            'targets_literal'                => 'ANNOTATION_ANNOTATION',
+            'targets'                        => Target::TARGET_ANNOTATION,
+            'default_property'               => 'name',
+            'properties'                     => [
                 'name'      => 'name',
                 'type'      => 'type',
                 'required'  => 'required',
             ],
-            'attribute_types'       => [
+            'attribute_types'                => [
                 'value'  => [
                     'required'  => true,
                     'type'      => 'string',
@@ -192,14 +192,14 @@ final class DocParser
             ],
         ],
         Annotation\Attributes::class => [
-            'is_annotation'         => true,
-            'has_constructor'       => false,
-            'has_named_constructor' => false,
-            'targets_literal'       => 'ANNOTATION_CLASS',
-            'targets'               => Target::TARGET_CLASS,
-            'default_property'      => 'value',
-            'properties'            => ['value' => 'value'],
-            'attribute_types'       => [
+            'is_annotation'                  => true,
+            'has_constructor'                => false,
+            'has_named_argument_constructor' => false,
+            'targets_literal'                => 'ANNOTATION_CLASS',
+            'targets'                        => Target::TARGET_CLASS,
+            'default_property'               => 'value',
+            'properties'                     => ['value' => 'value'],
+            'attribute_types'                => [
                 'value' => [
                     'type'      => 'array',
                     'required'  => true,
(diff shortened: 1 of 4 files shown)


greg0ire (commented):
Should this come with docs?

derrabus:
Of course. I've added a piece of documentation. 😃

beberlei:
Just realized we could simplify this by making the marker interface a boolean on `@Annotation` - what do you think?

derrabus:
I can give it a try. The problem that I see with that approach is that the `@Annotation` annotation is currently "loaded" via a `strpos()`, probably to avoid recursion.

derrabus:
Implemented the property as #394. Both approaches would work for me. I'll let you decide. :-)

greg0ire:
> The problem that I see with that approach is that the @Annotation annotation is currently "loaded" via a strpos(), probably to avoid recursion.

Sorry for the dumb questions, I am unfamiliar with the package… what do you mean by "loaded" via `strpos()`? Can you add a permalink to the relevant piece of code? :pray: 
Also, how does it relate to recursion, and where would that cause a problem?

derrabus:
> Sorry for the dumb questions, I am unfamiliar with the package…

So was I until I started working on this patch. 😅 

> what do you mean by "loaded" via `strpos()`? Can you add a permalink to the relevant piece of code? 🙏

Sure, here you go:

https://github.com/doctrine/annotations/blob/bc13c658cd1aba7d2ea751ecf25186cb22fa1e70/lib/Doctrine/Common/Annotations/DocParser.php#L514

> Also, how does it relate to recursion, and where would that cause a problem?

In order to enable a class to be used as annotation, you have to annotate it with the `@Annotation` annotation. Unlike other annotations, `@Annotation` is not implemented as an actual class. Otherwise you would end up with code like this:

```php
/**
 * @Annotation
 */
class Annotation {}
```

In other words: The `Annotation` class would be annotated with itself. This is a recursion: While loading `@Annotation`, you look up the class `Annotation` which is annotated with `@Annotation` represented by the class `Annotation` which is annotated with `@Annotation`…

In #394 I had to add the `Annotation` class so I could attach a property to it. It works, but I believe not having a class representation of `@Annotation` is more robust and easier to maintain.

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -145,33 +145,35 @@ final class DocParser
...
-            'is_annotation'    => true,
-            'has_constructor'  => true,
-            'properties'       => [],
-            'targets_literal'  => 'ANNOTATION_CLASS',
-            'targets'          => Target::TARGET_CLASS,
-            'default_property' => 'value',
-            'attribute_types'  => [
+            'is_annotation'         => true,
+            'has_constructor'       => true,
+            'has_named_constructor' => false,
```
greg0ire: This might be confusing: to me, a named constructor is a static method that returns `self`

greg0ire:
With this solution, what happens when we annotate a class with `NamedArgumentConstructor` but forget to annotate it with `Annotation` too? Is it something that could want to do deliberately?

beberlei:
I assume its the same problem with target annotation, it does nothing. Not something we need to care for anymore

greg0ire:
I have as slight preference for the second solution because it's better DX IMO, but I'm fine with both solutions. What's your opinion @beberlei , taking into account @derrabus 's concerns?

lib/Doctrine/Common/Annotations/DocParser.php
derrabus: You are right. I wanted to keep it short since it's only an internal representation, but I understand that this name causes confusion. I'll change it.

derrabus:
> but forget to annotate it with `Annotation` too?

The line of code I've linked earlier is the one that determines if a class is a valid annotation. If `@Annotation` is missing, `@NamedArgumentConstructor` won't do anything. As Benjamin said, we have this situaton with `@Target` and other annotations already.

beberlei:
I think this approach here is better then, i didn't realize `Annotation` was such a workaround to begin with.

Vincz:
Could we also introduce a way to indicate the `default_property` to target a property other than the property `value` with `NamedArgumentConstructor`? Unless it is already the case but I didn't understand how to do it?

Vincz:
```php
use Attribute;
use Doctrine\Common\Annotations\NamedArgumentConstructorAnnotation;

/**
 * @Annotation
 * @Target({"CLASS"})
 */
#[Attribute(Attribute::TARGET_CLASS)]
class Foo implements NamedArgumentConstructorAnnotation
{
    /**
     * The field name.
     *
     * @var string
     */
    public ?string $name;

    public function __construct(?string $name)
    {
        $this->name = $name;
    }
}
```
The supported syntax are:
```php
#[Foo("ok")]
class TestClass{}

#[Foo(name: "ok")]
class TestClass{}

/** @Foo(name="ok") */
class TestClass{}
```

but this one is not supported as it assume a `value` property.
```php
/** @Foo("not ok") */
class TestClass{}
```

We could add an attribute `defaultProperty` (default: `value`) on the `@Annotation` or on the new `@NamedArgumentConstructor`?
What do you think?

derrabus:
@Vincz Let me tinker with that idea a bit. We would need something like that for Symfony as well. But I think, we can solve this in a follow-up PR.

Vincz:
@derrabus It is already used internally by the `Doc Parser` for specific annotations like `Annotation\Attribute`:

```php
Annotation\Attribute::class => [
            'is_annotation'    => true,
            'has_constructor'  => false,
            'targets_literal'  => 'ANNOTATION_ANNOTATION',
            'targets'          => Target::TARGET_ANNOTATION,
            'default_property' => 'name',
           ...
]
```
As a workaround, we can add a property 'value' if it is not defined in the constructor and use it's value on another property but it doesn't work with required properties. I could submit a PR but I don't know where would be the best place to add this option and I wonder if it makes sense to have this feature outside the `NamedArgumentConstructor` context.

Vincz:
Hey guys. Any update about this? Is there anything I can do to help?

derrabus:
Sorry, I've been a bit busy this week. I hope I can work on this on the weekend.

beberlei:
This PR should not extend in scope for now, lets tackle this as an extra issue

Vincz:
Ok, I created a new issue here https://github.com/doctrine/annotations/issues/396 in order to handle this feature.

lib/Doctrine/Common/Annotations/DocParser.php
beberlei: @derrabus ok i think we will take this approach here, so that leaves renaming this option as the last todo if @greg0ire agrees. I would merge afterwards.

derrabus: Done.

greg0ire (approved):
LGTM, I would squash merge.

SenseException (approved):

greg0ire:
Thanks @derrabus !

TomasVotruba:
@derrabus This is game-changer! Thank you for this feature :clap: :bow: 

Fun fact: I always assumed annotations work this way by default :D

derrabus:
Thanks a lot. 😃

PR #392, opened 2021-01-01
Ignore .phpunit.result.cache

derrabus:
Running the test suite with PHPUnit 9.5 will create a `.phpunit.result.cache` file. I'd like to propose to ignore it.


greg0ire (approved):

greg0ire:
Thanks @derrabus !

Issue #396, opened 2021-01-09
Handle annotation default property


Vincz:
It would be great to be able to define the default property for annotation. At the moment, it always target the property named `value`. The ability to change this behaviour could lead to more explicit variables names and a better compatibility with PHP8 attributes. 

See the following example.

```php
use Attribute;
use Doctrine\Common\Annotations\NamedArgumentConstructorAnnotation;

/**
 * @Annotation
 * @Target({"CLASS"})
 */
#[Attribute(Attribute::TARGET_CLASS)]
class Foo implements NamedArgumentConstructorAnnotation
{
    /**
     * The field name.
     *
     * @var string
     */
    public ?string $name;

    public function __construct(?string $name)
    {
        $this->name = $name;
    }
}
```
The supported syntax are:
```php
#[Foo("ok")]
class TestClass{}

#[Foo(name: "ok")]
class TestClass{}

/** @Foo(name="ok") */
class TestClass{}
```

but this one is not supported as it expects a `value` property.
```php
/** @Foo("not ok") */
class TestClass{}
```

Note that this is already handled in the Doc Parser for annotation like `@Annotation\Attribute`, as the default property in this case is `name`.

```php
Annotation\Attribute::class => [
    'is_annotation'    => true,
    'has_constructor'  => false,
    'targets_literal'  => 'ANNOTATION_ANNOTATION',
    'targets'          => Target::TARGET_ANNOTATION,
    'default_property' => 'name',
     ...
]
```

We could support the attribute `default_property` in many ways:

1. Add it to the `@Annotation` annotation, like `@Annotation(defaultProperty="name")`. But as explained here https://github.com/doctrine/annotations/pull/391 , this annotation is kind of special as it is detected through `str_pos`. 
2. Add a new annotation `@DefaultProperty`, like `@DefaultPropety("name")` 
3. Add this attribute with `NamedArgumentConstructorAnnotation` only.

Another path would be to force the DocParser to respect argument constructor arguments order if there is a mix of named / not named arguments.

murtukov:
I stand for the second option, but with the name `@Default` similar to `@Required`.

```php
/**
 * @Annotation
 * @Target({"CLASS"})
 */
#[Attribute(Attribute::TARGET_CLASS)]
class Foo implements NamedArgumentConstructorAnnotation
{
    /**
     * The field name.
     *
     * @Default
     * @Required
     *
     * @var string
     */
    private string $name;

    public function __construct(string $name)
    {
[...]

PR #402, opened 2021-01-19
Default property with @NamedArgumentConstructor

Vincz:
Use the first constructor argument as default property when @NamedArgumentConstructor is used.

### DIFF
--- a/docs/en/custom.rst
+++ b/docs/en/custom.rst
@@ -58,13 +58,60 @@ Optional: Constructors with Named Parameters
 
 Starting with Annotations v1.11 a new annotation instantiation strategy
 is available that aims at compatibility of Annotation classes with the PHP 8
-attribute feature.
+attribute feature. You need to declare a constructor with regular parameter 
+names that match the named arguments in the annotation syntax.
 
-You can implement the
+To enable this feature, you can tag your annotation class with 
+``@NamedArgumentConstructor`` (available from v1.12) or implement the
 ``Doctrine\Common\Annotations\NamedArgumentConstructorAnnotation`` interface
-and then declare a constructor with regular parameter names that are matched
-from the named arguments in the annotation syntax.
+(available from v1.11 and deprecated as of v1.12).
+When using the ``@NamedArgumentConstructor`` tag, the first argument of the
+constructor is considered as the default one.
 
+
+Usage with the ``@NamedArgumentContrustor`` tag
+
+.. code-block:: php
+
+    namespace MyCompany\Annotations;
+
+    /** 
+     * @Annotation 
+     * @NamedArgumentConstructor
+     */
+    class Bar implements NamedArgumentConstructorAnnotation
+    {
+        private $foo;
+
+        public function __construct(string $foo)
+        {
+            $this->foo = $foo;
+        }
+    }
+
+    /** Usable with @Bar(foo="baz") */
+    /** Usable with @Bar("baz") */
+
+In combination with PHP 8's constructor property promotion feature
+you can simplify this to:
+
+.. code-block:: php
+
+    namespace MyCompany\Annotations;
+
+    /** 
+     * @Annotation 
+     * @NamedArgumentConstructor
+     */
+    class Bar implements NamedArgumentConstructorAnnotation
+    {
+        public function __construct(private string $foo) {}
+    }
+
+
+Usage with the 
+``Doctrine\Common\Annotations\NamedArgumentConstructorAnnotation``
+interface (v1.11, deprecated as of v1.12):
 .. code-block:: php
 
     namespace MyCompany\Annotations;
@@ -76,44 +123,10 @@ from the named arguments in the annotation syntax.
     {
         private $foo;
 
-        public function __construct(string $foo)
-        {
-            $this->foo = $foo;
-        }
-    }
-
-    /** Useable with @Bar(foo="baz") */
-
-In combination with PHP 8's constructor property promotion feature
-you can simplify this to:
-
-.. code-block:: php
-
-    namespace MyCompany\Annotations;
-
-    use Doctrine\Common\Annotations\NamedArgumentConstructorAnnotation;
-
-    /** @Annotation */
-    class Bar implements NamedArgumentConstructorAnnotation
-    {
         public function __construct(private string $foo) {}
     }
 
-Alternatively, you can annotate your annotation class with
-``@NamedArgumentConstructor`` in case you cannot use the marker interface.
-
-.. code-block:: php
-
-    namespace MyCompany\Annotations;
-
-    /**
-     * @Annotation
-     * @NamedArgumentConstructor
-     */
-    class Bar
-    {
-        public function __construct(private string $foo) {}
-    }
+    /** Usable with @Bar(foo="baz") */
 
 Annotation Target
 -----------------
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -601,7 +601,7 @@ final class DocParser
                         }
 
                         $metadata['enum'][$property->name]['value']   = $annotation->value;
-                        $metadata['enum'][$property->name]['literal'] = ( ! empty($annotation->literal))
+                        $metadata['enum'][$property->name]['literal'] = (! empty($annotation->literal))
                             ? $annotation->literal
                             : $annotation->value;
                     }
@@ -866,35 +866,7 @@ EXCEPTION
         }
 
         $arguments = $this->MethodCall();
-
-        $positionalArguments = $arguments['positional_arguments'] ?? [];
-        $namedArguments      = $arguments['named_arguments'] ?? [];
-
-        if (
-            self::$annotationMetadata[$name]['has_named_argument_constructor']
-            && self::$annotationMetadata[$name]['default_property'] !== null
-        ) {
-            $values = $namedArguments;
-            foreach (self::$annotationMetadata[$name]['constructor_args'] as $property => $parameter) {
-                $position = $parameter['position'];
-                if (isset($values[$property]) || ! isset($positionalArguments[$position])) {
-                    continue;
-                }
-
-                $values[$property] = $positionalArguments[$position];
-            }
-        } else {
-            $values = $namedArguments;
-            if (! empty($positionalArguments) && ! isset($values['value'])) {
-                if (count($positionalArguments) === 1) {
-                    $value = $positionalArguments[0];
-                } else {
-                    $value = array_values($positionalArguments);
-                }
-
-                $values['value'] = $value;
-            }
-        }
+        $values    = $this->resolvePositionalValues($arguments, $name);
 
         if (isset(self::$annotationMetadata[$name]['enum'])) {
             // checks all declared attributes
@@ -1132,9 +1104,9 @@ EXCEPTION
                 case ! empty($this->namespaces):
                     foreach ($this->namespaces as $ns) {
                         if (class_exists($ns . '\\' . $className) || interface_exists($ns . '\\' . $className)) {
-                             $className = $ns . '\\' . $className;
-                             $found     = true;
-                             break;
+                            $className = $ns . '\\' . $className;
+                            $found     = true;
+                            break;
                         }
                     }
 
@@ -1161,7 +1133,7 @@ EXCEPTION
             }
 
             if ($found) {
-                 $identifier = $className . '::' . $const;
+                $identifier = $className . '::' . $const;
             }
         }
 
@@ -1433,5 +1405,44 @@ EXCEPTION
         }
 
         return false;
+    }
+
+    /**
+     * Resolve positional arguments (without name) to named ones
+     *
+     * @param array<string,mixed> $arguments
+     *
+     * @return array<string,mixed>
+     */
+    private function resolvePositionalValues(array $arguments, string $name): array
+    {
+        $positionalArguments = $arguments['positional_arguments'] ?? [];
+        $values              = $arguments['named_arguments'] ?? [];
+
+        if (
+            self::$annotationMetadata[$name]['has_named_argument_constructor']
+            && self::$annotationMetadata[$name]['default_property'] !== null
+        ) {
+            foreach (self::$annotationMetadata[$name]['constructor_args'] as $property => $parameter) {
+                $position = $parameter['position'];
+                if (isset($values[$property]) || ! isset($positionalArguments[$position])) {
+                    continue;
+                }
+
+                $values[$property] = $positionalArguments[$position];
+            }
+        } else {
+            if (count($positionalArguments) > 0 && ! isset($values['value'])) {
+                if (count($positionalArguments) === 1) {
+                    $value = $positionalArguments[0];
+                } else {
+                    $value = array_values($positionalArguments);
+                }
+
+                $values['value'] = $value;
+            }
+        }
+
+        return $values;
     }
 }
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -1061,10 +1061,6 @@ EXCEPTION
             $token = $this->lexer->lookahead;
             $value = $this->Value();
 
-            if (! is_object($value) && ! is_array($value)) {
-                throw $this->syntaxError('Value', $token);
-            }
-
             $values[] = $value;
         }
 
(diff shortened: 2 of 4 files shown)


docs/en/custom.rst
```diff
@@ -100,7 +100,8 @@ you can simplify this to:
     }
 
 Alternatively, you can annotate your annotation class with
-``@NamedArgumentConstructor`` in case you cannot use the marker interface.
+``@NamedArgumentConstructor`` in case you cannot use the marker interface. 
+In this case, the first argument of the constructor will be considered as the default property.
```
stof: The way this is written makes me think that this works only when using the `@NamedArgumentConstructor` annotation but not when using the marker interface, as you added that in the paragraph about the alternative. Is it actually true ?

Vincz: It is true. It is the intended behaviour as we didn't want to introduce breaking change with current implementation of the interface.
See https://github.com/doctrine/annotations/issues/396

Vincz:
Hey guys :) Any feedback about this?
ping @greg0ire 
ping @beberlei 
ping @derrabus

greg0ire:
I see @derrabus mentioned that we should deprecate the interface, and I quite agree, I'd rather not have 2 ways of doing the same thing, with differences that are hard to grasp (the only comment on this PR so far is about exactly that BTW). Did you overlook that or is it planned for later?

Vincz:
@greg0ire Ok, the interface has been deprecated and the doc updated.

greg0ire:
Please kindly squash your commits together. Maybe not all, but some definitely needs to be squashed. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.

How to do that?

1. `git rebase -i origin/1.12.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already push. Don't forget the `--force` option otherwise git will try to merge both things together.

Vincz:
@greg0ire Done!

greg0ire (dismissed):

greg0ire:
ping @stof @beberlei :slightly_smiling_face:

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -859,7 +863,17 @@ private function Annotation()
             );
         }
 
-        $values = $this->MethodCall();
+        $defaultProperty = 'value';
```
beberlei: Instead of this code block and passing down `$defaultProperty` could you instead handle this logic in the original line 944 and following?

```php
 foreach ($values as $property => $value) {
    if ($property === 'value') {
        $property = self::$annotationMetadata[$name]['default_property'];
    }
```
That would be less invasive and keep the logic at fewer places. I am not 100% sure it could work this way though, just a rough idea.

beberlei (changes requested):
An idea to improve the patch a little to be less invasive.

lib/Doctrine/Common/Annotations/DocParser.php
Vincz: Hey @beberlei! Thx for the review. I didn't want to pass the `defaultProperty` down. The problem is, without it, we don't know with the return of `values()`, if `value` as been set because it's the default property... or because we have an argument named
`value`. So it would not work if for example the constructor looks like this:
```
public function __construct(string $name, int $value){}
```

beberlei: But isnt that already a problem? Or just for the new named constructors?

Vincz: I gave a bunch of examples here : https://github.com/doctrine/annotations/issues/396.
The fact that we can't know if `value` has been set because it's the default property or because we have an argument with the same name, is, indeed an "old problem".
But we didn't want to change the current behaviour to avoid BC breaks, so we decide to handle it properly only with the new `@NamedArgumentsConstructor` annotation as it is not published yet.

beberlei: The Problem is all methods with large letter starting are named after the rules in the grammer. None of them accept arguments, because they already operate on state of the class instead. Passing arguments now will create a huge inconsistency in the API.

What you could try instead is change the return value of `Values` to be an associative array with: `['named_arguments' => [], 'positional_arguments' => []]` and then modify `Annotation` to act accordingly.

Vincz: @beberlei Here is my first attempt trying what you suggested.
I was also wondering why this is allowed: `@Name(@Name, @Name)`, but this is not `@Name("1", "2")`. The second case is not allowed because of this line: https://github.com/doctrine/annotations/blob/b81688bc95a16b76b2849bb67004d8f34ddfffd7/lib/Doctrine/Common/Annotations/DocParser.php#L1037
In our case, it means that we can only use one positional argument. It would make sense for us to allow multiple positional arguments as it's also how attributes work. `#[Name("1", "2")]` would work for example as long as the related attribute constructor accept at least two arguments. 
Removing the above line don't break the tests and allow multiple positional arguments. 
What do you think about this?

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -859,7 +865,36 @@ private function Annotation()
...
+                $position = $parameter['position'];
+                if (isset($values[$property]) || ! isset($positionalArguments[$position])) {
+                    continue;
+                }
+
+                $values[$property] = $positionalArguments[$position];
+            }
+        } else {
+            $values = $namedArguments;
+            if (! empty($positionalArguments) && ! isset($values['value'])) {
```
beberlei: Please use count instead, empty does not communicate intent well

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -859,7 +865,36 @@ private function Annotation()
             );
         }
 
-        $values = $this->MethodCall();
+        $arguments = $this->MethodCall();
+
+        $positionalArguments = $arguments['positional_arguments'] ?? [];
+        $namedArguments      = $arguments['named_arguments'] ?? [];
+
+        if (
```
beberlei: Could you extract this block into a private helper please?


    $values = $this->assembleValues($arguments, $name);

Method name could be better

beberlei (changes requested):
Much better! Only some small things left from my pov, then good to merge

lib/Doctrine/Common/Annotations/DocParser.php
jkufner: Isn't `count() > 0` much less effective than `!empty()` ?

beberlei: Yes, but the output of this code is usually cached and even if not, this is not a tight loop

greg0ire: I wanted to toy with phpbench so I build a little benchmark from [this raw benchmark](https://stackoverflow.com/a/29635982/353612):

```php
<?php

class ArrayCheckerBench
{
    public function provideArrays(): \Generator
    {
        yield 'empty array' => ['array' => []];
        yield 'huge array' => ['array' => range(0, 100000)];
    }

    public function checkArrayWithEmpty(array $x): bool
    {
        return empty($x);
    }

    public function checkArrayWithCount(array $x): bool
    {
        return count($x) === 0;
    }

    /**
     * @Revs(1000000)
     * @Iterations(5)
     * @ParamProviders({"provideArrays"})
     */
    public function benchCheckIfArrayIsEmptyWithEmpty(array $parameters): void
    {
        $this->checkArrayWithEmpty($parameters['array']);
    }

    /**
     * @Revs(1000000)
     * @Iterations(5)
     * @ParamProviders({"provideArrays"})
     */
    public function benchCheckIfArrayIsEmptyWithCount(array $parameters): void
    {
        $this->checkArrayWithCount($parameters['array']);
    }
}
```

Here are the results:

```
vendor/bin/phpbench run ArrayCheckerBench.php --report=aggregate --retry-threshold=0.5
PhpBench @git_tag@. Running benchmarks.

\ArrayCheckerBench

    benchCheckIfArrayIsEmptyWithEmpty # emp.I4 [μ Mo]/r: 0.285 0.285 (μs) [μSD μRSD]/r: 0.000μs 0.06%
    benchCheckIfArrayIsEmptyWithEmpty # hug.R1 I1 [μ Mo]/r: 0.287 0.287 (μs) [μSD μRSD]/r: 0.000μs 0.12%
    benchCheckIfArrayIsEmptyWithCount # emp.R1 I1 [μ Mo]/r: 0.309 0.309 (μs) [μSD μRSD]/r: 0.001μs 0.28%
    benchCheckIfArrayIsEmptyWithCount # hug.R3 I3 [μ Mo]/r: 0.311 0.311 (μs) [μSD μRSD]/r: 0.001μs 0.34%

2 subjects, 20 iterations, 4,000,000 revs, 0 rejects, 0 failures, 0 warnings
(best [mean mode] worst) = 0.285 [0.298 0.298] 0.285 (μs)
⅀T: 5.958μs μSD/r 0.001μs μRSD/r: 0.201%
suite: 134621f1b80d1bb3f7ed3eaf6bc32c7bb639d7d1, date: 2021-02-07, stime: 15:08:37
+-------------------+-----------------------------------+-------------+---------+-----+-------------+---------+---------+---------+---------+---------+--------+-------+
| benchmark         | subject                           | set         | revs    | its | mem_peak    | best    | mean    | mode    | worst   | stdev   | rstdev | diff  |
+-------------------+-----------------------------------+-------------+---------+-----+-------------+---------+---------+---------+---------+---------+--------+-------+
| ArrayCheckerBench | benchCheckIfArrayIsEmptyWithEmpty | empty array | 1000000 | 5   | 526,184b    | 0.285μs | 0.285μs | 0.285μs | 0.285μs | 0.000μs | 0.06%  | 1.00x |
| ArrayCheckerBench | benchCheckIfArrayIsEmptyWithEmpty | huge array  | 1000000 | 5   | 64,194,552b | 0.287μs | 0.287μs | 0.287μs | 0.288μs | 0.000μs | 0.12%  | 1.01x |
| ArrayCheckerBench | benchCheckIfArrayIsEmptyWithCount | empty array | 1000000 | 5   | 526,184b    | 0.308μs | 0.309μs | 0.309μs | 0.310μs | 0.001μs | 0.28%  | 1.08x |
| ArrayCheckerBench | benchCheckIfArrayIsEmptyWithCount | huge array  | 1000000 | 5   | 64,194,552b | 0.310μs | 0.311μs | 0.311μs | 0.313μs | 0.001μs | 0.34%  | 1.09x |
+-------------------+-----------------------------------+-------------+---------+-----+-------------+---------+---------+---------+---------+---------+--------+-------+
```

Looks like count takes almost 10% more time, so in relative terms, it's far better indeed. In absolute terms though, it seems to be 30 ns slower… I don't think we should worry about such small differences.

lib/Doctrine/Common/Annotations/DocParser.php
Vincz: Updated!

Vincz:
I was also wondering why this is allowed: `@Name(@Name, @Name)`, but this is not `@Name("1", "2")`. The second case is not allowed because of this line: https://github.com/doctrine/annotations/blob/b81688bc95a16b76b2849bb67004d8f34ddfffd7/lib/Doctrine/Common/Annotations/DocParser.php#L1037
In our case, it means that we can only use one positional argument. It would make sense for us to allow multiple positional arguments as it's also how attributes work. `#[Name("1", "2")]` would work for example as long as the related attribute constructor accept at least two arguments. 
Removing the above line don't break the tests suite and allow multiple positional arguments. 
What do you guys think about this?

beberlei:
@Vincz this only works for the new named argument ctor attributes though or? Because otherwise multiple values are handled as an array of `value` for the old Annotations or not? I would be fine supporting this for new annotations.

Vincz:
I think it would allow `@Named("foo", "bar")` without the `@NamedArgumentConstructor` but it would set `value` as an array like `["foo", "bar"]`. Currently it works if provided values are objects or array but not with scalar values (it raise a Syntax error).
With `@NamedArgumentConstructor` it would use the constructor arguments, ie. "foo" for the first one and "bar" for the second one.

beberlei:
Ok

Vincz:
Ok, I'll add a bunch of tests then and make sure it doesn't break anything.

Vincz:
So, it is now possible to use `@Name("foo", "bar")` with regular annotation and we will get an array of string as explained.

With the `@NamedArgumentConstructor` we can now use multi positional values to initialise arguments.
I also added a check to ensure that it is not possible with `@NamedArgumentConstructor`  to use `@Annot("foo", bar=123, "baz")` as it's an incorrect syntax with attributes / PHP 8 named arguments, and we want to mimic the behaviour. 
I'm not sure about the raised exception type and message, let me know if you have a better idea. 
Also, should I update the documentation to explain more deeply how `@NamedArgumentConstructor` work?

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -1405,4 +1402,57 @@ private function isIgnoredAnnotation(string $name): bool
...
+        ) {
+            // We must ensure that we don't have positional arguments after named ones
+            $positions    = array_keys($positionalArguments);
+            $lastPosition = null;
+            foreach ($positions as $position) {
+                if (
+                    ($lastPosition === null && $position !== 0 ) ||
+                    ($lastPosition !== null && $position !== $lastPosition + 1)
+                ) {
+                    throw $this->syntaxError('Positional arguments after named arguments is not allowed');
```
derrabus, line 1430:
```suggestion
                    throw $this->syntaxError('Positional arguments are not allowed after named arguments');
```

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -1054,30 +1061,20 @@ private function Values(): array
...
-
-                $values['value'][] = $value;
+                $positionalArguments[$k] = $value;
             }
-
-            unset($values[$k]);
         }
 
-        return $values;
+        return ['named_arguments' => $namedArguments, 'positional_arguments' => $positionalArguments];
```
derrabus, line 1077: In `resolvePositionalValues`, you validate `$positionalArguments`. Can't we do this here already? I don't like that we pass around potentially invalid metadata.

lib/Doctrine/Common/Annotations/DocParser.php
derrabus: > In our case, it means that we can only use one positional argument. It would make sense for us to allow multiple positional arguments as it's also how attributes work.

Very well, but Doctrine Annotations only supports one positional argument: the default property. I would consider it out of scope of this PR to introduce the ability to have multiple ones.

Besides, a developer using an annotation does not know whether it has been implemented with a named constructor or not (and they should not need to know, imho). It would be kind of odd if multiple positional arguments only worked in the named argument constructor case.

derrabus: Anyway, I'm a bit late to the party and I see that the discussion has already moved on. I'm fine with allowing multiple positional arguments. Thank you for working on this feature. 😃

Vincz: @derrabus : Actually, Annotations already support multiple positional arguments. It's just that without `@NamedArgumentConstructor` all positional values will be populated as an array into `value`. This `@Name(@Name, @Name)` already work. So in a developer point of view, it doesn't really change anything.

lib/Doctrine/Common/Annotations/DocParser.php
Vincz: The `Values()` method is part of the parsing process. If we move the validation inside, it will have to know what kind of annotation we are talking about. The validation is different is you use `@NamedArgumentConstructor` or not. 
This is valid without `@Name("1", foo="bar", "2")` but should not be valid with `@NamedArgumentConstructor` as it's supposed to mimic a regular constructor call with named arguments.

derrabus: All right, thank you for clarifying. 👍🏻

beberlei (approved):
Very good! I think we got a good state now that is ready to merge and then ship

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -1405,4 +1402,57 @@ private function isIgnoredAnnotation(string $name): bool
...
+                if (isset($values[$property]) || ! isset($positionalArguments[$position])) {
+                    continue;
+                }
+
+                $values[$property] = $positionalArguments[$position];
+            }
+        } else {
+            if (count($positionalArguments) > 0 && ! isset($values['value'])) {
+                if (count($positionalArguments) === 1) {
+                    $value = $positionalArguments[0];
```
kevinpapst, line 1447: See #405

PR #404, opened 2021-02-15
Add PSR-compatible cached annotations reader

alcaeus:
This PR adds a `PsrCachedReader` and deprecates the existing CachedReader as doctrine/cache itself will be deprecated in the near future. I've updated documentation to no longer use the deprecated classes (including the `FileCacheReader`) and duplicated existing tests from `CachedReader` to verify functionality.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,13 +15,15 @@
     "require": {
         "php": "^7.1 || ^8.0",
         "ext-tokenizer": "*",
-        "doctrine/lexer": "1.*"
+        "doctrine/lexer": "1.*",
+        "psr/cache": "^1 || ^2 || ^3"
     },
     "require-dev": {
         "doctrine/cache": "1.*",
         "doctrine/coding-standard": "^6.0 || ^8.1",
         "phpstan/phpstan": "^0.12.20",
-        "phpunit/phpunit": "^7.5 || ^9.1.5"
+        "phpunit/phpunit": "^7.5 || ^9.1.5",
+        "symfony/cache": "^4.4 || ^5.2"
     },
     "config": {
         "sort-packages": true
--- a/docs/en/annotations.rst
+++ b/docs/en/annotations.rst
@@ -116,39 +116,20 @@ This creates a simple annotation reader with no caching other than in
 memory (in php arrays). Since parsing docblocks can be expensive you
 should cache this process by using a caching reader.
 
-You can use a file caching reader, but please note it is deprecated to
-do so:
-
-.. code-block:: php
-
-    use Doctrine\Common\Annotations\FileCacheReader;
-    use Doctrine\Common\Annotations\AnnotationReader;
-
-    $reader = new FileCacheReader(
-        new AnnotationReader(),
-        "/path/to/cache",
-        $debug = true
-    );
-
-If you set the ``debug`` flag to ``true`` the cache reader will check
-for changes in the original files, which is very important during
-development. If you don't set it to ``true`` you have to delete the
-directory to clear the cache. This gives faster performance, however
-should only be used in production, because of its inconvenience during
-development.
-
-You can also use one of the ``Doctrine\Common\Cache\Cache`` cache
-implementations to cache the annotations:
+To cache annotations, you can create a ``Doctrine\Common\Annotations\PsrCachedReader``.
+This reader decorates the original reader and stores all annotations in a PSR-6
+cache:
 
 .. code-block:: php
 
     use Doctrine\Common\Annotations\AnnotationReader;
-    use Doctrine\Common\Annotations\CachedReader;
-    use Doctrine\Common\Cache\ApcCache;
+    use Doctrine\Common\Annotations\PsrCachedReader;
 
-    $reader = new CachedReader(
+    $cache = ... // instantiate a PSR-6 Cache pool
+
+    $reader = new PsrCachedReader(
         new AnnotationReader(),
-        new ApcCache(),
+        $cache,
         $debug = true
     );
 
--- a/lib/Doctrine/Common/Annotations/PsrCachedReader.php
+++ b/lib/Doctrine/Common/Annotations/PsrCachedReader.php
@@ -2,31 +2,29 @@
 
 namespace Doctrine\Common\Annotations;
 
-use Doctrine\Common\Cache\Cache;
+use Psr\Cache\CacheItemPoolInterface;
 use ReflectionClass;
 use ReflectionMethod;
 use ReflectionProperty;
+use Reflector;
 
 use function array_map;
 use function array_merge;
 use function assert;
 use function filemtime;
 use function max;
+use function rawurlencode;
 use function time;
 
 /**
  * A cache aware annotation reader.
- *
- * @deprecated the CachedReader is deprecated and will be removed
- *             in version 2.0.0 of doctrine/annotations. Please use the
- *             {@see \Doctrine\Common\Annotations\PsrCachedReader} instead.
  */
-final class CachedReader implements Reader
+final class PsrCachedReader implements Reader
 {
     /** @var Reader */
     private $delegate;
 
-    /** @var Cache */
+    /** @var CacheItemPoolInterface */
     private $cache;
 
     /** @var bool */
@@ -38,10 +36,7 @@ final class CachedReader implements Reader
     /** @var int[] */
     private $loadedFilemtimes = [];
 
-    /**
-     * @param bool $debug
-     */
-    public function __construct(Reader $reader, Cache $cache, $debug = false)
+    public function __construct(Reader $reader, CacheItemPoolInterface $cache, bool $debug = false)
     {
         $this->delegate = $reader;
         $this->cache    = $cache;
@@ -59,11 +54,7 @@ final class CachedReader implements Reader
             return $this->loadedAnnotations[$cacheKey];
         }
 
-        $annots = $this->fetchFromCache($cacheKey, $class);
-        if ($annots === false) {
-            $annots = $this->delegate->getClassAnnotations($class);
-            $this->saveToCache($cacheKey, $annots);
-        }
+        $annots = $this->fetchFromCache($cacheKey, $class, 'getClassAnnotations', $class);
 
         return $this->loadedAnnotations[$cacheKey] = $annots;
     }
@@ -94,11 +85,7 @@ final class CachedReader implements Reader
             return $this->loadedAnnotations[$cacheKey];
         }
 
-        $annots = $this->fetchFromCache($cacheKey, $class);
-        if ($annots === false) {
-            $annots = $this->delegate->getPropertyAnnotations($property);
-            $this->saveToCache($cacheKey, $annots);
-        }
+        $annots = $this->fetchFromCache($cacheKey, $class, 'getPropertyAnnotations', $property);
 
         return $this->loadedAnnotations[$cacheKey] = $annots;
     }
@@ -129,11 +116,7 @@ final class CachedReader implements Reader
             return $this->loadedAnnotations[$cacheKey];
         }
 
-        $annots = $this->fetchFromCache($cacheKey, $class);
-        if ($annots === false) {
-            $annots = $this->delegate->getMethodAnnotations($method);
-            $this->saveToCache($cacheKey, $annots);
-        }
+        $annots = $this->fetchFromCache($cacheKey, $class, 'getMethodAnnotations', $method);
 
         return $this->loadedAnnotations[$cacheKey] = $annots;
     }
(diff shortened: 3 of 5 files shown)


nicolas-grekas (dismissed):

composer.json
```diff
@@ -15,13 +15,15 @@
     "require": {
         "php": "^7.1 || ^8.0",
         "ext-tokenizer": "*",
-        "doctrine/lexer": "1.*"
+        "doctrine/lexer": "1.*",
+        "psr/cache": "^1 || ^2 || ^3"
```
nicolas-grekas, line 19: maybe `psr/cache-implementation` instead of `psr/cache`?
but does it really depend on it, or is this an optional dep?

nicolas-grekas: this 2nd question might be the answer to the 1st, which would mean `psr/cache` is correct.

docs/en/annotations.rst
```diff
@@ -116,39 +116,19 @@ This creates a simple annotation reader with no caching other than in
...
-.. code-block:: php
-
-    use Doctrine\Common\Annotations\AnnotationReader;
-    use Doctrine\Common\Annotations\CachedReader;
-    use Doctrine\Common\Cache\ApcCache;
-
-    $reader = new CachedReader(
-        new AnnotationReader(),
-        new ApcCache(),
+        new ApcuAdapter(),
```
stof: Maybe write this without hardcoding symfony/cache in the readme as any PSR-6 implementation is OK:

```
$cache = ... // instantiate a PSR-6 Cache pool

$reader = new PsrCachedReader(
    new AnnotationReader(),
    $cache,
    $debug = true
);
```

lib/Doctrine/Common/Annotations/PsrCachedReader.php
```diff
@@ -0,0 +1,234 @@
...
+    private $debug;
+
+    /** @var array<string, array<object>> */
+    private $loadedAnnotations = [];
+
+    /** @var int[] */
+    private $loadedFilemtimes = [];
+
+    /**
+     * @param bool $debug
```
stof: as the min supported PHP version is 7.1, you could use an actual typehint

lib/Doctrine/Common/Annotations/PsrCachedReader.php
```diff
@@ -0,0 +1,234 @@
...
+
+        $item = $this->cache->getItem($cacheKey);
+        if (! $item->isHit() || ($this->debug && ! $this->refresh($cacheKey, $class))) {
+            $this->cache->save($item->set($this->delegate->{$method}($reflector)));
+        }
+
+        return $item->get();
+    }
+
+    private function refresh(string $cacheKey, ReflectionClass $class): bool
```
stof, line 167: the method is not clear IMO. What does the boolean return value for `refresh()` means ? that it needs refreshing or that it is still fresh ?

docs/en/annotations.rst
alcaeus: Ah, good point. Will update 👍

lib/Doctrine/Common/Annotations/PsrCachedReader.php
alcaeus: Added an explanatory comment.

stof: maybe the method should be named `isFresh` instead of `refresh` then, so that the name reflects what it does.

lib/Doctrine/Common/Annotations/PsrCachedReader.php
```diff
@@ -0,0 +1,237 @@
...
+    /** @var bool */
+    private $debug;
+
+    /** @var array<string, array<object>> */
+    private $loadedAnnotations = [];
+
+    /** @var int[] */
+    private $loadedFilemtimes = [];
+
+    public function __construct(Reader $reader, CacheItemPoolInterface $cache, bool $debug = false)
```
nicolas-grekas, line 39: Given this line:
https://github.com/doctrine/orm/blob/4e0f6837d04306417640c6be1ab5a20f9cb9e5d6/lib/Doctrine/ORM/Configuration.php#L165

I think we should make `$cache` nullable, and make the implem deal with having no PSR cache. That would free `doctrine/orm` from requiring any `psr/cache` implementation when this line will go away.

nicolas-grekas: ~And to make `ensureProductionSettings()` useful, we should add a method that returns the PsrCache instance, to be used in:
https://github.com/doctrine/orm/blob/4e0f6837d04306417640c6be1ab5a20f9cb9e5d6/lib/Doctrine/ORM/Configuration.php#L374~

nicolas-grekas: (forget about my last comment, ensureProductionSettings is unrelated to the annotation reader)

lib/Doctrine/Common/Annotations/PsrCachedReader.php
nicolas-grekas: `refresh` also does update the timestamp of the cached value
`isFresh` wouldn't convey that meaning

lib/Doctrine/Common/Annotations/PsrCachedReader.php
stof: how would a PsrCachedReader work without a PSR-6 implementation (and so no CachePool and no CacheItem) ?

lib/Doctrine/Common/Annotations/PsrCachedReader.php
stof: but this method is not refreshing the data at all. So the current names lies about what it does. That's more a `doesNotNeedRefresh()` method than a `refresh()` one.

lib/Doctrine/Common/Annotations/PsrCachedReader.php
nicolas-grekas: there is an internal array cache, so it would do something useful :)

lib/Doctrine/Common/Annotations/PsrCachedReader.php
nicolas-grekas: it refreshes the freshness timestamp :)

lib/Doctrine/Common/Annotations/PsrCachedReader.php
```diff
@@ -0,0 +1,237 @@
...
+
+        return null;
+    }
+
+    /**
+     * Clears loaded annotations.
+     *
+     * @return void
+     */
+    public function clearLoadedAnnotations()
```
greg0ire: This method is not part of an explicitly declared interface, should we use a return type declaration and get rid of the tautological description? My hope is that we will migrate existing methods to this form at some point.

```suggestion
    public function clearLoadedAnnotations(): void
```

tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
```diff
@@ -0,0 +1,299 @@
...
+
+        $reader->clearLoadedAnnotations();
+
+        $this->assertCount(0, $loadedAnnotationsProperty->getValue($reader));
+        $this->assertCount(0, $loadedFilemtimesProperty->getValue($reader));
+    }
+
+    /**
+     * As there is a cache on loadedAnnotations, we need to test two different
+     * method's annotations of the same file
```
greg0ire:
```suggestion
     * methods' annotations of the same file
```

tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
```diff
@@ -0,0 +1,299 @@
...
+     *
+     * We test four things
+     * 1. we load the file (and its filemtime) for method1 annotation with fresh cache
```
greg0ire:
```suggestion
     * 1. we load the file (and its filemtime) for method1's annotation with fresh cache
```

tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
```diff
@@ -0,0 +1,299 @@
...
+            ->method('isHit')
+            ->willReturn(true);
+        $timeCacheItem
+            ->expects($this->any())
+            ->method('get')
+            ->willReturn($lastCacheModification);
+
+        $cache = $this->createMock(CacheItemPoolInterface::class);
+        $cache
+            ->expects($this->any())
```
greg0ire: This line has no effect.
```suggestion
```

tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
```diff
@@ -0,0 +1,299 @@
...
+
+        $cacheItem1     = $this->createCacheItem($cacheKeyMethod1, true, [$route1]);
+        $timeCacheItem1 = $this->createCacheItem('[C]' . $cacheKeyMethod1, true, $cacheTime);
+
+        $cacheItem2     = $this->createCacheItem($cacheKeyMethod2, true, [$route2]);
+        $timeCacheItem2 = $this->createCacheItem('[C]' . $cacheKeyMethod2, true, $cacheTime);
+
+        $cache = $this->createMock(CacheItemPoolInterface::class);
+        $cache
+            ->expects($this->any())
```
greg0ire:
```suggestion
```

tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
```diff
@@ -0,0 +1,299 @@
...
+
+        $route          = new Route();
+        $route->pattern = '/someprefix';
+
+        $cacheItem     = $this->createCacheItem($cacheKey, true, []);
+        $timeCacheItem = $this->createCacheItem('[C]' . $cacheKey, true, $lastCacheModification);
+
+        $cache = $this->createMock(CacheItemPoolInterface::class);
+        $cache
+            ->expects($this->any())
```
greg0ire:
```suggestion
```

tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
```diff
@@ -0,0 +1,299 @@
...
+
+    protected function doTestCacheFresh(string $className, int $lastCacheModification): void
+    {
+        $cacheKey       = rawurlencode($className);
+        $route          = new Route();
+        $route->pattern = '/someprefix';
+
+        $cacheItem = $this->createMock(CacheItemInterface::class);
+        $cacheItem
+            ->expects($this->any())
```
greg0ire:
```suggestion
```

tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
```diff
@@ -0,0 +1,299 @@
...
+            ->method('isHit')
+            ->willReturn(true);
+        $cacheItem
+            ->expects($this->any())
```
greg0ire:
```suggestion
```

tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
```diff
@@ -0,0 +1,299 @@
...
+        return $reader;
+    }
+
+    protected function doTestCacheFresh(string $className, int $lastCacheModification): void
+    {
+        $cacheKey       = rawurlencode($className);
+        $route          = new Route();
+        $route->pattern = '/someprefix';
+
+        $cacheItem = $this->createMock(CacheItemInterface::class);
```
greg0ire, line 238: Provided we use a recent enough version of PHPUnit, let's convey our intent about `$cacheItem`:
```suggestion
        $cacheItem = $this->createStub(CacheItemInterface::class);
```

tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
```diff
@@ -0,0 +1,299 @@
...
+        $cacheItem
+            ->expects($this->any())
+            ->method('isHit')
+            ->willReturn(true);
+        $cacheItem
+            ->expects($this->any())
+            ->method('get')
+            ->willReturn([$route]);
+
+        $timeCacheItem = $this->createMock(CacheItemInterface::class);
```
greg0ire, line 246:
```suggestion
        $timeCacheItem = $this->createStub(CacheItemInterface::class);
```

tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
```diff
@@ -0,0 +1,299 @@
...
+        $timeCacheItem
+            ->expects($this->any())
```
greg0ire:
```suggestion
```

tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
```diff
@@ -0,0 +1,299 @@
...
+            ->method('isHit')
+            ->willReturn(true);
+        $timeCacheItem
+            ->expects($this->any())
```
greg0ire:
```suggestion
```

tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
```diff
@@ -0,0 +1,299 @@
...
+        $this->cache = new ArrayAdapter();
+
+        return new PsrCachedReader(new AnnotationReader(), $this->cache);
+    }
+
+    /** @param mixed $value */
+    private function createCacheItem(string $key, bool $isHit, $value = null): CacheItemInterface
+    {
+        return Closure::bind(
+            static function ($key, $value, $isHit) {
```
greg0ire:
```suggestion
            static function (string $key, $value, bool $isHit): CacheItem {
```

greg0ire:
@alcaeus sorry for suggesting `createStub`, you won't be able to use it because of the 7.1 job. I will revert this for you.

greg0ire (approved):

PR #406, opened 2021-02-21
Fix #405

### LINKED ISSUE #405: Bug "Undefined offset: 0" introduced in 1.12
There seems to be a new bug introduced with the latest 1.12 version.
Since I did not use it directly but used [l5-swagger](https://github.com/DarkaOnLine/L5-Swagger), which is using [swagger-php](https://github.com/zircote/swagger-php) which uses doctrine/annotations, I'm not exactly sure how to reproduce this easily.

What I can tell is that in [line 1447 of the DocParser.php](https://github.com/doctrine/annotations/compare/1.11.2...1.12.0#diff-e5d2e0837a5b53f6d38f5a1f6d4f154c757e2cba3eb3259751ed19944339a6edR1447) index "0" of the variable `$positionalArguments` is directly referenced, but the first set index in that variable at this point for me was "2".
This is what I saw when I `var_export`ed the content right before it failed:
image

So this should be an easy fix I think, but I'm not deep enough in that library to be able to both create a test and fix this issue myself.

For now I reverted back to 1.11.2 and everything is working again.

Vincz:
Fix a bug introduced by my previous PR. 
Positional arguments following named argument should be valid when `@NamedArgumentConstructor` is not used.

Fixes #405

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -15,6 +15,7 @@ use stdClass;
 
 use function array_keys;
 use function array_map;
+use function array_pop;
 use function array_values;
 use function class_exists;
 use function constant;
@@ -1424,7 +1425,7 @@ EXCEPTION
             $lastPosition = null;
             foreach ($positions as $position) {
                 if (
-                    ($lastPosition === null && $position !== 0 ) ||
+                    ($lastPosition === null && $position !== 0) ||
                     ($lastPosition !== null && $position !== $lastPosition + 1)
                 ) {
                     throw $this->syntaxError('Positional arguments after named arguments is not allowed');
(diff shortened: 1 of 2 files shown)


derrabus (dismissed):

greg0ire:
Please improve your commit message according to [the contributing guide](https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches). Here, it's hard to tell why it doesn't work, and why your changes make it work just by reading the commit message.

Vincz:
Done. Is it better?

greg0ire:
The contents is better yeah 👍
I just fixed the form:
- reference to GH ticket is optional, and goes at the end of the body
- the body is wrapped at 72 chars (consider configuring your editor for that)
- the title is about what you did, and no longer references GH

greg0ire (approved):

michaljusiega:
I can confirm this fix solve issue ;)

greg0ire:
Thanks @Vincz !

Issue #408, opened 2021-03-17
Support for annotation in class constant docBlock


matthieu88160:
Hello all,

Doctrine annotations library allows to parse annotations from lots of class members but seems to miss the class constants from the reader scope. Even if the runtime usage of class constant annotations will not be widely used, it could be a nice addition to the current features, and an actual use case for me and my company.

By the way, I propose to add two new methods in the Reader interface:
* `public function getConstantAnnotations(ReflectionClassConstant $constant);`
* `public function getConstantAnnotation(ReflectionClassConstant $constant, $annotationName);`

As `ReflectionClassConstant` is part of PHP since version `7.1` it fit with current doctrine/annotations requirements.

greg0ire:
Out of curiosity, what use case would you have for this in your company?

matthieu88160:
Working on a system to sync classes between multiple language, driven by PHP as main development language. Our principal use case is to generate Enums from class/interface constants. We are using annotations to control generation behaviors.

greg0ire:
Funny, I just saw this: https://twitter.com/PHPRFCBot/status/1372252058028449797

matthieu88160:
> Funny, I just saw this: https://twitter.com/PHPRFCBot/status/1372252058028449797

😄 Such a great feature. Enums are just an example case, constants still fit well for internal class configuration constant even if Enums are implemented.

greg0ire:
I was not implying it at all that it wouldn't be worth it, don't worry ;)

derrabus:
Closing because we won't add new features to this library.

PR #409, opened 2021-03-17
Class constant annotation addition

matthieu88160:
Draft for #408

Adding Target::TARGET_CONSTANT
Adding getConstantAnnotations method to Reader
Adding getConstantAnnotation method to Reader

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Common\Annotations;
 
+use ReflectionClassConstant;
 use Doctrine\Common\Annotations\Annotation\IgnoreAnnotation;
 use Doctrine\Common\Annotations\Annotation\Target;
 use ReflectionClass;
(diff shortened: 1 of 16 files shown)


lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -386,4 +419,27 @@ private function collectParsingMetadata($reflection): void
 
         $this->ignoredAnnotationNames[$type][$name] = $ignoredAnnotationNames;
     }
+
+    /**
+     * Retrieves imports for constants.
+     *
+     * @return array<string, class-string>
+     */
+    private function getConstantImports(ReflectionClassConstant $constant)
```
greg0ire:
```suggestion
    private function getConstantImports(ReflectionClassConstant $constant): array
```

lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -386,4 +419,27 @@ private function collectParsingMetadata($reflection): void
 
         $this->ignoredAnnotationNames[$type][$name] = $ignoredAnnotationNames;
     }
+
+    /**
+     * Retrieves imports for constants.
+     *
+     * @return array<string, class-string>
```
greg0ire:
```suggestion
     * @return array<string, class-string|trait-string>
```

greg0ire (commented):
This should target 1.14.x since it's a new feature

lib/Doctrine/Common/Annotations/Reader.php
```diff
@@ -77,4 +78,26 @@ public function getPropertyAnnotations(ReflectionProperty $property);
...
+
+    /**
+     * Gets the annotations applied to a constant.
+     *
+     * @param ReflectionClassConstant $constant The ReflectionClassConstant of the constant
+     *                                          from which the annotations should be read.
+     *
+     * @return array<object> An array of Annotations.
+     */
+    public function getConstantAnnotations(ReflectionClassConstant $constant);
```
stof, line 90: Adding new methods in an interface is a BC break

derrabus:
Closing because we won't add new features to this library.

PR #410, opened 2021-04-22
Allow doctrine/cache 2.x for tests

alcaeus:
This PR changes the `CachedReaderTest` class to not test if doctrine/cache 2.x is installed, as that will no longer contain the required `ArrayCache` class. Since the reader is deprecated, I don't think it's worth updating the test to work with an adapter.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "psr/cache": "^1 || ^2 || ^3"
     },
     "require-dev": {
-        "doctrine/cache": "1.*",
+        "doctrine/cache": "^1.11 || ^2.0",
         "doctrine/coding-standard": "^6.0 || ^8.1",
         "phpstan/phpstan": "^0.12.20",
         "phpunit/phpunit": "^7.5 || ^9.1.5",
--- a/tests/Doctrine/Performance/Common/Annotations/CachedReadPerformanceWithInMemoryBench.php
+++ b/tests/Doctrine/Performance/Common/Annotations/CachedReadPerformanceWithInMemoryBench.php
@@ -5,17 +5,17 @@ declare(strict_types=1);
 namespace Doctrine\Performance\Common\Annotations;
 
 use Doctrine\Common\Annotations\AnnotationReader;
-use Doctrine\Common\Annotations\CachedReader;
-use Doctrine\Common\Cache\ArrayCache;
+use Doctrine\Common\Annotations\PsrCachedReader;
 use Doctrine\Tests\Common\Annotations\Fixtures\Controller;
 use ReflectionMethod;
+use Symfony\Component\Cache\Adapter\ArrayAdapter;
 
 /**
  * @BeforeMethods({"initialize"})
  */
 final class CachedReadPerformanceWithInMemoryBench
 {
-    /** @var CachedReader */
+    /** @var PsrCachedReader */
     private $reader;
 
     /** @var ReflectionMethod */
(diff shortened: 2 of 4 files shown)


stof (dismissed):

SenseException (dismissed):

alcaeus:
I don't quite understand the test failures. I can't reproduce them locally, they don't appear on every run, and it's always the same kind of test (in both CachedReader and PsrCachedReader) that makes me feel something is off there. Furthermore, the last 1.13.x build passed, and the logic tested by the test in question was not modified since. If y'all have any ideas (beyond skipping the offending test) please let me know.

greg0ire:
This is hard to troubleshoot: running either failing tests separately fails because of another reason: `Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The annotation "@Name" in class Doctrine\Tests\Common\Annotations\TestIgnoresNonAnnotationsClass was never imported. Did you maybe forget to add a "use" statement for this annotation?`

If you do a bisection, you can find that it's sufficient to have `DocParserTest` for it to pass again:

```xml
<?xml version="1.0" encoding="utf-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
         colors="true"
         beStrictAboutOutputDuringTests="true"
         beStrictAboutTodoAnnotatedTests="true"
         failOnRisky="true"
         failOnWarning="true"
         bootstrap="./tests/Doctrine/Tests/TestInit.php"
>
    <php>
        <ini name="error_reporting" value="-1" />
    </php>

    <testsuites>
        <testsuite name="Doctrine Annotations Test Suite">
            <file>tests/Doctrine/Tests/Common/Annotations/DocParserTest.php</file>
            <file>tests/Doctrine/Tests/Common/Annotations/CachedReaderTest.php</file>
        </testsuite>
    </testsuites>

    <filter>
        <whitelist>
            <directory suffix=".php">./lib/Doctrine/</directory>
        </whitelist>
    </filter>
</phpunit>
```

UPDATE: turns out `tests/Doctrine/Tests/Common/Annotations/DocParserTest.php` contains a class called `Name` :see_no_evil: 

I have to "go" to work now, will investigate this later.

greg0ire:
@alcaeus I fixed the issue I found in a separate PR, and on that PR, I'm experiencing the same issue as here, see #411 
So we know it doesn't come from the code in this PR.

greg0ire:
Interestingly, the 7.2 build with lowest dependencies does not fail, so maybe this is due do a package we use (`doctrine/cache` looks like it could be a culprit?)

UPD: looks like it, preventing the installation of 1.11 fixed the build in #411

greg0ire:
Ah no that was confirmation bias, I restarted the build and now it fails again, and it also fails with prefer lowest, so that is not an issue with dependencies. It probably has to do with time. @bastnic , could you please look into it?

alcaeus:
Hmm, indeed it does. I'll have to check what's going on 🤔

alcaeus:
@greg0ire Any objections to merging this since the build failure seems unrelated?

greg0ire:
Let's try to see if we can get a green build with this PR? Because if it always fails, then something is off…

bastnic:
> Ah no that was confirmation bias, I restarted the build and now it fails again, and it also fails with prefer lowest, so that is not an issue with dependencies. It probably has to do with time. @bastnic , could you please look into it?

would love to, but quite busy right now

alcaeus:
@greg0ire I skipped the flaky test since it's always a different set of builds failing. I'd suggest we revisit this in a separate PR once @bastnic (or someone else) has the time to investigate.

greg0ire (approved):

PR #412, opened 2021-05-16
Don't hit the reader if cache is fresh

derrabus:
Spotted while working on symfony/symfony#41230.

In debug mode, the `PsrCachedReader` hits the wrapped reader although the cache is still fresh. This happens because the cache items that stores the modification date is not written initially. This PR attempts to fix this problem.

### DIFF
--- a/lib/Doctrine/Common/Annotations/PsrCachedReader.php
+++ b/lib/Doctrine/Common/Annotations/PsrCachedReader.php
@@ -151,7 +151,7 @@ final class PsrCachedReader implements Reader
         $cacheKey = rawurlencode($cacheKey);
 
         $item = $this->cache->getItem($cacheKey);
-        if (! $item->isHit() || ($this->debug && ! $this->refresh($cacheKey, $class))) {
+        if (($this->debug && ! $this->refresh($cacheKey, $class)) || ! $item->isHit()) {
             $this->cache->save($item->set($this->delegate->{$method}($reflector)));
         }
 
--- a/tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/PsrCachedReaderTest.php
@@ -8,6 +8,7 @@ use Doctrine\Common\Annotations\PsrCachedReader;
 use Doctrine\Common\Annotations\Reader;
 use Doctrine\Tests\Common\Annotations\Fixtures\Annotation\Route;
 use Doctrine\Tests\Common\Annotations\Fixtures\ClassThatUsesTraitThatUsesAnotherTraitWithMethods;
+use Doctrine\Tests\Common\Annotations\Fixtures\ClassWithClassAnnotationOnly;
 use Psr\Cache\CacheItemInterface;
 use Psr\Cache\CacheItemPoolInterface;
 use ReflectionClass;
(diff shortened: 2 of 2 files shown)


alcaeus (dismissed):

derrabus:
The PHPStan failure is unrelated.

greg0ire:
`ArrayCache` was indeed removed 3 weeks ago by @alcaeus in https://github.com/doctrine/cache/commit/b3fc4bc
The analyzed test that crash are meant to be run with `doctrine/cache` 1, but PHPStan uses v2

lib/Doctrine/Common/Annotations/PsrCachedReader.php
```diff
@@ -148,10 +148,19 @@ private function fetchFromCache(
         string $method,
         Reflector $reflector
     ): array {
-        $cacheKey = rawurlencode($cacheKey);
+        $cacheKey          = rawurlencode($cacheKey);
+        $fetchFromDelegate = false;
 
         $item = $this->cache->getItem($cacheKey);
-        if (! $item->isHit() || ($this->debug && ! $this->refresh($cacheKey, $class))) {
+        if (! $item->isHit()) {
```
ostrolucky: You could have just done this instead of current patch
```suggestion
    if (($this->debug && ! $this->refresh($cacheKey, $class)) || ! $item->isHit()) {
```

greg0ire:
https://github.com/doctrine/annotations/pull/413 fixes the build issue (by not analyzing that file)

lib/Doctrine/Common/Annotations/PsrCachedReader.php
greg0ire: :thinking: it looks like you are just swapping the operands of `||` in that suggestion, is that it?

ostrolucky: yes

derrabus: Indeed. 😅

derrabus:
And we're green. 🎉

alcaeus (approved):

ostrolucky (approved):

alcaeus:
Thanks @derrabus!

soyuka:
I have an issue on API Platform since that change. It looks like `refresh` can't be called if `isHit` is truthy. You can reproduce that error using:
```
docker run -p 27017:27017 mongo:latest
```

```
git clone git@github.com:soyuka/core
cd core
git switch fix/phpstan
composer update
composer req --dev symfony/uid symfony/intl
MONGODB_URL=mongodb://localhost:27017 APP_ENV=mongodb php -d memory_limit=-1 ./vendor/bin/behat --profile=mongodb --stop-on-failure features/doctrine/search_filter.feature:7
```

Check the stacktrace by adding in `features/doctrine/search_filter:11`:

```
    Then print last JSON response
```

Maybe that a fix should be done in mongodb-odm? Let me know if I can help further.

alcaeus:
@soyuka can you please create a new issue? I fail to see the relationship between what this PR fixed, the issue you reported in the comment, and the PR that linked to this.

dunglas:
I confirm that reverting this patch fixes the problem in API Platform.

alcaeus:
Thanks for confirming @dunglas! We'll take another look.

alcaeus:
@dunglas unfortunately I can't find the time to investigate this issue. Would you be able to create a minimal reproducer for us to start our investigation with? Thanks!

PR #419, opened 2021-07-16
Add NamedArgumentConstructor to reserved annotations

derrabus:
When introducing the `NamedArgumentConstructor` annotation, I accidentally created the edge-case of acore annotation that needs to be imported. This is not the case for other core annotations like `Annotation` or `Target`. This PR attempts to fix that.


greg0ire (approved):

greg0ire:
Thanks @derrabus !

PR #422, opened 2021-07-28
Add psalm annotations to ignored annotation name

azjezz:
@pure annotation is mainly used by psalm to mark pure functions.

### DIFF
--- a/lib/Doctrine/Common/Annotations/ImplicitlyIgnoredAnnotationNames.php
+++ b/lib/Doctrine/Common/Annotations/ImplicitlyIgnoredAnnotationNames.php
@@ -139,16 +139,21 @@ final class ImplicitlyIgnoredAnnotationNames
 
     private const SlevomatCodingStandard = ['phpcsSuppress' => true];
 
-    private const PhpStan = [
+    private const Phan = ['suppress' => true];
+
+    private const Rector = ['noRector' => true];
+
+    private const StaticAnalysis = [
+        // PHPStan, Psalm
         'extends' => true,
         'implements' => true,
         'template' => true,
         'use' => true,
+
+        // Psalm
+        'pure' => true,
+        'immutable' => true,
     ];
-
-    private const Phan = ['suppress' => true];
-
-    private const Rector = ['noRector' => true];
 
     public const LIST = self::Reserved
         + self::WidelyUsedNonStandard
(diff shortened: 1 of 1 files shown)


greg0ire (commented):
Hi! Please squash both commits together.

What is the reason for Psalm supporting `@pure` in addition to `@psalm-pure`? Can it really claim that annotation for itself, as opposed to PHPStan (for instance)?

azjezz:
done.

This issue appears when using `azjezz/psl` library, and PSL has a CS rule that forbids Psalm specific annotations, in hopes that other tools in the future understand these ( e.g if PhpStan would to add pure function support, PSL wouldn't need to add a new annotation ) - ref: https://github.com/azjezz/psl/blob/1.8.x/src/Psl/Collection/Vector.php#L47

malarzm:
> PSL has a CS rule that forbids Psalm specific annotations

Maybe nitpicking, but linked file has other `@psalm-` annotations: `@psalm-readonly` and `@psalm-mutation-free`

greg0ire:
> This issue appears when using azjezz/psl library, and PSL has a CS rule that forbids Psalm specific annotations, in hopes that other tools in the future understand these ( e.g if PhpStan would to add pure function support, PSL wouldn't need to add a new annotation ) - ref: https://github.com/azjezz/psl/blob/1.8.x/src/Psl/Collection/Vector.php#L47


Ah yes, that library I should really try out someday :slightly_smiling_face: 

When I see the rest of the file, there are mostly annotations that are unlikely to be picked by any other tools. There are also annotations that are attributed to `PhpStan` but [work with Psalm too](https://psalm.dev/docs/annotating_code/templated_annotations/)

Maybe using tool names as consts is a bad idea because of this, in some cases? How about having

```php
    private const StaticAnalysis = [
        // PhpStan, Psalm 
        'extends' => true,
        'implements' => true,
        'template' => true,
        'use' => true,

       // Psalm
        'pure' => true,
        'immutable' => true,
    ];
```

?

azjezz:
> Maybe nitpicking, but linked file has other @psalm- annotations: @psalm-readonly and @psalm-mutation-free

because @mutation-free and @readonly don't work with psalm IIRC.

azjezz:
> Maybe using tool names as consts is a bad idea because of this, in some cases? How about having

That sounds like a good idea 👍

greg0ire (approved):

SenseException (approved):

malarzm (approved):

greg0ire:
Thanks @azjezz !

PR #424, opened 2021-08-09
Add support for function annotations in PsrCachedReader.

jenkoian:
Working on a project which uses functions and I want to add annotations to the functions and have them cached. Am aware I could just create my own cached reader and do it that way, but thought it might be nice to have support in the library itself. 

One thing, I wasn't sure on is if it should throw an exception if attempt is made to call `getFunctionAnnotations` on the cached reader when the delegate doesn't support it (as it's not on the interface)? Any thoughts?

### DIFF
--- a/lib/Doctrine/Common/Annotations/PsrCachedReader.php
+++ b/lib/Doctrine/Common/Annotations/PsrCachedReader.php
@@ -4,6 +4,7 @@ namespace Doctrine\Common\Annotations;
 
 use Psr\Cache\CacheItemPoolInterface;
 use ReflectionClass;
+use ReflectionFunction;
 use ReflectionMethod;
 use ReflectionProperty;
 use Reflector;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Annotations/PsrCachedReader.php
```diff
@@ -135,23 +136,69 @@ public function getMethodAnnotation(ReflectionMethod $method, $annotationName)
...
+    /**
+     * @param ReflectionClass|ReflectionFunction $reflection The ReflectionClass or ReflectionFunction in which to check
+     * last modification time on.
+     *
+     * @return mixed[]
+     */
     private function fetchFromCache(
         string $cacheKey,
-        ReflectionClass $class,
+        /**ReflectionClass|ReflectionFunction*/ $reflection,
```
jenkoian, line 194: Not sure where you stand on this kind of thing, happy to drop if you'd prefer.

derrabus:
Hello. Sorry, nobody's picked up your PR yet. We won't add any new features to Doctrine Annotations at this point which is why I'm going to close this PR. Thank you very much for your proposal.

PR #426, opened 2021-10-06
open up AnnotationReader inheritance

Hikariii:
This PR opens up inheritance possibilities for the AnnotationReader.
Properties are static but difficult to manipulate when (for example) filtering is needed.
When this is changed, intercepting and adding custom options is way easier.
This does not break anything, so in my opinion it can be easily merged in 1.13

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -24,7 +24,7 @@ class AnnotationReader implements Reader
      *
      * @var array<string, class-string>
      */
-    private static $globalImports = [
+    protected static $globalImports = [
         'ignoreannotation' => Annotation\IgnoreAnnotation::class,
     ];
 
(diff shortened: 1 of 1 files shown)


greg0ire:
I'd rather we keep things as closed as possible. In most cases decoration can be used to workaround issues, but maybe for some reasons that's not the case here? What is it you're trying to achieve?

Hikariii:
I really need 2 configured AnnotationReader instances configured differently. But the static private properties makes it impossible. The better way forward may be to move from static configuration to class properties, but that is a BC break.

stof:
@Hikariii making everything protected means that *any* refactoring of the class becomes a BC break, meaning that this PR freezes the annotation parser forever in the 1.x series. So this looks even worse

Hikariii:
Hmm valid point, this is not the right way forward. Are you open to removing the static keyword for the first few properties?
This would solve the issues of configuring two AnnotationReader instances differently.

greg0ire:
> Are you open to removing the static keyword for the first few properties?

That sounds good to me. No BC-break here since the properties are private (this is why `private` is good ;) )

stof:
@greg0ire but the public methods configuring them are static too, so that's not that simple.

We need to provide a new non-static configuration API (and merge the static config with the non-static one for BC)

Hikariii:
As suggested a new PR is drafted that does just what @stof suggested: https://github.com/doctrine/annotations/pull/433

PR #429, opened 2021-10-15
Reuse workflows from upstream

greg0ire:
I already changed the settings to require the new jobs.
Static analysis is not included in this job because Psalm is not implemented (yet).


greg0ire:
Closing in favor of #430

PR #430, opened 2021-10-15
Setup Psalm and reuse upstream workflows

greg0ire:


psalm.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<psalm
+    errorLevel="7"
```
stof, line 3: isn't it quite high as level (knowing that for psalm, higher levels are less strict) ?

derrabus: I assume, @greg0ire wants to start easy, so we can iterate further once the CI job is in place.

derrabus (approved):

psalm.xml
greg0ire: I let Psalm pick the level itself with `psalm --init`. Level 6 finds 40 errors, which would make this PR a bit fat IMO.

greg0ire: Yes exactly @derrabus . Also, this allows me to reuse all upstream workflows: the static analysis workflow cannot be reused if we don't have Psalm in the project.

SenseException (approved):

PR #433, opened 2021-10-26
added non-static ignoredNamespace config for Reader

Hikariii:
In light of the discussion of https://github.com/doctrine/annotations/pull/426 I created this PR to allow per-Reader instance configuration of ignored namespaces.
The static configuration is kept for BC compatibility and tests are added mixing static and private configuration.

### DIFF
--- a/lib/Doctrine/Common/Annotations/AnnotationReader.php
+++ b/lib/Doctrine/Common/Annotations/AnnotationReader.php
@@ -67,6 +67,15 @@ class AnnotationReader implements Reader
     }
 
     /**
+     * A list with annotation namespaces that are not causing exceptions when not resolved to an annotation class.
+     *
+     * The names are case sensitive.
+     *
+     * @var array<string, true>
+     */
+    private $ignoredNamespaces = [];
+
+    /**
      * Annotations parser.
      *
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Annotations/AnnotationReader.php
```diff
@@ -133,6 +142,16 @@ class_exists(IgnoreAnnotation::class);
         $this->phpParser = new PhpParser();
     }
 
+    /**
+     * Add a new annotation to the ignored annotation namespaces with regard to exception handling.
+     *
+     * @param string $namespace
+     */
+    public function addIgnoredNamespace($namespace)
```
greg0ire: This is a new method. As such it can use type declarations, and we should IMO mark it as `final`.

```suggestion
    final public function addIgnoredNamespace(string $namespace): void
```

greg0ire:
Should we also deprecate `globalIgnoredNamespaces` and related methods? Should we also deprecate other static properties?

SenseException:
Having a common structure makes it more intuitive. This seems to look like a replacement of the static properties and is changing how the class is used. It looks like that this needs an entry in the docs too.

PR #436, opened 2022-02-17
#186 check if file exists before asking for mtime

### LINKED ISSUE #186: CachedReader fails if Reflection class is eval'd code
If the ReflectionClass is eva'd code, then it crashes when checking when the cache is fresh. It tries to [CachedReader::getLastModification](https://github.com/doctrine/annotations/blob/master/lib/Doctrine/Common/Annotations/CachedReader.php#L236) of the eval'd code and results in:

**Warning:` filemtime(): stat failed for /app/src/MyCode/EntityGenerator.php(36) : eval()'d code**

lstrojny:
`doctrine/annotations` handles reading annotations from eval’ed code (as in mock objects) just fine, except that it attempts to determine the modification time via `ReflectionClass::getFile()` which will return the file and line where the `eval()` happened. This change adds fix and test for the `PsrCachedReader` and also fixes `CachedReader` but without a tests since it’s deprecated.

Fixes #186

### DIFF
--- a/lib/Doctrine/Common/Annotations/CachedReader.php
+++ b/lib/Doctrine/Common/Annotations/CachedReader.php
@@ -10,7 +10,9 @@ use ReflectionProperty;
 use function array_map;
 use function array_merge;
 use function assert;
+use function file_exists;
 use function filemtime;
+use function is_file;
 use function max;
 use function time;
 
@@ -231,7 +233,7 @@ final class CachedReader implements Reader
         $parent = $class->getParentClass();
 
         $lastModification =  max(array_merge(
-            [$filename ? filemtime($filename) : 0],
+            [self::getFileMtime($filename)],
             array_map(function (ReflectionClass $reflectionTrait): int {
                 return $this->getTraitLastModificationTime($reflectionTrait);
             }, $class->getTraits()),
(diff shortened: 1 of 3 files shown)


lib/Doctrine/Common/Annotations/CachedReader.php
```diff
@@ -265,4 +267,9 @@ private function getTraitLastModificationTime(ReflectionClass $reflectionTrait):
 
         return $this->loadedFilemtimes[$fileName] = $lastModificationTime;
     }
+
+    private static function getFileMtime(string $fileName): int
+    {
+        return file_exists($fileName) && is_file($fileName) ? filemtime($fileName) : 0;
```
greg0ire, line 273: Why do we need to call both `file_exists` and `is_file`?

SenseException:
``` php
return file_exists($fileName) ? filemtime($fileName) : 0;
```
should do the trick too, shouldn't it?

SenseException (changes requested):

lib/Doctrine/Common/Annotations/CachedReader.php
gus3000: The edge case that ` is_file`  checks for here is if ` $fileName` is a directory. I guess it makes sense to return 0 in this case, but I don't know if :
1. It's the focus of this PR
2.  `getFileMtime()` can get called with a directory name at all

greg0ire: Isn't `file_exists` redundant if we have `is_file`?

gus3000: [indeed](https://www.php.net/manual/en/function.is-file.php#refsect1-function.is-file-returnvalues).

PR #438, opened 2022-05-13
Catch and process errors that occure during annotation instantiation

### LINKED ISSUE #437: Catch TypeError during annotation instantiation and throw a more meaningful AnnotationException
I ran into this error:

```
In OneToMany.php line 45:

  [TypeError]
  Argument 3 passed to Doctrine\ORM\Mapping\OneToMany::__construct() must be of the type array or null,
  string given, called in vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php on line 971

Exception trace:
  at vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/OneToMany.php:45
 Doctrine\ORM\Mapping\OneToMany->__construct() at vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php:971
 Doctrine\Common\Annotations\DocParser->Annotation() at vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php:719
 Doctrine\Common\Annotations\DocParser->Annotations() at vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php:376
 Doctrine\Common\Annotations\DocParser->parse() at vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationReader.php:178
 Doctrine\Common\Annotations\AnnotationReader->getPropertyAnnotations() at vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/PsrCachedReader.php:155
```

I traced calls and variables back to a `@OneToMany` annotation on a property of an entity containing a misconfiguration and fixed it:

```diff
     /**
      * @var Collection<int, Page>|null
-     * @ORM\OneToMany(targetEntity="App\Entity\Page", mappedBy="user", cascade="persist")
+     * @ORM\OneToMany(targetEntity="App\Entity\Page", mappedBy="user", cascade={"persist"})
      */
    private ?Collection $pages;
```

It's not clear from the error neither the trace (coming from a query being executed and objects being hydrated) on which entity and which property this occurred.

I looked into `\Doctrine\Common\Annotations\DocParser` and there are many checks being done and many places where there might be a `\Doctrine\Common\Annotations\AnnotationException` being thrown with clear messages.

Should we catch `TypeError`s upon the annotation instantiation [here](https://github.com/doctrine/annotations/blob/1.13.x/lib/Doctrine/Common/Annotations/DocParser.php#L944) and [here](https://github.com/doctrine/annotations/blob/1.13.x/lib/Doctrine/Common/Annotations/DocParser.php#L971) and throw an `AnnotationException` with a proper message containing the annotation and property on which it occurred?

7ochem:
When an annotation class is instantiated and values are passed to the constructor, TypeErrors and ArgumentCountErrors can be thrown. This change catches those exceptions and shows an AnnotationException containing the annotation name and context.

Resolves: #437

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -12,6 +12,7 @@ use ReflectionException;
 use ReflectionProperty;
 use RuntimeException;
 use stdClass;
+use Throwable;
 
 use function array_keys;
 use function array_map;
@@ -941,7 +942,7 @@ EXCEPTION
 
         if (self::$annotationMetadata[$name]['has_named_argument_constructor']) {
             if (PHP_VERSION_ID >= 80000) {
-                return new $name(...$values);
+                return $this->instantiateAnnotiation($originalName, $this->context, $name, $values);
             }
 
             $positionalValues = [];
@@ -968,16 +969,16 @@ EXCEPTION
                 $positionalValues[self::$annotationMetadata[$name]['constructor_args'][$property]['position']] = $value;
             }
 
-            return new $name(...$positionalValues);
+            return $this->instantiateAnnotiation($originalName, $this->context, $name, $positionalValues);
         }
 
         // check if the annotation expects values via the constructor,
         // or directly injected into public properties
         if (self::$annotationMetadata[$name]['has_constructor'] === true) {
-            return new $name($values);
+            return $this->instantiateAnnotiation($originalName, $this->context, $name, [$values]);
         }
 
-        $instance = new $name();
+        $instance = $this->instantiateAnnotiation($originalName, $this->context, $name, []);
 
         foreach ($values as $property => $value) {
             if (! isset(self::$annotationMetadata[$name]['properties'][$property])) {
@@ -1455,5 +1456,32 @@ EXCEPTION
         }
 
         return $values;
+    }
+
+    /**
+     * Try to instantiate the annotation and catch and process any exceptions related to failure
+     *
+     * @param class-string        $name
+     * @param array<string,mixed> $arguments
+     *
+     * @return object
+     *
+     * @throws AnnotationException
+     */
+    private function instantiateAnnotiation(string $originalName, string $context, string $name, array $arguments)
+    {
+        try {
+            return new $name(...$arguments);
+        } catch (Throwable $exception) {
+            throw AnnotationException::creationError(
+                sprintf(
+                    'An error occurred while instantiating the annotation @%s declared on %s: "%s".',
+                    $originalName,
+                    $context,
+                    $exception->getMessage()
+                ),
+                $exception
+            );
+        }
     }
 }
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -13,6 +13,7 @@ parameters:
     ignoreErrors:
         - '#Instantiated class Doctrine_Tests_Common_Annotations_Fixtures_ClassNoNamespaceNoComment not found#'
         - '#Property Doctrine\\Tests\\Common\\Annotations\\DummyClassNonAnnotationProblem::\$foo has unknown class#'
+        - '#Call to an undefined static method PHPUnit\\Framework\\TestCase::expectExceptionMessageRegExp\(\)#'
 
         # That tag is empty on purpose
         - '#PHPDoc tag @var has invalid value \(\)\: Unexpected token "\*/", expected type at offset 9#'
--- a/tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
@@ -14,15 +14,19 @@ use Doctrine\Tests\Common\Annotations\Fixtures\AnnotationWithConstants;
 use Doctrine\Tests\Common\Annotations\Fixtures\ClassWithConstants;
 use Doctrine\Tests\Common\Annotations\Fixtures\InterfaceWithConstants;
 use InvalidArgumentException;
+use PHPUnit\Framework\Constraint\ExceptionMessage;
 use PHPUnit\Framework\TestCase;
 use ReflectionClass;
+use TypeError;
 
 use function array_column;
 use function array_combine;
 use function assert;
 use function class_exists;
 use function extension_loaded;
+use function get_parent_class;
 use function ini_get;
+use function method_exists;
 use function sprintf;
 use function ucfirst;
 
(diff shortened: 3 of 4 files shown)


lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -1456,4 +1458,102 @@ private function resolvePositionalValues(array $arguments, string $name): array
...
+     * @param array<string,mixed>|null $arguments
+     *
+     * @return object
+     *
+     * @throws AnnotationException
+     */
+    private function instantiateAnnotiation(string $originalName, string $context, string $name, ?array $arguments = null)
+    {
+        try {
+            if ($arguments === null) {
```
stof: couldn't you use an empty array as default value and then always use `return new $name(...$arguments);` ?

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -1456,4 +1458,102 @@ private function resolvePositionalValues(array $arguments, string $name): array
...
+
+            return new $name(...$arguments);
+        } catch (ArgumentCountError $argumentCountError) {
+            $matchCount = preg_match(
+                '/^Too few arguments to function (.*?)::__construct\(\), (\d+) passed in .*? on line \d+ and (at least|exactly) (\d+) expected/',
+                $argumentCountError->getMessage(),
+                $matches
+            );
+            if ($matchCount === 1 && $matches[1] === $name) {
+                throw AnnotationException::creationError(sprintf(
```
stof: Can we attach the caught error as previous exception ?

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -1456,4 +1458,102 @@ private function resolvePositionalValues(array $arguments, string $name): array
...
+expects "%s" to be declared.
+EXCEPTION
+                    ,
+                    $originalName,
+                    $context,
+                    $matches[2]
+                ));
+            }
+
+            // ArgumentCountError is not in the expected form or was not caused by the call to the `$name` constructor, so re-throw.
```
stof: Well, the exception happened during the instantiation. So I would still always provide the context of which annotation was instantiated, to allow debugging it.
Attaching the original error as previous throwable means we can support it for any error (Saying something like `Instantiating the annotation @%s declared on %s failed.`)

stof: and this would remove the need to try matching PHP error messages (which are not covered by BC rules AFAIK)

stof (changes requested):
This should have tests

lib/Doctrine/Common/Annotations/DocParser.php
7ochem: Good point. While working on this, it all started to be a bit bloated. So this would indeed be a better solution.

7ochem:
> This should have tests

Thanks Christophe for the super fast feedback and valid points 🙏🏻 I'll work on it a bit more

7ochem:
I changed the code and made it catch all `\Throwable` when instantiating the annotation class. This meant that I had to update some existing tests that were testing for an `\InvalidArgumentException` thrown in the constructor of the annotation. The advantage is that these exceptions are now also provided with the context and target.

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -1456,4 +1457,31 @@ private function resolvePositionalValues(array $arguments, string $name): array
...
+     * @return object
+     *
+     * @throws AnnotationException
+     */
+    private function instantiateAnnotiation(string $originalName, string $context, string $name, array $arguments)
+    {
+        try {
+            return new $name(...$arguments);
+        } catch (Throwable $exception) {
+            throw AnnotationException::creationError(
```
7ochem, line 1476: Maybe... I should add here:
```php
if ($exception instanceof AnnotationException) {
    throw $exception;
}
```

Because annotations have annotations themself. And if an exception happens in those, then you get that exception, wrapped up in an AnnotationException and that will be wrapped up in an AnnotationException (and can go on...).

stof: instantiating a class will not read its doctrine annotations

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -897,9 +909,20 @@ public function testAnnotationEnumInvalidLiteralDeclarationException(): void
...
         $parser->setIgnoreNotImportedAnnotations(false);
-        $this->expectException(InvalidArgumentException::class);
-        $this->expectExceptionMessage('Undefined enumerator value "3" for literal "AnnotationEnumLiteral::THREE".');
-        $parser->parse($docblock);
+        $this->expectException(AnnotationException::class);
+        try {
+            $parser->parse($docblock);
+        } catch (AnnotationException $exc) {
+            $previous = $exc->getPrevious();
+            $this->assertNotEmpty($previous);
```
stof: can be removed. The `assertInstanceOf` will already reject `null`

lib/Doctrine/Common/Annotations/DocParser.php
7ochem: That's true

phpstan.neon
```diff
@@ -13,6 +13,7 @@ parameters:
     ignoreErrors:
         - '#Instantiated class Doctrine_Tests_Common_Annotations_Fixtures_ClassNoNamespaceNoComment not found#'
         - '#Property Doctrine\\Tests\\Common\\Annotations\\DummyClassNonAnnotationProblem::\$foo has unknown class#'
+        - '#Call to an undefined static method PHPUnit\\Framework\\TestCase::expectExceptionMessageRegExp\(\)#'
```
7ochem, line 16: Here https://github.com/doctrine/annotations/blob/065fad5dd716558d954075bbc32093cec55594e2/tests/Doctrine/Tests/Common/Annotations/DocParserTest.php#L1752-L1756

Either one of those exists: `expectExceptionMessageRegExp` exists in PHPUnit 7, deprecated in 8 and removed in 9 (and thus PHPStan gives an error). `expectExceptionMessageMatches` exists in PHPUnit 8 and up

malarzm (approved):

malarzm:
Thanks @7ochem for the improvement and sorry for the long merge time!

7ochem:
> Thanks @7ochem for the improvement and sorry for the long merge time!

Thanks for merging! And no problem 👍🏻

PR #444, opened 2022-07-11
Update PsrCachedReader.php

Univalgo:


malarzm:
@Univalgo thanks for the PR! Could you please elaborate what are you trying to fix? My guess goes for anonymous classes but I can't be sure without description. Also please note that we won't be merging PRs without a test as we need a way to ensure the bug in question won't come back.

derrabus:
Closing due to lack of feedback.

PR #447, opened 2022-08-29
8294 - Write test class valid name

JohJohan:


JohJohan:
Please do not merge see https://github.com/doctrine/orm/pull/8296

PR #451, opened 2022-10-04
Check DocParser::PlainValue for false return value

DieterHolvoet:
`DocParser::PlainValue` always returns the result of `DocParser::Annotation`, even if it returns false because of an invalid/ignored annotation. `DocParser::Annotations` on the contrary checks the return value and returns early if it's false, actually ignoring the annotation, which seems like the right behaviour.

This causes an issue where a nested annotation in an ignored namespace appears in the parsing result as false, breaking Symfony validation with the following error:

`
An error occurred while instantiating the annotation @Assert\Collection declared on property SomeClass::$someProperty: "The value "" is not an instance of Constraint in constraint "Symfony\Component\Validator\Constraints\Required".".
`

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -1239,7 +1239,11 @@ EXCEPTION
         }
 
         if ($this->lexer->isNextToken(DocLexer::T_AT)) {
-            return $this->Annotation();
+            $annotation = $this->Annotation();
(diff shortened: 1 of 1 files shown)


greg0ire:
Can you please add a test? It will prevent regressions but also help understanding the issue.

DieterHolvoet:
I cannot seem to figure out how to reproduce this issue anymore.

greg0ire:
Let's close this then?

PR #452, opened 2022-10-04
Added support for backed enum

satalaondrej:
In php 8.1 support for enums was added.

Currently, if enum case is used instead of constant, annotations fail with unintuitive message, such as
`Fatal error: Uncaught TypeError: Illegal offset type in /data/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/DocParser.php on line 1357`.

This PR adds support for usage of backend enums in annotations (used same as class constants), and throws exception in case non-backed enum is used (which cannot be easily converted to int/string ).

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -426,7 +426,7 @@ final class DocParser
      * If any of them matches, this method updates the lookahead token; otherwise
      * a syntax error is raised.
      *
-     * @phpstan-param list<int|string> $tokens
+     * @phpstan-param list<mixed[]> $tokens
      *
      * @throws AnnotationException
      */
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -1157,7 +1157,7 @@ EXCEPTION
 
 		/** @noinspection PhpElementIsNotAvailableInCurrentPhpVersionInspection */
 		if (interface_exists('\UnitEnum') && $value instanceof \UnitEnum) {
-			if (!property_exists($value, 'value')) {
+			if (!interface_exists('\BackedEnum') || !($value instanceof \BackedEnum)) {
 				throw AnnotationException::semanticalErrorConstants('Enum ' . $identifier. ' is not backed enum, only backed enums can be used!');
 			}
 			$value = $value->value;
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -426,7 +426,7 @@ final class DocParser
      * If any of them matches, this method updates the lookahead token; otherwise
      * a syntax error is raised.
      *
-     * @phpstan-param list<mixed[]> $tokens
+     * @phpstan-param list<int|string> $tokens
      *
      * @throws AnnotationException
      */
@@ -1153,7 +1153,17 @@ EXCEPTION
             throw AnnotationException::semanticalErrorConstants($identifier, $this->context);
         }
 
-        return constant($identifier);
+		$value = constant($identifier);
+
+		/** @noinspection PhpElementIsNotAvailableInCurrentPhpVersionInspection */
+		if (interface_exists('\UnitEnum') && $value instanceof \UnitEnum) {
+			if (!property_exists($value, 'value')) {
+				throw AnnotationException::semanticalErrorConstants('Enum ' . $identifier. ' is not backed enum, only backed enums can be used!');
+			}
+			$value = $value->value;
+		}
+
+        return $value;
     }
 
     private function identifierStartsWithBackslash(string $identifier): bool


lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -1153,7 +1153,17 @@ private function Constant()
             throw AnnotationException::semanticalErrorConstants($identifier, $this->context);
         }
 
-        return constant($identifier);
+		$value = constant($identifier);
+
+		/** @noinspection PhpElementIsNotAvailableInCurrentPhpVersionInspection */
+		if (interface_exists('\UnitEnum') && $value instanceof \UnitEnum) {
+			if (!property_exists($value, 'value')) {
```
stof: the right way to check that would be checking for `BackedEnum` instead

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -1153,7 +1153,17 @@ private function Constant()
...
+				throw AnnotationException::semanticalErrorConstants('Enum ' . $identifier. ' is not backed enum, only backed enums can be used!');
+			}
+			$value = $value->value;
```
stof, line 1163: I don't think turning `MyEnum::FOO` into `MyEnum::FOO->value` is the right way to support enums in annotations. To me, that's quite confusing.

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -426,7 +426,7 @@ private function match(int $token): bool
      * If any of them matches, this method updates the lookahead token; otherwise
      * a syntax error is raised.
      *
-     * @phpstan-param list<mixed[]> $tokens
+     * @phpstan-param list<int|string> $tokens
```
stof: This seems to be a phpdoc fix for the existing code, unrelated to supporting enums. So I suggest submitting it as a separate PR instead.

stof (commented):
This clearly misses tests

lib/Doctrine/Common/Annotations/DocParser.php
satalaondrej: ~~BackedEnum is not yet implemented natively by backed enums as of php 8.1, so calling/checking instaneocf BackedEnum always returns false even with backed enums unless you implement if manually, check first user contributed note here https://www.php.net/manual/en/class.backedenum.php~~

lib/Doctrine/Common/Annotations/DocParser.php
satalaondrej: Fixed

lib/Doctrine/Common/Annotations/DocParser.php
satalaondrej: This way when someone changes class with constants used in annotations to backed enum with cases, no additional changes in annotations are required. What would be alternative ?

lib/Doctrine/Common/Annotations/DocParser.php
stof: This is totally false. BackedEnum is fully working in PHP 8.1: https://3v4l.org/QcN8S#v8.1.11

satalaondrej: Apologies, the IntBackedEnum and StringBackedEnum interfaces (which sadly do have stubs in phpstorm) were problematic/not implemented so i mistook it. Should be fixed/improved now

stof: IntBackedEnum and StringBackedEnum are PHPStorm inventions in their stub types. Those interfaces don't exist in PHP (and are not planned to exist as they are not proposed at all).

lib/Doctrine/Common/Annotations/DocParser.php
stof: well, if `MyEnum::FOO` is used in an annotation, I would expect it to actually mean `MyEnum::FOO` and not a weird behavior of making it mean `MyEnum::FOO->value`

stof: Btw, I'm quite sure `MyEnum::FOO` actually works fine today and that your special logic actually breaks such usage.

satalaondrej: For example when using backed enum in DiscripimatorMap (which i do think should work/be supported)

/\*\* 
 \* @ORM\DiscriminatorMap({
 \*     MyEnum::FOO=Example::class
 \*	 })
 \* /

Annotations fail with message `Illegal offset type`  because the enum case (and not the enum case value) is used as array key in function Arrayx in DocParser (DocParser:1357). 
And using MyEnum::FOO->value directly in the annotation does not work either.
I haven't found a proper way (other than copying the value a second time as constant instead of case) to use backed enum in annotation for discriminator map

greg0ire:
I think we should not merge new features on this library. People experiencing this issue should migrate to attributes IMO.

derrabus:
I agree with @greg0ire that we should not work on actively supporting enums in this library. A migration towards native attributes should be encrouaged instead.

And even if we would work on enum support, I disagree with the implementation proposed in this PR which would basically replace enum instances with their backing value.

I'm closing this PR because of the reasons given by @greg0ire, @stof and myself. That being said, thank you very much for your proposal.

PR #459, opened 2022-10-25
Added test cases for enums

derrabus:
By coincidence, we already support enums in annotations. Let's add some tests to make sure it stays that way.

### DIFF
--- a/tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
+++ b/tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
@@ -6,6 +6,8 @@ use Doctrine\Common\Annotations\AnnotationReader;
 use Doctrine\Common\Annotations\DocParser;
 use Doctrine\Common\Annotations\Reader;
 use Doctrine\Tests\Common\Annotations\Fixtures\Annotation\SingleUseAnnotation;
+use Doctrine\Tests\Common\Annotations\Fixtures\AnnotationWithEnumProperty;
+use Doctrine\Tests\Common\Annotations\Fixtures\ClassWithEnumAnnotations;
 use Doctrine\Tests\Common\Annotations\Fixtures\ClassWithFullPathUseStatement;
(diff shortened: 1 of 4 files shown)


malarzm (dismissed):
Do we want to advettise the fact in docs?

tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
```diff
@@ -295,4 +298,29 @@ public function testFunctionAnnotation(): void
...
+        self::assertSame($expectedValue, $annotation->suit);
+    }
+
+    /**
+     * @return list<array{string, Suit}>
+     */
+    public function provideEnumProperties(): array
+    {
+        return [
+            ['annotationWithDefaults', Suit::Hearts],
```
malarzm: Sadly you can't use enums here as PHP will choke on parsing file when running phpunit on older versions

derrabus: I'm surprised the data provider is executed. I'll fix that.

derrabus:
I don't know if it's worth mentioning enums in the docs. 🤷🏻‍♂️

stof (approved):

tests/Doctrine/Tests/Common/Annotations/AnnotationReaderTest.php
malarzm: PHP breaks on parsing the file, not during execution :)

derrabus: No, the file is parsed just fine as you can tell by the green test suite.

malarzm: Oh, that's actually good to know! I was basing on my last experience with doctrine/common where I hid actual enum usage as much as possible, must have drawn wrong conclusions :)

stof: parsing the file containing the enum is what fails. So as long as you don't trigger autoloading for it, you're fine.

PR #460, opened 2022-12-10
Allow doctrine/lexer 2

greg0ire:
It is still possible to avoid a BC-break while doing so.

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocLexer.php
+++ b/lib/Doctrine/Common/Annotations/DocLexer.php
@@ -15,6 +15,8 @@ use function substr;
 
 /**
  * Simple lexer for docblock annotations.
+ *
+ * @template-extends AbstractLexer<DocLexer::T_*>
  */
 final class DocLexer extends AbstractLexer
 {
@@ -39,7 +41,7 @@ final class DocLexer extends AbstractLexer
     public const T_COLON               = 112;
     public const T_MINUS               = 113;
 
-    /** @var array<string, int> */
+    /** @var array<string, self::T*> */
     protected $noCase = [
         '@'  => self::T_AT,
         ','  => self::T_COMMA,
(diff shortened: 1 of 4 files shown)


greg0ire:
SA is broken, I will fix it once doctrine/lexer 2 is removed, but I just wanted to show here that things are OK according to PHPUnit.

lib/Doctrine/Common/Annotations/DocLexer.php
```diff
@@ -126,4 +126,12 @@ protected function getType(&$value)
 
         return $type;
     }
+
+    /** @return array{value: int|string, type:string|int|null, position:int} */
+    public function peek(): array
```
greg0ire, line 133: No need to also override `glimpse` since it returns whatever `peek()` returns.

derrabus (approved):
Sacrificing the Liskov substitution principle on the altar of backwards compatibility. 🤞🏻

lib/Doctrine/Common/Annotations/DocLexer.php
```diff
@@ -126,4 +128,12 @@ protected function getType(&$value)
 
         return $type;
     }
+
+    /** @return array{value: int|string, type:self::T_*|null, position:int} */
+    public function peek(): array
+    {
+        $token = parent::peek();
+
+        return (array) $token;
```
alcohol, line 137: Should the array type always be forced here? In `v1.13.x` that was not the case because the method was inherited from the parent class. The docblock does not state that an empty array could be returned, but that is exactly what will be returned when the parent returns `null` right now.

derrabus: No, we should not cast `null` to an array. Up for a PR?

alcohol: Fixed in https://github.com/doctrine/annotations/pull/463 :tada:

PR #463, opened 2022-12-12
Do not cast null to array

### LINKED ISSUE #462: [1.13.3 to 1.14.0] Undefined array key "type"
When using PHP-CS-Fixer the following code breaks:

```php
<?php

declare(strict_types=1);

namespace App;

final class SomeTest extends BaseTest
{
    /**
     * @test
     * @doesNotPerformAssertions
     */
    public function it_should_validate() : void
    {

    }
}
```

```
        [ErrorException]
        Undefined array key "type"

      {closure}()
        in vendor/friendsofphp/php-cs-fixer/src/Doctrine/Annotation/Tokens.php at line 74
      PhpCsFixer\Doctrine\Annotation\Tokens::createFromDocComment()
        in vendor/friendsofphp/php-cs-fixer/src/AbstractDoctrineAnnotationFixer.php at line 62
      PhpCsFixer\AbstractDoctrineAnnotationFixer->applyFix()
        in vendor/friendsofphp/php-cs-fixer/src/AbstractFixer.php at line 75
```

I'm seeing this more when using short tags like `@doesNotPerformAssertions` or `@BeforeScenario` (Behat).

greg0ire:
The possibility that null could be returned was overlooked in 2286f7ba1e943d3be4ce22503feac8c683e35bd5 .

Fixes #462

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -22,5 +22,3 @@ parameters:
 
         # That tag is empty on purpose
         - '#PHPDoc tag @var has invalid value \(\)\: Unexpected token "\*/", expected type at offset 9#'
-        # Backwards-compatibility
-        - '#^Return type.*of method.*DocLexer::peek.*should be compatible.*$#'
--- a/lib/Doctrine/Common/Annotations/DocLexer.php
+++ b/lib/Doctrine/Common/Annotations/DocLexer.php
@@ -130,9 +130,13 @@ final class DocLexer extends AbstractLexer
     }
 
     /** @return array{value: int|string, type:self::T_*|null, position:int} */
-    public function peek(): array
+    public function peek(): ?array
     {
         $token = parent::peek();
+
+        if ($token === null) {
+            return null;
+        }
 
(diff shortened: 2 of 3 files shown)


derrabus (dismissed):

phpstan.neon
```diff
@@ -22,5 +22,3 @@ parameters:
 
         # That tag is empty on purpose
         - '#PHPDoc tag @var has invalid value \(\)\: Unexpected token "\*/", expected type at offset 9#'
-        # Backwards-compatibility
-        - '#^Return type.*of method.*DocLexer::peek.*should be compatible.*$#'
```
greg0ire, line 26: Phpstan warned me :see_no_evil:

greg0ire: Actually, the parent type returns `array`, so it should not be happy with this :thinking:

greg0ire: Ah Phpstan is only at level 3 :thinking:

PR #465, opened 2022-12-16
Drop Lexer 1

greg0ire:
This is not a big deal because `doctrine/lexer` 1 has the same version constraints as `doctrine/lexer` 2.
Note that I'm removing the BC-layer for `DocLexer::peek()` and `DocLexer::glimpse()`: in retrospect, I think it was a bad idea because there was already a BC-break with `DocLexer::$lookahead` and `DocLexer::$token`.

~If you don't think that is acceptable, let me know and I will retarget to 2.0.x  (and release it in a short amount of time).~

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,0 +1,6 @@
+# Upgrade from 1.0.x to 2.0.x
+
+`DocLexer::peek()` and `DocLexer::glimpse` now return
+`Doctrine\Common\Lexer\Token` objects. When using `doctrine/lexer` 2, these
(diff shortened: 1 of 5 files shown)


greg0ire:
:thinking: now I remember #462 and I think it might be a bad idea to do this on 1.x… php-cs-fixer seems to be using the `DocLexer` directly: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/src/Doctrine/Annotation/Tokens.php#L61

Let's not shy away from a new major.

SenseException (approved):

PR #467, opened 2022-12-17
Add parameter type declarations

greg0ire:


derrabus:
We should bump to PHP 7.2 at least, otherwise it's tough for downstream projects to support v1 and v2.

greg0ire:
:thinking: if I bump to 8.1 I can add more parameter type declarations (some with `mixed`, some with union types). Then, the ORM can still add compatibility with 2.0, and allow lexer 3.

derrabus:
> :thinking: if I bump to 8.1 I can add more parameter type declarations

Yes, then again a PHP-8.1-only version of doctrine/annotations feels a bit pointless given that attributes exist.

greg0ire:
If you want to migrate from annotations to attributes, you have to upgrade to PHP 8 first. If you do, you are probably not going to update to PHP 8.0, right?

greg0ire:
Anyway, I agree it doesn't make a lot of sense, the thing is, I have to release 2.0 with doctrine/lexer 3 support soon, because that way, I can add support for doctrine/lexer 3 in ORM 2.15. I thought I might as well do this, but if you feel I should just tag 2.0 now, let me know.

derrabus:
I think, the PR does make sense. It's just that I'd bump to PHP 7.2 (to gain parameter type widening) or maybe 7.4 if you like. But not beyond that.

greg0ire:
Ok, let's bump to 7.2 then 👍

derrabus (dismissed):

PR #469, opened 2022-12-19
Remove SimpleAnnotationReader

derrabus:


greg0ire (approved):
Didn't spot that one… I'm not sure what I grepped, maybe there is more deprecated stuff? I'll check before releasing 2.0

PR #472, opened 2022-12-22
Add supporting for a variadic parameter [2.x]

roxblnfk:
The PR adds supporting for variadic parameters in annotations with `NamedArgumentConstructor`.  
Before this PR, the following code would throw an exception:

```php
/**
 * @Annotation
 * @NamedArgumentConstructor
 */
class SomeAnnotation
{
    public function __construct(
        private string $foo,
        private string ...$bar,
    ) {
    }
}
//  ReflectionException: Internal error: Failed to retrieve the default value
```

---

Also, I've found some strange behavior in the case when a named argument conflicts with a positional argument.
If we use the annotation from the previous example, in the following code the `$foo` parameter will have the `"bar"` value.

```php
/**
 * @SomeAnnotation('foo', foo: 'bar')
 */
```

In the case with PHP attributes, there an exception [will be thrown](https://3v4l.org/bMZhT).

I've added TODOs to improve this behavior but, I think, you will choose the legacy one.

CodeStyle will be passed after TODOs resolving.

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -14,8 +14,10 @@ use RuntimeException;
 use stdClass;
 use Throwable;
 
+use function array_key_exists;
 use function array_keys;
 use function array_map;
+use function array_merge;
 use function array_pop;
 use function array_values;
 use function class_exists;
@@ -524,6 +526,7 @@ final class DocParser
         ];
 
         $metadata['has_named_argument_constructor'] = false;
+        $metadata['is_variadic']                    = $metadata['has_constructor'] && $constructor->isVariadic();
 
         // verify that the class is really meant to be an annotation
         if ($metadata['is_annotation']) {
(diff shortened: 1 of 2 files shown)


derrabus:
I'm not really sure if it's worth adding this feature tbh. Annotations were always meant to be used with named arguments and variadics (at least on PHP 7) don't play well with those.

Other than that, we consider this library to be more or less feature-complete these days and advise to migrate to native attributes. This is another reason I'm very reluctant to accept new features.

roxblnfk:
You can discard introducing for new features, but it might be better to make sure that there are no exceptions on valid code cases.  
Adding for variadic parameter breaks the library because DocParser is trying to get default value.

> Other than that, we consider this library to be more or less feature-complete these days and advise to migrate to native attributes. This is another reason I'm very reluctant to accept new features.

I agree, but some projects have some legacy. And I would like to go forward without breaking changes in my packages.

derrabus:
> Adding for variadic parameter breaks the library because DocParser is trying to get default value.

That should not happen, of course. Please submit a bugfix against the 1.14.x branch. But let's not actively support variadics.

> > Other than that, we consider this library to be more or less feature-complete these days and advise to migrate to native attributes. This is another reason I'm very reluctant to accept new features.
> 
> I agree, but some projects have some legacy. And I would like to go forward without breaking changes in my packages.

Sure. This is why we still maintain the library for the time being.

roxblnfk:
> Please submit a bugfix against the 1.14.x branch.

Done.

> But let's not actively support variadics.

Sure. I have no plans about that :)

derrabus:
> > But let's not actively support variadics.
> 
> Sure. I have no plans about that :)

Well yes, that's precisely what your PR is doing.

PR #473, opened 2022-12-23
Ignore a variadic parameter [1.x]

roxblnfk:
The PR adds supporting for variadic parameters in annotations with `NamedArgumentConstructor`.

See #472

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -614,16 +614,16 @@ final class DocParser
                 $metadata['default_property'] = reset($metadata['properties']);
             } elseif ($metadata['has_named_argument_constructor']) {
                 foreach ($constructor->getParameters() as $parameter) {
-                    $metadata['constructor_args'][$parameter->getName()] = [
-                        'position' => $parameter->getPosition(),
-                    ];
                     if ($parameter->isVariadic()) {
-                        continue;
+                        break;
                     }
 
-                    $metadata['constructor_args'][$parameter->getName()]['default'] = $parameter->isOptional()
-                        ? $parameter->getDefaultValue()
-                        : null;
+                    $metadata['constructor_args'][$parameter->getName()] = [
+                        'position' => $parameter->getPosition(),
+                        'default'  => $parameter->isOptional()
+                            ? $parameter->getDefaultValue()
+                            : null,
+                    ];
                 }
             }
         }
@@ -954,10 +954,6 @@ EXCEPTION
 
             $positionalValues = [];
             foreach (self::$annotationMetadata[$name]['constructor_args'] as $property => $parameter) {
-                if (! array_key_exists('default', $parameter)) {
-                    break;
-                }
-
                 $positionalValues[$parameter['position']] = $parameter['default'];
             }
 
@@ -1446,11 +1442,6 @@ EXCEPTION
             }
 
             foreach (self::$annotationMetadata[$name]['constructor_args'] as $property => $parameter) {
-                // The parameter is variadic
-                if (!array_key_exists('default', $parameter)) {
-                    break;
-                }
-
                 if (array_key_exists($property, $namedArguments)) {
                     $values[$property] = $namedArguments[$property];
                     unset($namedArguments[$property]);
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -14,8 +14,10 @@ use RuntimeException;
 use stdClass;
 use Throwable;
 
+use function array_key_exists;
 use function array_keys;
 use function array_map;
+use function array_merge;
 use function array_pop;
 use function array_values;
 use function class_exists;
@@ -536,6 +538,7 @@ final class DocParser
 
         $metadata['has_named_argument_constructor'] = $metadata['has_constructor']
             && $class->implementsInterface(NamedArgumentConstructorAnnotation::class);
+        $metadata['is_variadic']                    = $metadata['has_constructor'] && $constructor->isVariadic();
 
         // verify that the class is really meant to be an annotation
         if ($metadata['is_annotation']) {
@@ -615,8 +618,14 @@ final class DocParser
                 foreach ($constructor->getParameters() as $parameter) {
                     $metadata['constructor_args'][$parameter->getName()] = [
                         'position' => $parameter->getPosition(),
-                        'default' => $parameter->isOptional() ? $parameter->getDefaultValue() : null,
                     ];
+                    if ($parameter->isVariadic()) {
+                        continue;
+                    }
+
+                    $metadata['constructor_args'][$parameter->getName()]['default'] = $parameter->isOptional()
+                        ? $parameter->getDefaultValue()
+                        : null;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -225,6 +226,123 @@ public function testTypicalMethodDocBlock(): void
...
+        $annot = $result[0];
+
+        self::assertInstanceOf(SomeAnnotationWithConstructorWithVariadicParam::class, $annot);
+
+        self::assertSame('Some data', $annot->name);
+        if (PHP_VERSION_ID >= 80000) {
+            self::assertSame(['foo' => 'Foo', 'bar' => 'Bar'], $annot->data);
+        } else {
+            self::assertSame(['Foo', 'Bar'], $annot->data);
+        }
```
derrabus: No go. The library must not change its behavior when upgrading PHP.

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -225,6 +226,123 @@ public function testTypicalMethodDocBlock(): void
         self::assertInstanceOf(Marker::class, $marker);
     }
 
+    public function testAnnotationWithConstructorWithVariadicParam(): void
```
derrabus: This method contains more than one test scenario. Please split it.

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -225,6 +226,123 @@ public function testTypicalMethodDocBlock(): void
...
+
+        self::assertInstanceOf(SomeAnnotationWithConstructorWithVariadicParam::class, $annot);
+
+        self::assertSame('Some data', $annot->name);
+        self::assertSame([], $annot->data);
+
+        // With named arguments
+        $docblock = <<<'DOCBLOCK'
+/**
+ * @SomeAnnotationWithConstructorWithVariadicParam(name = "Some data", foo = "Foo", bar = "Bar")
```
derrabus: This scenario must fail because the annotation does not have a parameter `$foo`.

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -225,6 +226,123 @@ public function testTypicalMethodDocBlock(): void
...
+        if (PHP_VERSION_ID >= 80000) {
+            self::assertSame(['foo' => 'Foo', 'bar' => 'Bar'], $annot->data);
+        } else {
+            self::assertSame(['Foo', 'Bar'], $annot->data);
+        }
+
+        // With named arguments shuffled
+        $docblock = <<<'DOCBLOCK'
+/**
+ * @SomeAnnotationWithConstructorWithVariadicParam(foo = "Foo", name = "Some data", bar = "Bar")
```
derrabus: Same here, the annotation class does not have parameters `$foo` and `$bar`. This scenario must fail.

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -225,6 +226,123 @@ public function testTypicalMethodDocBlock(): void
...
+        if (PHP_VERSION_ID >= 80000) {
+            self::assertSame(['foo' => 'Foo', 'bar' => 'Bar'], $annot->data);
+        } else {
+            self::assertSame(['Foo', 'Bar'], $annot->data);
+        }
+
+        // Unnamed arguments
+        $docblock = <<<'DOCBLOCK'
+/**
+ * @SomeAnnotationWithConstructorWithVariadicParam("Some data", "Foo", "Bar")
```
derrabus: I'm torn on this one, but to be consistent here, we should not allow to access the variadic parameter at all.

derrabus (changes requested):
As discussed on the other PR: The bug that we fix here is that an error is raised if the annotation class has a variadic constructor. Please only fix that bug and don't implement active support for variadics.

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
roxblnfk: It's relevant to PHP behavior. I mean spread operator. And all existing tests are passed in v2.

Ok. If the PHP behavior is not preffered (PHP behavior would be better for migration from annotations to attributes any way), let's define the needed behavior.

```php
/** @annotatnin @NamedArgumentConstructor */
class Ann {
    public function __construct(string $foo, string ..$bar) { ... }
}
```

Just explain thise cases and i'll try to code a solution:

1. ~~There mustn't be any exception when we just add a variadic parameter in an existing annotation~~
2. Named case 1: `Ann(foo = "foo", bar = "bar")`. $bar = ?
3. Named case 2: `Ann(foo = "foo", bar = {"bar", "baz"})`.  Error?
4. Positional case 1: `Ann("foo", "bar")`.  $bar = ?
5. Positional case 2: `Ann("foo", "bar", "baz")`.  $bar = ?
6. Will we have any ability to pass many values to $bar?

derrabus: > 6. Will we have any ability to pass many values to $bar?

My preferred solution would be to not allow to pass anything to `$bar` at all.

roxblnfk:
> My preferred solution would be to not allow to pass anything to $bar at all.

Let's try it 👌

roxblnfk:
@derrabus any updates?
I made ignoring for arguments related to variadic parameter.
All old tests passed.

derrabus:
> @derrabus any updates?

Patience, please. It's Christmas and we have families.

lib/Doctrine/Common/Annotations/DocParser.php
```diff
@@ -615,8 +616,15 @@ class_exists(NamedArgumentConstructor::class);
                 foreach ($constructor->getParameters() as $parameter) {
                     $metadata['constructor_args'][$parameter->getName()] = [
                         'position' => $parameter->getPosition(),
-                        'default' => $parameter->isOptional() ? $parameter->getDefaultValue() : null,
+                        'is_variadic' => $parameter->isVariadic(),
```
derrabus: Do we need to remember the variadic parameter at all? Wouldn't it be enough to not add it to `$metadata['constructor_args']`?

tests/Doctrine/Tests/Common/Annotations/DocParserTest.php
```diff
@@ -1890,6 +1998,25 @@ public function __construct()
...
+ */
+class SomeAnnotationWithConstructorWithVariadicParam
+{
+    public function __construct(string $name, string ...$data)
+    {
+        $this->name = $name;
+        $this->data = $data;
+    }
+
+    /** @var mixed */
```
derrabus:
```suggestion
    /** @var string[] */
```

lib/Doctrine/Common/Annotations/DocParser.php
roxblnfk: I think we can ignore the `is_variadic` parameter and check the `default` key only. The key will be missing for a variadic param

roxblnfk: Done.

derrabus: Sorry for the nit-picking here. But this generates the edge-case of a potentially undefined `default` key that we didn't have previously.

Again, if we moved the `$parameter->isVariadic()` check to the very beginning of the `foreach` and thus skipped adding the variadic parameter to the `$metadata['constructor_args']` array, we wouldn't need to skip it when iterating over that array later on. Or am I missing something?

roxblnfk: > Sorry for the nit-picking here

Don't worry. That's OK 👌. We need better solution :)

> But this generates the edge-case of a potentially undefined default key that we didn't have previously.

This is an unreachable case for existing code as it only applies to variadic parameters.
And it is correct (undefined `default` key) because a variadic param can't have a default value (even null).

> thus skipped adding the variadic parameter to the $metadata['constructor_args'] array

If we don't want to customize the new error for variadics only then yes, we can just skip that parameter on parsing.

Done.

derrabus:
I've submitted a simpler solution as #479. Thank you for your work on this topic!

PR #479, opened 2023-01-30
Allow annotation classes with a variadic parameter

derrabus:
Replaces #473.

This PR follows a simpler approach to fix the issue that #473 and #472 attempted to address: If an annotation class has a variadic parameter, the parser should not fail with a reflection exception.

The variadic parameter is still not accessible though, it's simply being ignored. As discussed in #472, we don't want to introduce support for variadics as a new feature. Furthermore, there is no portable way to pass named arguments to a variadic parameter that would work on PHP 7.4.

### DIFF
--- a/lib/Doctrine/Common/Annotations/DocParser.php
+++ b/lib/Doctrine/Common/Annotations/DocParser.php
@@ -613,6 +613,10 @@ final class DocParser
                 $metadata['default_property'] = reset($metadata['properties']);
             } elseif ($metadata['has_named_argument_constructor']) {
                 foreach ($constructor->getParameters() as $parameter) {
+                    if ($parameter->isVariadic()) {
+                        break;
+                    }
+
                     $metadata['constructor_args'][$parameter->getName()] = [
                         'position' => $parameter->getPosition(),
                         'default' => $parameter->isOptional() ? $parameter->getDefaultValue() : null,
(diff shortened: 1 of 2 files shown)


greg0ire:
Sorry if it's a dumb question, but what would be the point of creating an annotation with a variadic parameter if it is ignored? You might as well remove it, right? Why not instead catch the exception and throw a more meaningful one saying "there isn't support variadic exceptions, and there never will be one". Is it maybe in case a PHP class acts both as an annotation and an attribute, and using the variadic argument is only available when using the class as an attribute? Supposing the class is still usable albeit with less feature when using solely other arguments?

derrabus:
> Sorry if it's a dumb question, but what would be the point of creating an annotation with a variadic parameter if it is ignored? You might as well remove it, right?

That is correct, if you only ever use that class as an annotation. If you also use that class for other purposes, you might want to have a variadic argument as well, even if this library does not support it.

>  Why not instead catch the exception

Well, the exception is raised because we perform an illegal method call. Catching our own mistakes is just nod good style, but we might as well…

> throw a more meaningful one saying "there isn't support variadic exceptions, and there never will be one".

Yes, we could, but then again, that feels like an arbitrary limitation: We don't support that feature, so we forbid you, dear consumer of our library, to use it at all. Why.

> Is it maybe in case a PHP class acts both as an annotation and an attribute, and using the variadic argument is only available when using the class as an attribute?

Yeah, something like that.

> Supposing the class is still usable albeit with less feature when using solely other arguments?

Yes.

greg0ire (approved):

PR #483, opened 2023-06-13
Add hint to use PsrCachedReader instead of `CachedReader` and `FileCacheReader`

alexander-schranz:
Currently in the UPGRADE.md not mention what replaces this classes.


greg0ire:
Thanks @alexander-schranz !

Issue #485, opened 2023-08-13
[RFC] Flag the package as abandoned on Packagist


derrabus:
With the introduction of native attributes to the PHP language, the Doctrine Annotations package has become legacy.

Since PHP attributes are an equal or better replacement in every aspect, we should take steps to flush out doctrine/annotations of the package ecosystem. The primary goal would be that no actively maintained composer package should list the `doctrine/annotations` package as a mandatory dependency in their latest release. Instead, PHP attributes should be offered as an alternative, so downstream projects can migrate at their own pace.

I wonder if flagging the package as abandoned would help in that regard. It would make Composer issue a warning when installing dependencies. Maintainers of PHP applications can check why `doctrine/annotations` is still in their dependencies.

* If it's being pulled by an upstream dependency, they could work with the maintainers of that dependency on making `doctrine/annotations` optional.
* If they themselves still use Doctrine Annotations, the warning would make them aware that there's a replacement.

Thoughts?

malarzm:
Sure! Two things I think we should do:

1. the disclaimer in README and in docs need to be bigger (or maybe just in the readme)
2. write a goodbye blogpost, maybe with some small reference (or just pointers) how easy the switch could be?

derrabus:
> the disclaimer in README and in docs need to be bigger (or maybe just in the readme)

Agreed.

> write a goodbye blogpost, maybe with some small reference (or just pointers) how easy the switch could be?

Well, we have https://www.doctrine-project.org/2022/11/04/annotations-to-attributes.html.

derrabus:
> > the disclaimer in README and in docs need to be bigger (or maybe just in the readme)
> 
> Agreed.

see #486.

malarzm:
> Well, we have https://www.doctrine-project.org/2022/11/04/annotations-to-attributes.html.

Totally forgot about it 🙈

stof:
This blog post refers to migrating the ORM annotations (the initial parts explaining the history is not specific to the ORM but the rector-based migration is).
Given that doctrine/annotations is used by many other packages than Doctrine projects, abandoning the package would deserve a dedicated blog post IMO. It could give it more visibility and explain the expectation for the ecosystem.

Side note: the Symfony framework is currently working on deprecating its support for annotations in its various components. Maybe marking the package as abandoned could wait until after the 6.4 and 7.0 releases of Symfony (scheduled for end of November), to avoid getting tons of reports saying that users of Symfony cannot stop using this abandoned package yet.

[...]

PR #486, opened 2023-08-14
Extend deprecation notice

derrabus:
See https://github.com/doctrine/annotations/issues/485#issuecomment-1676447349

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,13 @@ which are a native replacement for annotations. As such, this library is
 considered feature complete, and should receive exclusively bugfixes and
 security fixes.
 
+We do not recommend using this library in new projects and encourage authors
+of downstream libraries to offer support for attributes as an alternative to
+Doctrine Annotations.
+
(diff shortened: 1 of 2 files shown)


greg0ire (approved):

README.md
```diff
@@ -4,6 +4,10 @@ which are a native replacement for annotations. As such, this library is
 considered feature complete, and should receive exclusively bugfixes and
 security fixes.
 
+We do not recommend using this library in new projects and encourage authors
```
malarzm, line 7: Let's add `# Project deprecated` or something like this at the top so there's something big screaming "don't use me" :) 

Also maybe a link to our blogpost https://www.doctrine-project.org/2022/11/04/annotations-to-attributes.html like "This is how Doctrine handled migration" to help folks discover this?

derrabus: I think, the ⚠️ is enough in that regard. I've added a link to the blog.

derrabus:
I completely forgot about this one. Shall we merge the change before we make a bugfix release?

@malarzm @greg0ire

greg0ire:
Sure, it just touches documentation so…

PR #491, opened 2023-09-05
Checking if the reflection is anonymous before parsing file tokens

### LINKED ISSUE #490: Bug with PhpParser when instantiating the Tokenizer
Reference:

https://github.com/doctrine/annotations/blame/6d4929214d8593ed80364267d6665edd70c2ecbe/lib/Doctrine/Common/Annotations/PhpParser.php#L60

This piece of code prepends a "<?php " string to the original content, but if the content already has the "<?php " opening tag, it errors and breaks the runtime.

odahcam:
Resolves #490

This Pull Request aims to solve #490 by bypassing the tokenizer when an anonymous class is received, because the current logic only supports normal classes declared inside a namespaced file.

### DIFF
--- a/lib/Doctrine/Common/Annotations/PhpParser.php
+++ b/lib/Doctrine/Common/Annotations/PhpParser.php
@@ -43,6 +43,11 @@ final class PhpParser
             return $reflection->getUseStatements();
         }
 
+        // we can't rely on the file's namespaces when the class is anonymous
+        if ($reflection->isAnonymous()) {
+            return [];
+        }
+
         $filename = $reflection->getFileName();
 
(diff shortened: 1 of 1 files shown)


derrabus:
Thank you. Please provide a description that explains why we should merge this change. Also, please provide a test that covers it. That being said, your change apparently breaks our existing tests.

odahcam:
Hey @derrabus sorry about that, I was so tired I forgot to. I've just updated the description. I'll see if I can update the tests later.

derrabus:
Closing the PR as stale. Please submit a new PR if you want to pick up the work.

PR #494, opened 2024-09-04
Fix filemtime() warning because of eval'd code

andrey-tech:
| Q             | A
| ------------- | ---
| Branch?       | 1.14.x
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Ticket       | Fix [#492](https://github.com/doctrine/annotations/issues/492)
| License       | MIT
| Doc PR        |  -

#### Description

This small PR fix the old problem "`filemtime()` warning on eval'd code" described in issue [#492](https://github.com/doctrine/annotations/issues/492) and in PR [#436](https://github.com/doctrine/annotations/pull/436).

I faced this bug in my current big project when upgrading PHP from 8.1 to 8.2. But I can't migrate away from doctrine/annotations and use attributes just now. So, please, check and accept this PR ASAP. Thank you!

PR #495, opened 2024-09-04
Ignore complaints about dead code 

greg0ire:
~In the code of these constructors, we are checking the type of the 'value' property of the '$values' array.
Let us reflect the fact that we are not assuming the type in the phpdoc. I think this is OK because since both of these classes are used as Doctrine annotations, there is no way for such precise types to be useful for the end user anyway. I chose that over creating a baseline for a project that does not have one yet.~

We've declared types in phpdoc, but they are not enforced by type
declarations, so we are enforcing them with checks. Psalm complains it
is dead code because it assumes the types can be trusted.

This addresses a Psalm issue I spotted while reviewing #494

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -15,4 +15,12 @@
             <directory name="vendor" />
         </ignoreFiles>
     </projectFiles>
+    <issueHandlers>
+        <NoValue>
+            <errorLevel type="suppress">
+                <file name="lib/Doctrine/Common/Annotations/Annotation/Enum.php" />
+                <file name="lib/Doctrine/Common/Annotations/Annotation/Target.php" />
+            </errorLevel>
+        </NoValue>
+    </issueHandlers>
 </psalm>


derrabus:
What issue does this fix exactly?

greg0ire:
The issue is not very clearly expressed:

```
ERROR: NoValue - lib/Doctrine/Common/Annotations/Annotation/Enum.php:51:49 - All possible types for this argument were invalidated - This may be dead code (see https://psalm.dev/179)
                    is_object($var) ? get_class($var) : gettype($var)


ERROR: NoValue - lib/Doctrine/Common/Annotations/Annotation/Target.php:77:61 - All possible types for this argument were invalidated - This may be dead code (see https://psalm.dev/179)
                    is_object($values['value']) ? get_class($values['value']) : gettype($values['value'])
```

Since the phpdoc says `scalar`, and we are inside a condition on `!is_scalar(…)`, the inferred type is `never`, and that should never happen, so Psalm suspects there is dead code.

SenseException:
While there isn't a baseline file, you can find `ignoreErrors` directly in [phpstan.neon](https://github.com/doctrine/annotations/blob/2.0.x/phpstan.neon).

derrabus:
But those "errors" happen during input validation where we handle exactly the cases that we declare unacceptable in the doc blocks that you're about to change. Psalm complaining about input validation is just Psalm being Psalm.

I would simply ignore them.

greg0ire:
Ok, let me adjust.

greg0ire:
@SenseException I did the Psalm equivalent of what you mentioned.

derrabus (approved):
LGTM. The PR title needs to be adjusted though.

PR #506, opened 2024-10-28
Test against php 8.4

sergiy-petrov:


greg0ire (approved):
This is OK, but it prompts me to revisit https://github.com/doctrine/annotations/issues/485 because I don't think we should add more and more PHP versions forever.

SenseException (approved):

greg0ire:
Thanks @sergiy-petrov !

PR #507, opened 2024-10-28
Stop maintaining 1.x

greg0ire:
This should make the maintenance of this package even more minimal than it currently is. No more merge-ups.

If we are OK with this we should delete 1.14.5 from https://github.com/doctrine/annotations/milestones

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -17,7 +17,7 @@
         {
             "name": "1.14",
             "branchName": "1.14.x",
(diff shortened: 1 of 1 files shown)


greg0ire:
Also, maybe we should proceed with https://github.com/doctrine/annotations/issues/485

SenseException (approved):
As far as I can see are the remaining merged PRs of the last milestone already in 2. This should be :+1: .

